PR #175, opened 2013-06-09
Moonflower

MirceaKitsune:
I was notified on IRC that the developers agreed with adding the Moonflower to Minetest. This commit integrates it with the default Flowers mod. The moonflower code is still in its own script file given its specific functionality, and keeps its own mapgen spawn system... let me know if you prefer that to be changed.

http://forum.minetest.net/viewtopic.php?pid=82051


C1ffisme:
This is the kind of things for mods, I think. As said by @HybridDog : 
> in my opinion you need to try to make minetest realistic, if you don't, the nyancat would loose its  virtue, which would make me sad ;¬{

Also, this pull hasn't seen any activity for too long. It and the duplicate should be closed.

BlockMen:
duplicate and still controversial. if wanted it can be disucssed on original pull further

PR #181, opened 2013-06-23
Make the default:grass texture tile better

cnelsonsic:
It bothered me every time I saw the grass texture, so I ran it through Gimp's Resynthesizer plugin and came up with this.

Rather than sit on my own changes to scratch my own itch, I figured I'd share. :)
## Before

default_grass_original_tiled
screenshot_1906448945
## After

default_grass_tiled
screenshot_1906596236


kilbith:
I'm disagree with paramat. This grass is really too bright/eye-agressive, and the grid is too obvious.

The current grass is the perfect balance between the bright/dark, without a grid. The shadows are pretty enhanced with a soft gradient of natural green.

PR #188, opened 2013-07-09
Add wield light for torch and lava

Zeg9:
Only merge if https://github.com/minetest/minetest/pull/816 is merged.


neoascetic:
I would also like to hear a sound of flaming torch, just like I've implemented in [my mod] (https://github.com/neoascetic/minetest_flaming_torch).

neoascetic:
May also be closed, since dependency is closed

PR #200, opened 2013-09-08
Make furnaces burn items even when they are unloaded

Novatux:
It uses minetest.get_gametime() to simulate them when they are loaded again, so there is no increase in server load.


HybridDog:
idk if it works
https://github.com/HybridDog/minetest_game/commit/595bb7e63d5a96d0ab7aada4d2ba36253e22dbe6

Novatux:
Closing, since #395 is far better

PR #232, opened 2013-12-17
Workbench

MirceaKitsune:
Workbench intended for official inclusion in minetest_game. When used, it offers an inventory window with a 3 x 3 crafting grid. The non-creative inventory now uses a 2 x 2 crafting grid, but creative inventory remains unchanged.

Unlike other workbench mods, this is not subject to any lag. That's because the workbench doesn't use an inventory of its own, but works with the player's inventory. When activated, the player's craft area is resized to 3 x 3 and the workbench formspec is shown. When the formspec is closed, the player's craft area is set back to standard 2 x 2 size, so the grid in the inventory window continues to work correctly. If items are forgotten in the player's craft area, they are automatically put back into their main inventory, making sure no items are lost as the craft area changes size.

Textures and most of the code by me. The function to move the player's inventory was made by the generous diemartin on IRC. I was unable to find any glitches and bugs, but please do your own testing and post here if you find anything that's broken. Screenshot of the workbench:

http://i39.tinypic.com/11m5f9y.jpg


mahfoud13:
j

PR #324, opened 2014-09-30
New textures for the ores blocks

kilbith:
Before / after :

ores_blocks


kilbith:
Gold block updated : more yellow. Screenshot above updated too (Ctrl+F5 for refresh).

kilbith:
"Gold no" : why ? I can't do better I think... Compare that to the stone_ores and the ingot, same color.
Steel : combination of what ? His texture is already realistic.
Bronze : already the case.

The current steel and copper textures do not fits very well with the tools colors, btw.

kilbith:
I'm disagree to overlay my textures with the dirty current textures.

Can't do better, do whatever with this.

kilbith:
Gold block updated.  Better ?
screenshot_3597982394

kilbith:
Sorry, but metal is smooth, not dirt-ish or noise-ish or whatever with some contrasted grains.

kilbith:
I'm afraid that I'm not enough talentuous to make different light patterns for each of them, sowwy :(

kilbith:
_Update_ : to satisfy everyone, the textures are now a little more "noisy", but keep a smooth aspect. The patterns are slightly differents for each block. Also, bronze texture is less saturated and the copper one more realistic.

After all these tweaks, it's the best I can do.

block

PR #327, opened 2014-10-04
Add new door textures, made by Esteban

Amaz1:
These door textures were made by Esteban.

Forum thread: https://forum.minetest.net/viewtopic.php?f=3&t=10282


kilbith:
@BlockMen : 

I re-worked the Esteban's doors. The wooden door is smoother, and the steel door borders are less "strong".

doors

The new files are here : door_steel_a door_steel_b door_wood_a door_wood_b

kilbith:
An handle isn't fully black under the light... And I don't like the idea of a deep black handle anyways.

PR #330, opened 2014-10-10
Vessels : new textures

kilbith:
Before / after : 

vessels


kilbith:
In what sense "a bit lighter" ? More details please.

kilbith:
Updated. Better ?

bottle

PR #331, opened 2014-10-11
New grass & leaves textures

kilbith2:
Before / after : 

grass

tree


kilbith:
Paramat : I've tried your idea, make the lower pixels of the blades a bit "connected" with the grass from dirt below. The problem is that it causes a disharmony with the top pixels of the blades. Colouring must be entire... or not. Visually, it's better to stand out a bit the high grass from the ground, IMO.

kilbith:
[Update] New texture for the tree leaves (based on the Carbone one - modified) in this pull. Screenshot in the 1st comment. I'm waiting for your feedbacks.

kilbith:
Paramat : the current minetest's style is not cartoon-like. The recent changes in the environnement are in the theme of the "natural beauty", with an advanced gradual pixelation rather than vulgar showy patterns.

kilbith:
[Update] New junglegrass : DELETED.

kilbith:
I'm waiting for the devs feedbacks about all, and then I'll make the sort in the pull.

kilbith:
[Update] @ShadowNinja @BlockMen @sfan5 @nore 
- New leaves : I finally kept the same pattern than the current one. Each leaf has now more shade of green, and the light green is more natural and less bland.
- Reverted the junglegrass for the current one.

tree

PR #332, opened 2014-10-16
New ladder texture

kilbith3:
Before / after : 

ladder


kilbith:
I think it would be better if you give some reasons why ...

I spent hours to improve the textures as volunteer, it's a bit rude to receive expeditious refusal without reason.

kilbith:
Thanks for your feedback but realistically, when you break stone, its parts DOES NOT suddently and radically change of colors. All the previous cobble were lighter or darker than the raw stone, it's wrong & disharmonious when cobble / stone / stonebricks are close near.

The current stonebricks uses the same way that my new cobble texture. Anyhow, even if you dislike my texture, the current cobble is still wrong and need a replacement.

kilbith:
Update : ladder more lighter.

kilbith:
[Update] New ladder texture : I broke the "lines" as @ShadowNinja requested. New screenie in the 1st comment.

PR #347, opened 2014-11-29
Apples give 2 HP instead of 1 HP, bread gives 5 HP instead of 4 HP

Calinou:
Apples used to give 4 HP, currently they give 1 HP. This feels quite low, so I have increased it to 2.

Bread always gave 4 HP, but it is complicated to make, so it deserves giving 5 HP.


kilbith:
:+1:

Novatux:
https://github.com/minetest/minetest_game/commit/23a37e5e7937f7db2adafe33336cf77ce3b95ea0

PR #351, opened 2014-11-30
Make some tools dig faster

Calinou:
* **Stone pickaxe:** 1.20 ⇒ 1.00 seconds to dig `cracky = 3` nodes like stone and coal ores. This should make the game feel much better to play.

* **Wooden axe:** 2.00 ⇒ 1.60 seconds to dig `choppy = 3` nodes like trees. Before, the wooden axe was no faster than the hand, as they both had 2.00 seconds digging time.

* **Stone axe:** 1.50 ⇒ 1.30 seconds to dig `choppy = 3` nodes like trees.

### DIFF
--- a/mods/default/tools.lua
+++ b/mods/default/tools.lua
@@ -40,7 +40,7 @@ minetest.register_tool("default:pick_stone", {
 		full_punch_interval = 1.3,
 		max_drop_level=0,
 		groupcaps={
-			cracky = {times={[2]=2.0, [3]=1.20}, uses=20, maxlevel=1},
+			cracky = {times={[2]=2.0, [3]=1.00}, uses=20, maxlevel=1},
 		},
 		damage_groups = {fleshy=3},
 	},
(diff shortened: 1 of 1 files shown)


C1ffisme:
I think this pull request can do better. Instead of moving just a few tools, lets move...
Wood to the speed of Stone
Stone to the speed of Iron/Bronze
Iron/Bronze only gets enough of a speed improvement to be worth getting after stone.

See, mining with stone is too 
# Slow
and should really be sped up so you can get to iron fast. Iron is essential for locks on chests, and locked doors. So it should be available fast.

HybridDog:
your hand and wood is just too fast

C1ffisme:
@HybridDog No, the hand should remain where it is. If you need some reason to keep them that fast, just consider that the player can carry 99*27 (2673 blocks! Wow!) oven-sized blocks of pure gold. So technically the player should be able to break stone with creative powers in survival.

BlockMen:
5a2aab855ff1691c91c8797686d385a3bce07489

PR #353, opened 2014-11-30
Give 24 rails per craft instead of 15

Calinou:
This should hopefully make it easier to build railroads. Also, an even number makes more sense.


Novatux:
https://github.com/minetest/minetest_game/commit/300b14078f2d969f1285b6d4233515239f6a9dc8

PR #354, opened 2014-11-30
Add 3D signs

Calinou:
Image

This adds a texture required for them to work and are made using a single nodebox. This is backwards-compatible – placement of old signs is not broken.


kilbith:
:+1:

PR #360, opened 2014-12-03
Make ice transparent

PilzAdam:
Only merge if the engine finally fixes the glitches with half-transparent nodes.


BlockMen:
Plz reopen (only) when someone fixed "the glitches with half-transparent nodes" in engine

PilzAdam:
@BlockMen why do you use the "needs-engine-change" label if you close PRs that need an engine change?

BlockMen:
@PilzAdam 
> ShadowNinja added low-priority enhancement needs engine-change labels on 6 Dec 2014

Remove them if you think they are wrong

ShadowNinja:
[ open and merge when minetest/minetest#200 is fixed ]

paramat:
Translucency issue https://github.com/minetest/minetest/issues/95

HybridDog:
paramat, it doesn't matter, it's not realistic that there's no refraction
so working z sorting would only support this unrealistic behaviour,
and it doesn't look ugly if e.g. ice behind ice isn't visible when looking to Z+ (not the case when looking to Z-):
screenshot_20160505_090310
but ice placed in a desert may look ugly…

numberZero:
Looks nice =) I like such feature (but it should be optional)

About appearance: not only ice behind ice, but back faces too should be visible. Also it should reflect sunlight (not supported by the engine AFAIK)

HybridDog:
>Also it should reflect sunlight (not supported by the engine AFAIK)

it is supported by the irrlicht engine, supertuxkart 0.8.1 had reflections (see the Hochglanz map)
and l guess refraction, birefringence etc. can be calculated by a geometry shader

for the visibility of back faces, the connected nodebox drawtype could be used:
just make it add squares to each side where it touches air

numberZero:
> it is supported by the irrlicht engine

I meant MT engine, it seems to not support specularity at all. But without shadows it would look awkwardly too often anyway.

PR #382, opened 2014-12-22
Update nodes.lua

TheShadowZone12:
Added "falling_node=1" to dirt.


BlockMen:
To havy change, especially in term of existing worlds. closing.

PR #386, opened 2014-12-23
Update obsidian textures

kilbith2:
Before / after...

old
new

---

old_door
new_door


kilbith:
@Jordach : an obsidian node is represented by sharp small _pieces_ of obsidian in itself. The one you suggest me is like stone, i.e. full. And there's no full 1x1m obsidian piece in nature.

kilbith:
An important note : for the obsidian texture, I scaled down the 512px texture from the Sphax Texture Pack and tweaked the colors, I thought the result was original but I've been informed that a 16px version of Sphax exists and looks close to the obsidian texture submitted here. So I suggest to ignore this texture and make another one.

I let someone else doing that.

kilbith:
Another note : as I said on IRC, problem is if we use a black texture like the Jordach's one, it may be confused with plenty of other black-ish materials existing - the purple allows to distinguish better the obsidian. But indeed it should be darker than that.

PR #387, opened 2014-12-23
Add straw

kilbith3:
The farming mod should be completed by a straw node that allow another use of wheat, and makes nicer the farms with a traditional look.

straw


kilbith:
@PilzAdam : done.

kilbith:
@PilzAdam : I don't need to be in the credits.txt ;)

kilbith:
@donat-b : rubbish. Check closely in Gimp.

kilbith:
The pixels are not exactly the same, but originally the straw texture comes from Sokomine (the one I used) and the DarkAge author has employed the same methods that me (noising, contrast) without I was aware of that.

Anyway, not a similar texture.

kilbith:
@donat-b : I told you that I've not only used the flipping (vertical, not horizontal). Compare the original from Sokomine (random_buildings mod) and you'll see the pixeling difference ;)

But if you absolutely want to try to make me look like a thief, it's your problem.

@CraigyDavi : not, look _closely_, the colors are not exactly the same too.

kilbith:
@ShadowNinja : I think the original straw texture is from here : https://github.com/Sokomine/random_buildings

This is the one I re-worked a slightly bit, as I said before.

Besides that, I can point an important proportion of things on popular mods/games that have been closely inspired or based on previous works (Minecraft for the most) and now all you dares to rush on a ridiculous straw texture for annoying me with a license matter. Lack of distraction before Christmas ?

Well, good luck for the future. PR closed.

Sokomine:
@0-afflatus: That's right. 9 harvested wheat for 1 straw is very expensive. If a 4-wheat-receipe is not desired due to crafting conflicts, just return 3 straw for 9 wheat (plus a suitable reverse receipe - 1 straw -> 3 wheat). Roofing material is needed in large quantities.

RHRhino:
Sokomine :+1:

ShadowNinja:
#400

PR #390, opened 2014-12-25
Add Obsidian / Obsidian Brick stairs & slabs

CraigyDavi:
Image


Novatux:
@HybridDog: this would probably cause problems for mods which would have nodes fitting those things (technic machines for example).
Anyway: it is definitely needed.

sfan5:
cab01b6694cbc37b65d649501ef9f001466cc3e3

PR #391, opened 2014-12-27
bone-mod fixes: 1. don't delete protected nodes, 2. time out in loaded chunks, 3. don't crash when dying in certain nodes (like default doors or sign_lib signs)

t4im:
uses specific checks, due to minetest/minetest#2015

### DIFF
--- a/mods/bones/init.lua
+++ b/mods/bones/init.lua
@@ -91,6 +91,7 @@ minetest.register_node("bones:bones", {
 			meta:set_string("infotext", meta:get_string("owner").."'s old bones")
 			meta:set_string("owner", "")
 		else
+			meta:set_int("time", time)
 			return true
 		end
 	end,
--- a/mods/bones/init.lua
+++ b/mods/bones/init.lua
@@ -9,6 +9,8 @@ local function is_owner(pos, name)
 	return false
 end
 
+local share_bones_time = tonumber(minetest.setting_get("share_bones_time") or 1200)
+
 minetest.register_node("bones:bones", {
 	description = "Bones",
 	tiles = {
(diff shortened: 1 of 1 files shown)


t4im:
The last commit also protects unprotected non-`buildable_to` nodes.
It also reduces calls to `is_protected` in general due to that.

Do you still see anything important missing? :)

ShadowNinja:
I think that bones in protected areas should be opened immediately, if they're even placed.

t4im:
hm.. that seems extensivly abuseable though.

The servers i saw so far have often protected areas around spawns and other public places.
Imagine some guest running at you, clubbing you to death, even though unagreed pvp might be disallowed or discouraged and taking your mk3-drill, diamonds etc. just to leave the server shortly after.
Having some bones for 5 minutes non-destructivly on ones lawn every now and then seems considerably less griefing to a serverpopulation to me.
In fact, people have pointed out to me, that 5 minutes might be a bit too low as default; 10 might be more appropriate.

That said, it's currently configurable to work exactly as you just described on a per server basis by setting `share_bones_time_early` to '1' (i.e 1 second).

The bug preventing bones from timing out in loaded chunks was fixed by 26349ea, so that would work now.

t4im:
Renaming to make clear that this isn't a feature motivated pullrequest…

I had actually hoped that it would have been approved and merged in _before_ minetest_game had backward incompatible changes (which happened 4 days ago, when this was allready ready enough) so stable running servers could benefit from these potentially crash preventing fixes.

t4im:
Actually if you would reasonably except with your linear history rule and do a simple merge instead of a rebase as the dev wiki implies on this, you could still have that. But i kinda doubt that will happen, will it?

PilzAdam:
@t4im We currently have no system in place to "backport" fixes to previous releases (see [this forum topic](https://forum.minetest.net/viewtopic.php?f=3&t=10848)).

BlockMen:
75ea7e3160321fd60c80907accb427650391bb0e

PR #401, opened 2015-01-17
Add an outlined crosshair

Calinou:
This adds an outlined crosshair, it is visible everywhere.

Also clarifies license of my textures.


kilbith:
I suggest using an alpha on the white filling. A bit too flagrant currently.

BlockMen:
Too strong outlines and to bold at all.

Ferk:
Maybe instead of an outline, it would be better to add a shadow or maybe a stripped pattern.
I made this one just now: 

crosshair

It's thinner, not very intrusive and it's visible both in dark and light nodes

screenshot 15
screenshot 16

paramat:
I find Calinou's texture too thick, the crosshair is best as unobstrusive as possible like selection boxes, it's already visible enough against the dark.
So i like Ferk's simple shadowed one, it is very similar to the original against both dark and light, if you make it a new PR i'll +1 it. The shadow is correct as hotbar cubes are also lit from the top left.

sfan5:
:-1: agree with @paramat

paramat:
BlockMen was a dev when he commented so that's -3, i fell this is enough and the PR has been left open unimproved for a year so closing.

PR #403, opened 2015-01-17
Notify player with the location of their dropped bones

CraigyDavi:
Also, log the action so you (or the server owner) can find the bones even if you weren't looking at the chat at the time.

### DIFF
--- a/mods/bones/init.lua
+++ b/mods/bones/init.lua
@@ -181,6 +181,9 @@ minetest.register_on_dieplayer(function(player)
 	end
 	
 	minetest.set_node(pos, {name="bones:bones", param2=param2})
+
+	minetest.chat_send_player(player_name, "Your bones were left at "..minetest.pos_to_string(pos))
+	minetest.log("action", player_name.." left their bones at "..minetest.pos_to_string(pos))
 	
 	local meta = minetest.get_meta(pos)
 	local inv = meta:get_inventory()


proller:
its cheat!
also showing coords in debug cheat too.

Calinou:
This is not cheating, it's fair to know your death location IMO.

donat-b:
:-1: if not configurable

Calinou:
> :-1: if not configurable

Indeed, we can have a server-side setting for this: `tell_bones_location` (only to the player who died) and `publish_bones_location` (which would show the coordinates to everyone).

MT-Modder:
:+1: For @Calinou's idea.

ShadowNinja:
*Ahem* F5 info shows your position.  It's not some huge secret, in fact if it was this would be useless because you wouldn't know where your bones were relative to you.  This isn't a cheat, and it's useful.  I don't see a good reason do disable this if you keep the rest of `bones`.

A death broadcast is probably controversial and out of the scope of this patch.  This only adds logging and a private message.

ExeterDad:
I like @Calinou's idea.
 I can't find it now, but I made a modded bones in the past that would show the bones as a waypoint for the user when issuing a chat command. Also there was a command for server owner that would show all of the bones of players as waypoints to aid in cleanup.

BlockMen:
I'm against this, it makes finding you bones even more easy. For ppl that think its that hard there is already a solution: F5

C1ffisme:
@BlockMen I have some TERRIBLE memory, and with people who's minds decay on video games like me, I think this feature would be a good idea. In fact, I used to get the memmoradum mod onto some of my adventure maps to show me where command-block teleports would go.

PilzAdam:
A chat message popping up and telling you something like this is bad game design. Add some kind of voodoo-item that points in the direction of your bones or so.

ShadowNinja:
@PilzAdam How about a temporary waypoint HUD item?  :-)

PilzAdam:
@ShadowNinja could work.
How would it be crafted? It would need to be simple, since you want to craft it after you died and you have no items in this case.

HybridDog:
But if you die and then suddenly become spawnkilled, this item wouldn't help you.

C1ffisme:
IMO, everything shouild be customizable. So why not just make another minetest.conf setting called ```show_bone_pos = (true/false)``` so that server owners can decide their opinion.

Novatux:
Closing, since #464 adds this with the option to enable/disable it

PR #410, opened 2015-01-22
Fix typo for obsidian glass door texture

Xanthin:
I think that was just a typo?


Novatux:
https://github.com/minetest/minetest_game/commit/4468015cbe956a48875b3faad6561ac4220468e8

Issue #406, opened 2015-01-20
Writable Books...


tenplus1:
ignore this, better code below...

0-afflatus:
It would be wonderful.
Can you store metadata on a craft item? I thought it was specific to placed nodes.

MT-Modder:
@0-afflatus You can, ItemStack has metadata.

:+1: I would love this feature. Can you update it so it uses the default background or is it setup like signs?

0-afflatus:
In that case :+1:

ShadowNinja:
The concept is good, but this particular implementation is unacceptable:

  1. The `££` split is a huge hack.  Either make the first line the "header" (see 3) or serialize the data from a table (better).
  2. You're missing the `max_split` argument to `string.split`, which will break books with `££`s in them.
  3. `header` is very misleading.  It's really `player_name`.  I think there should be an actual title though.
  4. You should use the book's `on_receive_fields` instead of a global callback.
  5. `text` is also very misleading.  It's a `stack_table`!

tenplus1:
I tried using on_receive_fields in the actual craftitem but it wouldnt work properly at all...  Simple solution to  a much needed feature but I will re-think code to make it better...

MT-Modder:
Cool thanks. I'm pretty sure that ```default.gui_slots``` doesn't need to be there because it's only used for inventory slots. I could be wrong though.

ShadowNinja:
Here's a better implementation: https://gist.github.com/ShadowNinja/e06b3fbe165bbac4ae84

kilbith:
@ShadowNinja : pull-request please :+1: 

![screenshot_242658820](https://cloud.githubusercontent.com/assets/7883281/5845835/93a6100a-a1bf-11e4-8aeb-dc49709263b2.png)

MT-Modder:
Does the metadata persist after placing/removing a book into/from the bookshelf?

kilbith:
@MT-Modder : yes, checked.

MT-Modder:
Sweet, thanks for checking.

tenplus1:
@ShadowNinja : looks great to me :)  my only addition was a close button when reading and showing the author to stop server abuse...

```

--= Book (use to read or write to)

minetest.register_craftitem(":default:book", {
	description = "Book",
	inventory_image = "default_book.png",
	groups = {book=1},
	stack_max = 1,
	on_use = function(itemstack, user, pointed_thing)
[...]

PR #413, opened 2015-01-24
Make books writable

ShadowNinja:
See #406.

### DIFF
--- a/mods/default/craftitems.lua
+++ b/mods/default/craftitems.lua
@@ -11,10 +11,79 @@ minetest.register_craftitem("default:paper", {
 	inventory_image = "default_paper.png",
 })
 
+local function book_on_use(itemstack, user, pointed_thing)
+	local player_name = user:get_player_name()
+	local data = minetest.deserialize(itemstack:get_metadata())
+	local title, text, owner = "", "", player_name
+	if data then
+		title, text, owner = data.title, data.text, data.owner
+	end
+	local formspec
+	if owner == player_name then
+		formspec = "size[8,8]"..default.gui_bg..
+			"field[0.5,1;7.5,0;title;Title:;"..
+				minetest.formspec_escape(title).."]"..
+			"textarea[0.5,1.5;7.5,7;text;Contents:;"..
+				minetest.formspec_escape(text).."]"..
+			"button_exit[2.5,7.5;3,1;save;Save]"
+	else
+		formspec = "size[8,8]"..default.gui_bg..
+			"label[1,0.5;"..minetest.formspec_escape(title).."]"..
+			"label[0.5,1.5;"..minetest.formspec_escape(text).."]"
+	end
+	minetest.show_formspec(user:get_player_name(), "default:book", formspec)
+end
+
+minetest.register_on_player_receive_fields(function(player, form_name, fields)
+	if form_name ~= "default:book" or not fields.save or
+			fields.title == "" or fields.text == "" then
+		return
+	end
+	local inv = player:get_inventory()
+	local stack = player:get_wielded_item()
+	local new_stack, data
+	if stack:get_name() ~= "default:book_written" then
+		local count = stack:get_count()
+		if count == 1 then
+			stack:set_name("default:book_written")
+		else
+			stack:set_count(count - 1)
+			new_stack = ItemStack("default:book_written")
+		end
+	else
+		data = minetest.deserialize(stack:get_metadata())
+	end
+	if not data then data = {} end
+	data.title = fields.title
+	data.text = fields.text
+	data.owner = player:get_player_name()
+	local data_str = minetest.serialize(data)
+	if new_stack then
+		new_stack:set_metadata(data_str)
+		if inv:room_for_item("main", new_stack) then
+			inv:add_item("main", new_stack)
+		else
+			minetest.add_item(player:getpos(), new_stack)
+		end
+	else
+		stack:set_metadata(data_str)
+	end
+	player:set_wielded_item(stack)
+end)
+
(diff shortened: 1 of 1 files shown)


kilbith:
:heart:  It just miss ..default.gui_bg_img..

ShadowNinja:
@kilbith I don't notice any difference with or without any of the `gui` things.  Maybe the colors were made the Minetest default and should be removed?  It seems like `bg_img` is sized for a different formspec anyway.

MT-Modder:
@ShadowNinja I notice a difference.
Without default.gui_bg_img - 
screenshot_2255050
With default.gui_bg_img - 
screenshot_2310117

PilzAdam:
I don't like that books are `stack_max = 1` with this. It makes crafting a lot of bookshelfs a PITA.

MT-Modder:
@PilzAdam Good point. I'm pretty sure without ```stack_max = 1``` the whole stack has the same title and contents. I wonder if there is an easy solution to this? Maybe one book can be separated from the stack, if there is room in the inventory, when on_place is called. This single book would then have the metadata but the rest of the stack is untouched.

tenplus1:
I've tweaked the code a little so that text is word-wrapped when reading the book also (players forgetting to hit enter after each line), it looks a little better now:

```

--= Book (use to read or write to)

minetest.register_craftitem(":default:book", {
	description = "Book",
	inventory_image = "default_book.png",
	groups = {book=1},
	stack_max = 1,
		local player_name = user:get_player_name()
		local data = minetest.deserialize(itemstack:get_metadata())
		local title, text, owner = "", "", player_name
		if data then
			title, text, owner = data.title, data.text, data.owner
		end
		local formspec
		if owner == player_name then
			formspec = "size[8,8]"..default.gui_bg.. -- default.gui_bg_img..
				"field[0.5,1;7.5,0;title;Title:;"..
					minetest.formspec_escape(title).."]"..
				"textarea[0.5,1.5;7.5,7;text;Contents:;"..
					minetest.formspec_escape(text).."]"..
				"button_exit[2.5,7.5;3,1;save;Save]"
			minetest.show_formspec(user:get_player_name(), "default:book", formspec)
		else
			formspec = "size[8,8]"..default.gui_bg.. -- default.gui_bg_img..
				"button_exit[7,0.25;1,0.5;close;x]"..
				"label[0.5,0;"..minetest.formspec_escape(title).."]"..
				"label[0.5,0.5;by "..owner.."]"..
				"textarea[0.5,1.5;7.5,7;text;;"..
					minetest.formspec_escape(text).."]"
			minetest.show_formspec(user:get_player_name(), "default:lockedbook", formspec)
		end
	end,
})

minetest.register_on_player_receive_fields(function(player, form_name, fields)
	if form_name ~= "default:book" or not fields.save then
		return
	end
	local stack = player:get_wielded_item()
	if minetest.get_item_group(stack:get_name(), "book") == 0 then
		return
	end
	local data = minetest.deserialize(stack:get_metadata())
	if not data then data = {} end
	data.title, data.text, data.owner =
		fields.title, fields.text, player:get_player_name()
	stack:set_metadata(minetest.serialize(data))
	player:set_wielded_item(stack)
end)

kilbith:
@tenplus1 : instead of paste the whole code, tell us what exactly you have changed at which line(s). That'd be easier to find the change(s).

C1ffisme:
@ShadowNinja  Rather than default.gui_bg_img, why not make some (slightly) nicer looking background images like here: #417.

est31:
:+1:

BlockMen:
I agree with @PilzAdam.
I think a better solution would be a new item, like written_book. Then you could still stack blank books as ususal and could handle the written books seperatly

est31:
I don't see any point in the normal ""blank" books -- what can you do with them?
Books are for writing into them, not stacking them. We are on the road to 5.0, so stuff may change.

PilzAdam:
@est31 The main usage of books is to craft bookshelfs. If you want to craft a larger amount of bookshelfs then you need very large amount of books, so they need to be stackable.

est31:
ah ok, then I support @BlockMen's suggestion.

HybridDog:
>We are on the road to 5.0, so stuff may change.

you mean 0.5.0, don't you?

tenplus1:
Stacking writable books removed any text placed inside them or copies that text across to all of the other books...  Bookshelves can easily be expanded with another row to hold plenty more, or better yet, more books... more bookshelves :)

MT-Modder:
I don't think this should be merged without adding ```default.gui_bg_img```.

C1ffisme:
Bump!

Why is this controversial? This pull adds a reason to have bookshelve inventories kept.

I do see how bookshelves will be slow to craft. I think that there is a solution. Instead of the book item being writable, there should be a new item called the "Book and Dye" which is crafted, well, you can guess, and that would be written in. Optionally, a non_creative_inventory item called a "Finished Book" could be made when doing some action with the "Book and Dye". The finished book would not be edittable.

I basically just took minecraft's method of writing books. Ah well, better to add similar features than leave newbies confused at a featureless or different method game.

Novatux:
@ShadowNinja: could code @BlockMen 's suggestion please?

ShadowNinja:
@Novatux: Done.

I can now see the difference without `gui_background`, but I kind of like it without the background.

Megaf:
:+1:

Novatux:
https://github.com/minetest/minetest_game/commit/64fa8e6be5bae28beee3b3361e927be1d6b6c8d4

C1ffisme:
YAAAAAAAAAAAAAAAAAAAY!!!!!

PR #414, opened 2015-01-25
Sam III

kilbith:
This is **Sam III** from stu (WTFPL) : https://forum.minetest.net/viewtopic.php?f=5&t=5768
- Articulated knees and elbows.
- Hat layer support.
- 2x lighter than Sam II (138 KB against 316 KB), thanks to the efficient B3D conversion.

rendercr


ShadowNinja:
I like the new model, but not the new texture.

kilbith:
Suggestions for the texture ?

ShadowNinja:
I think the current one is good.

Novatux:
Looks good

sfan5:
I also think the current texture is better.

Can you show an im-game comparison?

kilbith:
(I remit the current skin for Sam III)

@sfan5 : sure.

**Sam II**
screenshot_582972826

**Sam III**
screenshot_582863463

P.S.: it concerns an animated model. The screenshots doesn't offer a good point of view for judging it.

PilzAdam:
Does this still support the hat layer?

kilbith:
@PilzAdam : yes, it support the hat layer : https://lut.im/5ACrXy4f/TqXJebLR (top-right part of the skin image colored in green).

paramat:
Absolutely no way, this would be a huge mistake. Elbows and knees make it look creepy and no longer cute, abstract straight limbs in their silly cartoon-like motion are what make a MC/MT character adorable (see 'uncanny valley' http://en.wikipedia.org/wiki/Uncanny_valley). Again this is making the mistake of moving towards realism, voxel games are inherently simplistic and absract and look worse with excessive realism.

kilbith:
@paramat : VoXus is all focused on cartoon-ish style. minetest_game tends to focus on visual realism (you seems have a chronic discontent against that theme). Otherwise I'm agree with you about the elbows permanently bent while walking, it should swing rather.

Jordach:
kilbith; I vote no.

Some skins that players use are broken due to how the UV map is handled by
the model.

Because the wrist's top face uses the same as the top of the shoulder,
meaning some skins look completely incorrect.

kilbith:
@Jordach : OK, that gives me a good reason to close this PR then.

paramat:
Yep certainly our textures are fairly realistic, i use them and like them. But because this is a voxel game the shapes and drawtypes used are extremely abstract and simple, so it's quite right i continue to warn against more realism / higher resoluion / more complex shapes.

PR #418, opened 2015-01-30
Add reverse recipes for hoes while keeping def.recipe compatibility

MT-Modder:


C1ffisme:
Good Idea. I don't think I would have made it through minetest if I couldn't craft stairs backwards.

MT-Modder:
> Although maybe you could add a parameter called "alternative recipe" which would be used if you didn't want a normal recipe.

To do this just leave out the ```material``` param and use the ```recipe``` param. Of course this won't produce a reverse recipe but it gives people the option to use unconventional recipes instead.

kilbith:
Why especially allow to unmount an hoe ? I don't like the idea... If a tool is crafted, it must be definitive.

Novatux:
Looks good

mods/farming/api.lua
```diff
@@ -83,10 +83,30 @@ farming.register_hoe = function(name, def)
 		end
 	})
 	-- Register its recipe
-	minetest.register_craft({
-		output = name:gsub(":", "", 1),
```
ShadowNinja: Instead of these `gsub`s, change the code so that the leading `:` is only added where needed.

If you did strip it after adding it `:sub(2)` would be much faster though.

MT-Modder: I just used the existing code. I've never used  gsub before and honestly have no clue what it does. Give me some time to look it up and fix this.

MT-Modder: Okay I did what you said and it seems to work fine:

```output = name:sub(2),```

So ```gsub``` replaces parts of strings but ```sub``` searches for parts of strings, right?

**Comment by MT-Modder** [comment]
How do you make these two commits one? I'm new to github. The only thing I know to do is delete this PR and start a new one.

**Comment by CraigyDavi** [comment]
> How do you make these two commits one? I'm new to github. The only thing I know to do is delete this PR and start a new one.

```
git rebase -i HEAD~2
(use fixup on latest commit)
git push --force
```

MT-Modder:
@CraigyDavi Thanks but I use the git extensions program. I don't know how to use the git bash.

I messed up the git extensions' tree so I'm gonna have to just delete and reopen..

PR #419, opened 2015-02-03
Clean up trapdoors code and make them more flexible

Jeija:
The trapdoors code was really weird before, with unneccessary `minetest.get_node(pos)` requests, variables that were defined but never used (`tmp_node2`) and things that, I guess, are mainly there because of historic reasons.

That's why I decided to clean up that piece of code and changed it so that  hird-party mods can now also use:

```Lua
doors.register_trapdoor("doors:trapdoor", {
	description = "Trapdoor",
	inventory_image = "doors_trapdoor.png",
	tile_front = "doors_trapdoor.png",
	tile_side = "doors_trapdoor_side.png",
	groups = {snappy=1, choppy=2, oddly_breakable_by_hand=2, flammable=2, door=1},
	sounds = default.node_sound_wood_defaults()
})
```

There might be bugs, please report them to me. I haven't tested this thoroughly enough, but it should work just the way the current trapdoors work and shouldn't break compatibility. That is unless any mod happens to depend on the variable `"state"` in the trapdoor node's meta. I don't believe that is the case, but maintaining compatibility there would be an easy fix.

### DIFF
--- a/game_api.txt
+++ b/game_api.txt
@@ -26,11 +26,17 @@ The bucket API allows registering new types of buckets for non-default liquids.
 
 Doors API
 ---------
-The doors mod allows modders to register custom doors.
+The doors mod allows modders to register custom doors and trapdoors.
 
-	doors.register_door(name, def)
-	^ name: "Door name"
-	^ def: See [#Door definition]
+doors.register_door(name, def)
+^ name: "Door name"
+^ def: See [#Door definition]
+ -> Registers new door
+
+doors.register_trapdoor(name, def)
+^ name: "Trapdoor name"
+^ def: See [#Trapdoor definition]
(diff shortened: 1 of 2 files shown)


C1ffisme:
Good idea.

0-afflatus:
Testing. So far so good. :+1:

BlockMen:
Please add ```doors.register_trapdoor()``` to the game api document(https://github.com/minetest/minetest_game/blob/master/game_api.txt) so it can get merged

Jeija:
Okay, @BlockMen, I made the changes you suggested.

BlockMen:
@Jeija thx
03c00a831d5c2fd37096449bee49557879068af1

PR #421, opened 2015-02-03
Add reverse recipes for hoes while keeping def.recipe compatibility v2

MT-Modder:
Reopened because I know very little on the use of git bash :)


MT-Modder:
Closing this for v3 (Added api info). #428

PR #422, opened 2015-02-04
Fix download URL

blha303:
Also: https://github.com/minetest/minetest/pull/2255


BlockMen:
60520b80321ae6066be208401f836505248334e9

PR #427, opened 2015-02-12
Make leaves climbable

Wayward1:
Adds "walkable = false," and  "climbable = true," to the leaves in nodes.lua

### DIFF
--- a/mods/default/nodes.lua
+++ b/mods/default/nodes.lua
@@ -422,6 +422,8 @@ minetest.register_node("default:leaves", {
 	visual_scale = 1.3,
 	tiles = {"default_leaves.png"},
 	paramtype = "light",
+	walkable = false,
+	climbable = true,
 	is_ground_content = false,
 	groups = {snappy=3, leafdecay=3, flammable=2, leaves=1},
 	drop = {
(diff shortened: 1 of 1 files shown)


C1ffisme:
I don't know about this...

PilzAdam:
I like this. It may break existing structures though, so I am against merging it.

kilbith:
Climbable makes sense but I wonder about non-walkable : those leaves are sometimes used as a hedge around homes for privacy matter.

@C1ffisme : not constructive at all.

Wayward1:
Hmm... Upon further consideration, it may be the same problem as #382- More realistic, but as @PilzAdam said, it would break existing structures.

C1ffisme:
Actually, I used to play on 0GB's server, somebody had a floor out of leaves, so maybe there is another solution to the climbing trees problem?

Wayward1:
Maybe it could be made toggleable in the minetest.conf or game.conf

CasimirKaPazi:
The main problem with this is when fancy leaves are turned off. You would see just a black screen when climbing.

C1ffisme:
@CasimirKaPazi Yeah I remember seeing that when PilzAdam did this in his subgame.

Novatux:
Closing, since it won't be added

PR #433, opened 2015-02-19
Update bucket mod's init.lua

tenplus1:
Changed empty bucket so it can be stacked and still be used without deleting remainder...


BlockMen:
Seems good to me

BlockMen:
df353603727e6084ae32b4c3d2c6e97f851f0ce2

PR #435, opened 2015-02-21
Add gunpowder trails

Wuzzy2:
This PR adds a graphical enhancement to the placed gunpowder. There is no gameplay change.
Now the gunpowder is drawn in gunpowder trails.

See here:
https://i.imgur.com/4niOc7B.png

I have drawn all the graphics which are neccessary for the raillike drawtype, both for normal gunpowder and burning gunpowder, which is animated.

In order for the normal gunpowder and the burning gunpowder to connect, I used the group “connect_to_raillike”.

Related legacy files have been removed.

Everything submitted here falls under the WTFPL.

Possible problems: Gunpowder will connect to EVERYTHING in the group “connect_to_raillike”, I guess I can't do much about it, its more like the engine does not provide a better solution here.
So it is not perfect.


C1ffisme:
This looks prettier than the last one.

MT-Modder:
:+1:  I love this idea.

Amaz1:
This is much better than the old one. :+1:

nerzhul:
sounds good for me too

LeMagnesium:
Look more realistic, I like this. :+1:
However, isn't there any issue with using the connect_to_raillike group? I mean, what happens if I place a rail node next to a gunpowder trail?

BlockMen:
Needs an engine change, since it would connect to rails or other nodes with same class

Wuzzy2:
There is no issue with gunpowder and normal rails, because rails do not belong the “connect_to_raillike” group.
But there can be an issue if the player adds a mod which adds a node which is member of the “connect_to_raillike” group.

But as long as you stay vanilla, there are no problems, because currently, only gunpower and burning gunpowder belong to that group.

But an engine change is needed anyways IMO.

C1ffisme:
@Wuzzy2 Most people don't play in just vannila. People tend to play with mods a LOT more than with a near featureless minetest. IMO, remove the connet_to_raillike group.

Wuzzy2:
Hmm, the thing is, if I remove the group, then burning gunpowder trails do not connect with non-burning gunpowder trails.
So either solution is suboptimal.

Maybe let's just wait for the engine change. lol

PilzAdam:
The carts mod overrides normal rails to use the connect_to_raillike group (it's needed for the acceleration rails), so this PR would make it (visually) incompatible with the carts mod.

BlockMen:
Please update this PR, the needed engine-change is done.
https://github.com/minetest/minetest/commit/b65c7ab4a43b082fd798eded3ca3c67fb66fad3f

Wuzzy2:
I suggest the following numbering scheme for minetest_game:

* rails = 2 (to be set later?)
* gunpowder = 3 (I have used this number)

I further suggest that 1 is reserved for “global use”, a generic category (applies to all mods in all subgames). To reflect the old behaviour of `connect_to_raillike` and it should also prevent surprises with old mods.

Note that all of this is purely convention so far.

Maybe we should create a wiki page where we collect the used raillike numbers so other modders can easily look stuff up.

MT-Modder:
The only issue I've found is stated in the above reference. I think it's an engine bug not a bug with this patch.

Wuzzy2:
I will update this PR as soon as https://github.com/minetest/minetest_game/issues/448 got merged.

I do not understand what you are saying here, MT-Modder.

@MT-Modder

MT-Modder:
@Wuzzy2 When a trail starts to burn the node turns direction. It depends on what direction the trail is going though. Here is a picture:
screenshot_18694133

Novatux:
Could you update it to work with https://github.com/minetest/minetest/commit/75622c30beb390b7ed578010a00390aa971b3951 please?

Wuzzy2:
I will update it after #511 got merged, because I have to touch `game_api.txt` twice.

Wuzzy2:
@Novatux: Done. This PR is now ready.

Novatux:
https://github.com/minetest/minetest_game/commit/0aebf864d1b57cc8f262da4ffe86b8b248bc4613

PR #443, opened 2015-03-02
Add vessels shelf

kilbith:
The vessels set should be completed by its own shelf which can store... the vessels :)

It goes well with the bookshelf for decoration (eg. in the kitchen).

vesselshelf


RHRhino:
:+1: This could be also useful for potion mods

MT-Modder:
What a great idea! Thanks Kilbith. :+1:

nerzhul:
:+1:

C1ffisme:
:+1:

Wayward1:
I like it!

C1ffisme:
Actually, I think the newest block is straw, but even then, straw is just added because it's dumb not to have it. I think it's about time we started going all 0.4.6 again, and adding a lot of new stuff. I want to see this and many more new blocks enter minetest_game.

Just because minetest_game is a modding base, dosen't mean we can't have some fun in it. And if it is a base for mods, then minetest should have more things to support the mods.

LeMagnesium:
:+1: for stocking vessels.
> This could be also useful for potion mods

 But if I read the code right, it couldn't stock potions, but it's not a problem for minetest_game itself.

kilbith:
@LeMagnesium It can store any node that has `vessel=1` in the node def.

LeMagnesium:
So it's even better, ready to be supported in potions mods.

C1ffisme:
Question : Don't you need to add your name and texture to the texture licenses?

And please, don't use a free-culture license. Just use Creative Commons.

LeMagnesium:
I don't know about you, but a shelf that could stock bottles and vessels is at its right place with them in the creative inventory : it stocks vessels, why putting it in default? Indeed, "why not putting it in default"? But it seems rather logic to put **vessels**'s shelfs in **vessels** mods.

C1ffisme:
True, true.

Ombridride:
I agree with Mg

C1ffisme:
Why does this have Low priority _and_ enhancement? Why not just put the label "low priority enhancement"?

Amaz1:
This is a brilliant idea! :+1: 
@C1ffisme If someone removed the vessels mod, and this node was in default, it would be pretty wierd for there to be something that was related to a removed mod to be in defualt... So I think this belongs to the vessels mod.
And regarding the last block that was added to mt_game, I think that would be beds, added 9 days ago!

C1ffisme:
Yeah, maybe it's just me.

Alright, let's just add this. We haven't done much block adding, anyway. Just because minetest_game is a modding base, dosen't mean that it has to be boring. This would improve minetest_game's fun.

MERGE IT! MERGE IT! MERGE IT!

BlockMen:
> MERGE IT! MERGE IT! MERGE IT!

Do you really think that would speed things up?

C1ffisme:
I don't know. I'm eager.

PilzAdam:
Why not use `group:vessel` in the crafting recipe?

0-afflatus:
:+1: add to vessels.

kilbith:
@PilzAdam Done.

PilzAdam:
@kilbith in `allow_metadata_inventory_move` you still use the hardcoded vessel nodes; use `group:vessel` there, too. Also remove the `vessels_nodes` table then.

kilbith:
@PilzAdam Done (again).

Novatux:
Looks good :)

PilzAdam:
You should use `minetest.get_item_group(stack:get_name(), "vessel") ~= 0` instead of `stack:get_name() == "group:vessel"`.

kilbith:
@PilzAdam Done.

BlockMen:
https://github.com/minetest/minetest_game/commit/3f2e35e8272ebc7a50a9e7af01ffc365558d3408

PR #445, opened 2015-03-04
Add ores support for singlenode

Gael-de-Sailly:
The ores aren't registered in singlenode. So we can't generate the ores with `minetest.generate_ores(vm)` in a custom mapgen.


PilzAdam:
`default.register_ores()` is a public function so custom mapgens can call it themselves if they want the default ores. It's also easier to not register them by default for custom mapgens that don't want the default ores.

Gael-de-Sailly:
Oh, of course ! So I've spent hours to review my code, and on IRC… for that ! xD

PR #446, opened 2015-03-05
Add Meselamps

BlockMen:
- Mese glows
- Glass + Mese = Meselamp
- New mese mineral texture


MT-Modder:
:+1:

C1ffisme:
:+1:

LeMagnesium:
:+1:
For all the people who don't want to download the patch to try it :

Mese lamp

C1ffisme:
It provides another light for modern builds!

I made my own meselamp once, but it had the mese block texture behind the obsidian glass. A little ugly, but it worked.

PilzAdam:
I am not sure about the mese ore / block to glow. Maybe add some kind of pre-processing before Mese starts to glow?

BlockMen:
@PilzAdam It might be more realistic to pre-process Mese before it can glow, but a game should always have some balance. A bad example for that missing balance is the break, that takes ages to bake. It might be realistic in comparision to ingame time and real time, but in comparision to ore melting its too long.

And since Mese is currently just a worse diamond its IMO nice giving it something special. So its a logical path from slightly glowing mese ore to a shiny mese-glass-lamp, without overcomplicated recipes to get that item.

Heybye44:
:+1:



—

PilzAdam:
@BlockMen sounds reasonable.

Btw, why do you change the mese ore texture?

BlockMen:
Because the current doesn't look good?

Novatux:
Looks good, but the lamp doesn't give much light... What about making it give the maximum possible light so it is possible to farm underground? Also, what about making it turn on and off when punched? (after all, mese is a magic material...)

HybridDog:
>mese is a magic material

You could make mese sharper than diamond.

ShadowNinja:
`meselamp` -> `mese_lamp`

I like the idea of making mese glow and adding a lamp, but:
  * I'm not sure about the new ore texture, I'll have to check it.
  * I think that it should have the maximum light level.

SmallJoker:
To solve the glowing problem: Add another item to the crafting recipe. Mese has chemical properties.

I would like to see this node in the MTG.

paramat:
Good PR yes please. I often consider we need our own 'glowstone' and it's a good use of the already bright yellow mese. I agree we need a new mese mineral texture.

PilzAdam:
What happens to mese blocks in existing maps? Light isn't recalculate when loading a block, right?

HybridDog:
>What happens to mese blocks in existing maps? Light isn't recalculated when loading a block, right?

l guess that already generated mese blocks don't look weird but l know how a generated glowing ore looks in a cave. The light isn't spread around to light the cave:
glow1
after placing a block next to it (e.g. a disabled green bobblock), the light spread becomes updated:
glow2

HybridDog:
BTW: fullbright mese which is not a light source looks good to me, too:
fbmese

paramat:
> l know how a generated glowing ore looks in a cave. The light isn't spread around to light the cave:

I had that problem too until hmmmm saw the mistake in my node definition, any node that is a light source needs:
paramtype = "light"

HybridDog:
>any node that is a light source needs:
paramtype = "light"

thanks

PilzAdam:
@BlockMen what about my previous comment?

>What happens to mese blocks in existing maps? Light isn't recalculated when loading a block, right?

BlockMen:
@PilzAdam didn't you read paramats comment?

PilzAdam:
@BlockMen I did, it is about a different issue (@paramts comment was about mapgen, my concern is about already existing blocks being loaded again).

HybridDog:
@PilzAdam you could find it out yourself

BlockMen:
@PilzAdam oh. Well, whith such a small amout of light emitted it seems to be no problem.
Al least i can't really see any difference between updated and not updated light (see img(s) below):
mese_glow

PR #448, opened 2015-03-06
Use connect_to_raillike for default:rail

Wuzzy2:
This tiny PR adds the group `connect_to_raillike=2` to `default:rail`.
The rationale behind this is to make it easier for other mods to extend on this without having to overwrite `default:rail`.

Related:
https://github.com/minetest/minetest/commit/b65c7ab4a43b082fd798eded3ca3c67fb66fad3f
https://github.com/minetest/minetest_game/pull/435


PilzAdam:
Document the used values in game_api.txt maybe? (WIth a hint that `minetest.registered_nodes["default:rail"].groups.connect_to_raillike` should be used instead of hard coding the value in mods)

Wuzzy2:
Okay, it is documented.
However, I disagree with that modders should use `minetest.registered_nodes["default:rail"].groups.connect_to_raillike`.
First, this is a mouthful, second, the choice of `default:rail` becomes arbitrary as soon as minetest_game itself adds further rail types. If you take gunpowder trails into account.

What do you think of this alternative idea?: Just add (and document) some global variables, like this:
```
default.connect_to_raillike.rails = 2
default.connect_to_raillike.gunpowder = 3 -- if gunpowder trails get merged
```

PilzAdam:
@Wuzzy2 
The choice is not arbitrary, since you use that group to connect your rails to the default rails, so you want to be sure that you have the same group rating as the default rail.
You can always break the API, this is not a valid argument.
Moving this to global variables would separate nodedef fields from the nodedef tables (as it was previously done with WATER_VISC).

Wuzzy2:
> Moving this to global variables would separate nodedef fields from the nodedef tables (as it was previously done with WATER_VISC).

How about this then?:
```
default.connect_to_raillike.rails = 2
```
(or whatever style you'd prefer)

Later, in the node definiton of `default:rail`:
```
groups = { --[[ other groups ..., ]] connect_to_raillike = default.connect_to_raillike.rails },
```

Novatux:
Something like default.connect_to_raillike.<groupname> seems to be the best idea for me, or perhaps what was suggested in https://github.com/minetest/minetest/issues/2385 in builtin could be good too.

ShadowNinja:
Something like the following needs to be added to builtin to make this easier for everybody and prevent collisions:
```Lua
do
	local ids = {}
	local cur_id = 0
	function core.raillike_group(name)
		local id = ids[name]
		if not id then
			cur_id = cur_id + 1
			ids[name] = cur_id
			return cur_id
		end
		return id
	end
end
```
And of course it has to be documented in `lua_api.txt`.

Wuzzy2:
Okay, I have now the changed the way how the connect_to_raillike numbers are used. I did is as described in my previous post.


@ShadowNinja: Feel free to open a new PR/issue.  :P

Novatux:
Could you update it to work with https://github.com/minetest/minetest/commit/75622c30beb390b7ed578010a00390aa971b3951 please?

Wuzzy2:
@Novatux: Done. WARNING: I have messed up this branch, don't merge it. Merge #511 instead.

PR #450, opened 2015-03-07
Make TNT respect on_blast, implement on_blast for some nodes

Wuzzy2:
This PR alters the way how TNT destroys nodes in a way which I think is much saner and more mod-friendly.

It makes TNT check first whether the node which is about to be destoyed implements `on_blast`, a function which is supposed to be called when explosions occour (see `lua_api.txt`). If this function exists, this function will be called. If this function does not exist, the TNT reverts to the default behavior, which is destroying the node with `minetest.remove_node`.
This should now finally make it possible to define nodes which are able to “defend themselves” against TNT explosions.

While I was at it, I also implemented on_blast for the following nodes:
- Steel Door: Ignores explosion (meaning that it does not get destroyed)
- Locked Chest: Ignores explosion
- Fire: Ignores explosion
- TNT: Starts burning
- Burning TNT: Ignores explosion
- Gunpowder: Starts burning
- Burning Gunpowder: Ignores explosion

This also replaces some of the destruction logic from the TNT node itself, it now only checks against air directly.

### DIFF
--- a/mods/doors/init.lua
+++ b/mods/doors/init.lua
@@ -25,6 +25,9 @@ function doors.register_door(name, def)
 	if not def.sound_open_door then
 		def.sound_open_door = "doors_door_open"
 	end
+	if def.only_placer_can_open then
+		def.on_blast = function() end
+	end
 	
 	
 	minetest.register_craftitem(name, {
@@ -173,7 +176,8 @@ function doors.register_door(name, def)
 		
 		can_dig = check_player_priv,
 		sounds = def.sounds,
-        	sunlight_propagates = def.sunlight
+		sunlight_propagates = def.sunlight,
+		on_blast = def.on_blast,
 	})
 
 	minetest.register_node(name.."_t_1", {
@@ -205,7 +209,8 @@ function doors.register_door(name, def)
 		
 		can_dig = check_player_priv,
 		sounds = def.sounds,
-        	sunlight_propagates = def.sunlight,
+		sunlight_propagates = def.sunlight,
+		on_blast = def.on_blast,
 	})
 
 	minetest.register_node(name.."_b_2", {
@@ -237,7 +242,8 @@ function doors.register_door(name, def)
 		
 		can_dig = check_player_priv,
 		sounds = def.sounds,
-        	sunlight_propagates = def.sunlight
+		sunlight_propagates = def.sunlight,
+		on_blast = def.on_blast,
 	})
 
 	minetest.register_node(name.."_t_2", {
@@ -269,7 +275,8 @@ function doors.register_door(name, def)
 		
 		can_dig = check_player_priv,
 		sounds = def.sounds,
-        	sunlight_propagates = def.sunlight
+		sunlight_propagates = def.sunlight,
+		on_blast = def.on_blast,
 	})
 
 end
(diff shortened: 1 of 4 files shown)


PilzAdam:
:+1:
Maybe but a comment in the empty on_blast functions, to clarify that it means that they are not affected by explosions.

LeMagnesium:
:+1: It allows us to protect some nodes which souldn't be destroyed, be randomly destroyed or stay saved.
Also, when a chest is exploded, it should drop its content to the ground, with `minetest.add_item`, to keep the items placed being so players could get them back.

C1ffisme:
:+1: 

What about obsidian and default:cloud? They could ignore the explosion, just to give players and server moderators a way to protect their builds.

Novatux:
Looks good, but some normal nodes (I mean, nodes you can build with) could be protected against blast (perhaps nodes such as some metal blocks would be a good idea).

C1ffisme:
Yeah, or perhaps each metal/obsidian block could have their own destruction scripts with a chance wheather or not they blow up.

Wuzzy2:
Comments have been added.
There does not seem to be a consensus about what to do with other nodes. Therefore, I won't add any further on_blast functions. If anyone insists, it can be done in a new PR. :P

mods/tnt/init.lua
```diff
@@ -82,7 +82,14 @@ local function destroy(drops, pos, cid)
 	if def and def.flammable then
 		minetest.set_node(pos, fire_node)
 	else
-		minetest.remove_node(pos)
+		local nodename = minetest.get_node(pos).name
+		local on_blast = minetest.registered_nodes[nodename].on_blast
```
ShadowNinja: Use the `cid_data` cache for this to prevent the expensive lookups.

sfan5:
:+1:

Wuzzy2:
@PilzAdam and @ShadowNinja: Done.

Novatux:
You should make on_blast destruct the other part of the object for multi-node objects (doors, beds, etc.) too; otherwise, looks good.

Wuzzy2:
@Novatux: Done.

I have added `on_blast` for doors. When one piece of the door gets its `on_blast` called, both parts are destroyed. There are no drops.

Beds seem to already work out of the box, because they use `on_destruct`. Implementing `on_blast` seems to be redundant here. I have also tested it, blasts seem to always destroy the whole bed.

There are no other multi-node objects in Minetest Game as far I know.

Novatux:
https://github.com/minetest/minetest_game/commit/8bc8dd64c56b6c5a7350fb7504864acf116f447f

mods/doors/init.lua
```diff
@@ -108,6 +108,33 @@ function doors.register_door(name, def)
...
+					pos.y = pos.y + 1
+					check_and_blast(pos, name..other_door_type)
+				end
+			elseif door_type == "_t_1" or door_type == "_t_2" then
+				return function(pos, intensity)
+					check_and_blast(pos, name..door_type)
+					pos.y = pos.y - 1
+					check_and_blast(pos, name..other_door_type)
+				end
+			end
```
ShadowNinja: This should use `elseif` rather than an `else if then end` chain.
Also, you can add a `step` value of `1` or `-1` instead of copy-pasting the function.

PR #456, opened 2015-03-11
play a sound if someone tries to open a locked chest that doesn't belong to him/her/it

HybridDog:
sound file from http://freesound.org/people/nhaudio/sounds/192416/ (CC0), edited with audacity
https://github.com/minetest/minetest_game/pull/456/files?w=1


kilbith:
Basically agree with the idea but :-1: with the sound itself, that has nothing to do with a latch sound.

SmallJoker:
This is like the toilet from homedecor. People love to make sounds and those can get very noisy.

HybridDog:
>People love to make sounds and those can get very noisy.

as far as l know this is realistic

PilzAdam:
Why only a sound for this case? Chests should also have sounds when normally opening them.

HybridDog:
>Chests should also have sounds when normally opening them.

but when someone opens the chest it looks closed

tenplus1:
Realistic would be if owner opens chest, chest looks opened, and other players can take stuff while chest is open :))))

PR #460, opened 2015-03-14
Add charcoal and notification of bone coordinate

Rui914:
Note: This is not a good texture


MT-Modder:
I like it :+1: Texture looks good to me.

C1ffisme:
:+1:  And the texture actually looks pretty good!

However, you need to split your pulls. Have one pull be to add charcoal, and another to notify bone coords.

Actually, do not add a pull for bone coordinate chatting, that would be a duplicate of another pull.

Rui914:
#463

PR #462, opened 2015-03-14
Keep a bucket of liquid in creative mode

Rui914:


C1ffisme:
Hold on, make it use proper indentation. The return needs to be moved up with a Tab.

Rui914:
Fixed.

PilzAdam:
Whats the usecase for this? You can just get the liquid source blocks if you want to place them a lot.

BlockMen:
@PilzAdam but why then having buckets at all in creative mode? Should get added IMO

PilzAdam:
@BlockMen maybe to remove liquids, so picking up liquids in creative could leave the bucket empty.

PR #463, opened 2015-03-14
Add Charcoal

Rui914:
split


kilbith:
:-1: Redundant with the coal.

C1ffisme:
Anyway, :+1: .

Calinou:
Charcoal was removed in early 2012 because celeron55 found it made the game too easy.

Rui914:
This is I think not detract game balance.
But, if you are a negative, I close this request.

C1ffisme:
@Calinou I don't think so. Players need to get torches REALLY early into the game. It's terrible having to search for a cliff face to mine coal off.

However, if this does make the game too easy, then you can change charcoal around. Perhaps it should have less burntime. Also, blocks of coal don't use group:coal, I believe. If charcoal can't make coal blocks, then you have another disadvantage of charcoal.

Or maybe charcoal makes the game too easy because there isn't much game after mining and getting diamonds.

As I've said many times, minetest needs goals.

HybridDog:
>It's terrible having to search for a cliff face to mine coal off.

In minetest you can dig straight down, the sunlight shines into your mine.
When you want to get back, you make a shift stopper and type /home.

C1ffisme:
> In minetest you can dig straight down,

Um, hello? 
screenshot_1745016150
Remember about this stuff? That dense, hard to swim through, white hot stuff called lava? And then there are those things called bones. They drop when you die in this stuff.

HybridDog:
>That dense, hard to swim through, white hot stuff called lava?

you need to type /home fast and you don't take all your valuable stuff with you when mining

LeMagnesium:
To go back to the subject, do you know how much time it takes for me to craft my first torches when I just joined a new game in survival mode without any privilege other than interact and shout? About 3 minutes in fact. Even if it's night, all you have to do is find the first tree you see, dig it completly, craft a wooden pickaxe, and look around all the caves' entries you have, or even plain stone, and you will find coal very fast. If you want, you can even craft a stone pickaxe which will give you more time to find your first coals. If charcoal is added just to let us craft torches faster, then the game becomes too much easy, and I don't want that.

C1ffisme:
You could, as I said, add advantages for coal and remove advantages for charcoal. Plus, it takes a lot of time to dig a tree with your hand, so getting a lot of charcoal is hard!

Why not make coal deeper underground, so you have to make charcoal, and then once you have hit coal you can use that.

And like I said, make charcoal a quicker consumed material.

LeMagnesium:
> Plus, it takes a lot of time to dig a tree with your hand
About 30 seconds.

If you add charcoal, keep coal as it is now, and make charcoal less useful than it is now (no big burntime, etc). I disapprove adding it, but other people may want it.

thatgraemeguy:
Not really sure what the point of adding charcoal is? How is it significantly different from coal?

Rui914:
Become easier to at making such as torches.
But, cannot make coal block.

thatgraemeguy:
I just started a fresh world and managed to make 12 torches in under 2 minutes. That's pretty easy as-is.

:-1:

Rui914:
OK.
This request is not required.

Megaf:
I believe this should be reopened and and merged.
@ShadowNinja @Zeno- @PilzAdam 

It would in fact facilitate players lives and make the game more fun.

thatgraemeguy:
Facilitate in what way? I still can't see what this adds over just plain coal.

Megaf:
@thatgraemeguy 
1. It's logical to burn wood to get coal.
2. It removes the need of having to make a pick axe to get coal.
3. Making crafting of torches much easier.

I has happened already to me, I was inside a cave and just ran out of torches and coal and tools, so I couldn't get anything else because I had only wood.

Instead of discussing why we should have coal from wood, can you give a strong point on why we shouldn't have?

It doesn't make the game heavier, slower nor more complex. It doesn't introduce any regression. It's easy to revert, so why not add it?

0-afflatus:
It should be a separate mod

HybridDog:
>It should be a separate mod

lf l remember correctly there are already lots of  charcoal mods.

BlockMen:
Added in Minetest NeXt.
https://github.com/BlockMen/minetest_next/commit/1f3aebd430db986c4287101c0db3a3e2dbf8bb84

C1ffisme:
I think that adding charcoal might be useful if you are running a sky island server. Where you can't mine coal. This creates a way for people with limited resources to make torches.

Besides, why should something like torches be so hard to get? It's not like there are any mobs anyway. Torches just mean you don't have to turn your brightness level up.

HybridDog:
>I think that adding charcoal might be useful if you are running a sky island server. Where you can't mine coal. 

yes, and leaves rot to dirt instead of disappearing when decaying

>Besides, why should something like torches be so hard to get?

It's very easy to get torches, just dig a few trunks, craft a wooden pickaxe and mine a bit down.

C1ffisme:
@HybridDog Exactly. If something like this is so easy to get, then why is it suddenly controversial to add another method to get it. It's like not letting someone use lava and water to make stone. It gets extremely common after a while, so one other method doesn't matter.

If you really have a problem with this easy way of getting coal, then make it less powerful than coal, but don't not add a feature because it "makes the game too easy" at the very start. (Hint: That's when the game is SUPPOSED to be easy. Is coal supposed to be rare, and diamond supposed to be everywhere?)

HybridDog:
l think the goal of the game is to find out new nodes and their functions, etc.
but not to slowly get more and more valuable items that the previous ones (stone pick) can be trashed

ShadowNinja:
FWIW I've added basically the same thing to one of my servers (coal has to be renewable on it), althoug I just made it convert to coal directly since I didn't want to modify mt_game or spend much time on it.
It doesn't really make the game much easier IMO because if you're mining enough to have decent tools (or build anything advanced like the Technic stuff) you'll have plenty of coal, and you only really need coal for torches.  You just need to find some stone and build a furnace instead of finding some coal.  It might be slightly easier at the start of the game, but coal soon becomes a non-issue.

PR #466, opened 2015-03-17
Add Furnace particle

Rui914:
Coded:Hybrid Dog
Code:https://forum.minetest.net/viewtopic.php?f=47&t=4668&start=1725#p172778

Note: This is not a good texture


C1ffisme:
:+1:

LeMagnesium:
Are you sure it doesn't create lag?
If it doesn't then why not adding this, but since particles are still a bit laggy it should be tested.

C1ffisme:
@Rui914 Why did you close this? I think you should comment your reasons before closing.

Rui914:
After resolving the lag problems, to open.

HybridDog:
the "Small Fix" tries to register a new abm every time the usual abm's function is run and causes an error (because of the end which…)

Rui914:
Maybe Fixed.

BlockMen:
Particles add too much traffic for such an unimportant effect. Shouldn't be added

PilzAdam:
@BlockMen 704 bytes (including the UDP header) per 2 seconds per furnace are too much? (I assume that the 1.5 interval is equal to 2, since only integers are allowed).
Player movement is 240 bytes (including UDP header) which is sent each step (with `dedicated_server_step = 0.1` by default). So a player, by simply moving, causes almost 7 times the traffic a furnace would cause with these particles.

BlockMen:
@PilzAdam Yes, at least when there are more than 5 active players. Particles send on a second (or 2 second) loop to each client can slow down a server a lot. And that we add these packets for a minimal visual change is not worth it.

nerzhul:
Network is not a problem here. Packet sending is efficient. If particle add lag , please look at the spawner algorithm

C1ffisme:
@PilzAdam 70% of a kilobyte? My goodness, there are text files smaller than that. And several of those per furnace... 

What happens if after several (real-time) days of mining, I light up my 16 furnaces all at once? I'd rather spend only 1 kilobyte of memory on that, rather than have possibly hundreds of furnaces generating, almost a megabyte...

PilzAdam:
@C1ffisme are you sure you understand what you are talking about?
See my previous post about player movement, to have a reference.
Also you'd need several thousand furnaces to generate a megabyte of traffic, not hundreds.

C1ffisme:
@PilzAdam True, but thinking about the future, there are minecraft servers with hundreds of people. And I did say almost.

PR #471, opened 2015-03-20
Alter mintest_game/mods/default/nodes.lua

tenplus1:
Locked chests are checked for ownership when right-clicking to enter and trying to destroy, it is not needed in inventory_move, inventory_put and inventory_take so will save a little server time checking...

### DIFF
--- a/mods/default/nodes.lua
+++ b/mods/default/nodes.lua
@@ -1196,24 +1196,12 @@ minetest.register_node("default:chest_locked", {
 		return inv:is_empty("main") and has_locked_chest_privilege(meta, player)
 	end,
 	allow_metadata_inventory_move = function(pos, from_list, from_index, to_list, to_index, count, player)
-		local meta = minetest.get_meta(pos)
-		if not has_locked_chest_privilege(meta, player) then
-			return 0
-		end
 		return count
 	end,
     allow_metadata_inventory_put = function(pos, listname, index, stack, player)
-		local meta = minetest.get_meta(pos)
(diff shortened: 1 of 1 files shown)


PilzAdam:
I'm not sure that this is safe this way. The server does not know whether a client is displaying a formspec, so malicious clients may send the inventory actions without opening the chest via a right click first.

Sokomine:
The checking for ownership on rightclick is a very annoying "feature". The client can no longer just display the stored formspec but has to request it from the server. Can't you add secret chests instead? Chests people who want to hide their inventory can use, but which can be ignored by players who want to take a look at their chests fast without having to resort to unprotected open chests?

HybridDog:
Are secret chests locked chests with a bolt-on lock?

C1ffisme:
Maybe we could have "Glass Chests" which you can see inside.

kaeza:
Indeed, this is not be safe. It's like a web server assuming the client is a well-intentioned web browser that will play by the rules.

HybridDog:
>It's like a web server assuming the client is a well-intentioned web browser that will play by the rules.

for example you don't have to login if you want to get that sound:
http://freesound.org/people/jspath1/sounds/125431/

just see the address of the preview sound and there replace lq with hq because of guessing lq means low quality and hq…
http://freesound.org/data/previews/125/125431_2154289-hq.mp3

sofar:
These checks need to remain - I stumbled upon the requirement as I was coding up a chest "sharing" idea, and without these checks there's no way to safely do things.

PR #486, opened 2015-04-07
Convert beds in mesh

kilbith:
Fix (partly) https://github.com/minetest/minetest_game/issues/451

zqdq


PilzAdam:
Your code does nothing to prevent placing nodes in the top half of the bed, see https://github.com/minetest/minetest_game/issues/451#issuecomment-88100363
Also there is a bug in your code that places 2 beds (the bottom of the second one in the top of the first one).

I am against this.

kilbith:
@PilzAdam Against the idea or the current code ? In the first case it'd be surprinsing to refuse fixing a bug with the screwdriver. There are some other benefits besides (looking, fastness, maintainability).

PilzAdam:
@kilbith This is not a proper fix for #451 (see my comment on that issue). You still need to maintain a node that blocks the top half of the bed, so it doesn't matter if thats an invisible node or if it's used as part of the bed.
Even if you would add an invisible node to this to block the top half, you'd still need to rotate it when using a screwdriver, so this doesn't help at all.

kilbith:
@PilzAdam That prevention should be the purview of the engine's handling (eg. based on the collision_box).

EDIT : found a solution (see my comment below).

C1ffisme:
Okay, this has negative comments, however, is this does get merged, I think we should use b3d instead of obj, since the player model uses it. (Just a consistency thing.)

kilbith:
@PilzAdam PR updated. The simplest bypass for prevent the node placing is to enlarge the selection box to 2-nodes max. size.

@C1ffisme Your hollow comment just prove you have never dealt with Blender. Blitz3D excels only for the animated models, but it's always heavier than .obj for the static ones. Also I don't care at all of your usual noise. Go ahead.

PilzAdam:
@kilbith Thats not enough. Anything other than direct placement of the player can still place nodes inside of the bed. Also a too large selection box doesn't look good.

C1ffisme:
@kilbith Rather rude.

I'm just making a suggestion. Are you saying that game consistency comes before a grudge?

BlockMen:
IMO a better fix for #451 would be a "no_rotation" group that is checked by the screwdriver to prevent rotating.
And PilzAdam is right, this way other nodes could be placed in theupper half of the bed.

kilbith:
@PilzAdam @BlockMen OK for closing the PR then ?

kilbith:
OK closed. @PilzAdam Hope you'll act reactively in the engine for that callback as more than blasting 2 PRs successively that reduce problems overall...

PR #488, opened 2015-04-09
Fix screwdriver and beds

HybridDog:


PilzAdam:
I'd call this workaround hacky.

A better way would be a on_rotate callback in the nodedef of beds / doors and calling this from the screwdriver if present.

HybridDog:
but that wouldn't fix pushing beds with a mesecons piston

HybridDog:
sorry that l touched your bed

HybridDog:
l forgot the screenshot
HELLOIFYOUREADTHISTHESCREENSHOTDIDNTLOADYET

HybridDog:
and l forgot to reference this https://github.com/minetest/minetest_game/issues/451

Issue #469, opened 2015-03-18
Small improve of the Flowers minetest_game mod


Ombridride:
Small improve of the Flowers minetest_game mod, the code is more smaller and do the same thing.

If you want it, take it (the most important changes are from Davedevils who makes a very well optimisation work) : 
https://github.com/Ombridride/minetest-minetestforfun-server/blob/master/minetestforfun_game/mods/flowers/init.lua

C1ffisme:
If you have code, then make a pull request. (Unless you have a specific reason for not making a pull request, which you should describe in the issue, rather than telling a commenter afterwards.)

LeMagnesium:
@C1ffisme , @Ombridride didn't create a pull-request because he wanted to know if people where interested, but it's possible in a pull-request. I think he should have created a pull-request with the isolated changed code too. Maybe he had not time to apply his (@davedevils in fact) changes over the actual flowers mod.

C1ffisme:
Ah, okay.

Ombridride:
A minor bug with dye was fixed, this code in now ready for a Pull Request, i close this issue ticket

PR #492, opened 2015-04-15
Improve flowers mod

Ombridride:
Improve of the Flowers minetest_game mod, the code is more smaller and
do the same thing. 

We fix a small mistake 2 weeks ago, now the code is clean and 100% functional.

We speak about this changes here : https://github.com/minetest/minetest_game/issues/469

(PS : the most important changes are from Davedevils who makes a very well optimisation work, the fixs are by the devs of MinetestForFun)

### DIFF
--- a/mods/flowers/init.lua
+++ b/mods/flowers/init.lua
@@ -1,9 +1,6 @@
 -- Minetest 0.4 mod: default
 -- See README.txt for licensing and other information.
 
--- Namespace for functions
-flowers = {}
-
 -- Map Generation
 dofile(minetest.get_modpath("flowers").."/mapgen.lua")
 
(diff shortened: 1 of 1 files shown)


PilzAdam:
* The new chance for the ABM should be 625 to have the same regrowing speed, if I'm not mistaking
* Don't have all these `if then return end` statements in one line, keep them in 3 lines
* Why don't all flowers use the new function to be registered?
* Use English comments (`Fleur Simple (une case)`)
* Remove the logging at the end of the mod
* Adding new flowers (lily pad) should be in a separate pull request

LeMagnesium:
@PilzAdam , there are indeed a few mistakes from Ombridride's code, mainly because it's a draft written by @davedevils for his server, and because I have not took time yet to improve it. I will remove useless things, and change what you pointed. Thanks for your comment.

Ombridride:
When we have more times, we will improve this code just for MT_game...

LeMagnesium:
There it is : #494
I signed the commit as @Ombridride (I just need his mail) and solved the following problems : 
 - I put back the actual chance/interval in the ABM
 - I unwrapped lines containing returns
 - All flowers use the new function
 - Comments are in english
 - The logging and the lily pad have been erased

Novatux:
Closing, since it is replaced by #494

PR #499, opened 2015-04-30
Optimize all sounds

kilbith:
Same length durations, optimized : `519 KB` -> `379 KB` (140 KB less).


Calinou:
What software did you use to optimize sounds? Did you just reduce their bitrate?

kilbith:
> What software did you use to optimize sounds?

Audacity.

>  Or did you just reduce their bitrate?

Their bitrate is exactly the same.

tenplus1:
Lossy format being converted back to lossy format will indeed be smaller because some of the audio quality has been reduced further...

kilbith:
You would not notice a quality loss. Please test and compare before any comment.

sfan5:
IMO a small size reduction like this does not warrant the quality loss caused by re-encoding lossy sources.

kilbith:
no interest, closing.

HybridDog:
l would like a setting in minetest which disables sound preloading (only works in singleplayer mode).

PR #501, opened 2015-05-06
allow disabling the vm:update_map() after spawning trees

HybridDog:
When l updated my swamps mod, which generates many trees in a mapchunk, l found out that updating the map only once but for the whole mapchunk works faster than using update_map for each tree.
https://github.com/HybridDog/sumpf/blob/master/sumpf/mapgen.lua#L164
l thought, if mods want to generate e.g. default pine trees, they should have the possibility to do this, too.

### DIFF
--- a/mods/default/trees.lua
+++ b/mods/default/trees.lua
@@ -125,7 +125,7 @@ end
 
 -- Appletree
 
-function default.grow_tree(pos, is_apple_tree, bad)
+function default.grow_tree(pos, is_apple_tree, bad, disable_map_update)
 	--[[
 		NOTE: Tree-placing code is currently duplicated in the engine
 		and in games that have saplings; both are deprecated but not
@@ -152,12 +152,14 @@ function default.grow_tree(pos, is_apple_tree, bad)
 
 	vm:set_data(data)
 	vm:write_to_map()
-	vm:update_map()
+	if not disable_map_update then
+		vm:update_map()
+	end
 end
 
 -- Jungletree
 
-function default.grow_jungle_tree(pos, bad)
+function default.grow_jungle_tree(pos, bad, disable_map_update)
 	--[[
 		NOTE: Jungletree-placing code is currently duplicated in the engine
 		and in games that have saplings; both are deprecated but not
@@ -203,7 +205,9 @@ function default.grow_jungle_tree(pos, bad)
 
 	vm:set_data(data)
 	vm:write_to_map()
-	vm:update_map()
+	if not disable_map_update then
+		vm:update_map()
+	end
 end
 
 -- Pinetree from mg mapgen mod, design by sfan5, pointy top added by paramat
@@ -220,7 +224,7 @@ local function add_snow(data, vi, c_air, c_ignore, c_snow)
 	end
 end
 
-function default.grow_pine_tree(pos)
+function default.grow_pine_tree(pos, disable_map_update)
 	local x, y, z = pos.x, pos.y, pos.z
 	local maxy = y + random(9, 13) -- Trunk top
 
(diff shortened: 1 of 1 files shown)


C1ffisme:
Have you actually tested this? Is it faster?

HybridDog:
What do you mean?

C1ffisme:
Have you tried this with your current code? As in, with this pull request and not just your swamp mod?

LeMagnesium:
Works for me :+1:
If you want to test, apply this over trees.lua : http://pastebin.ubuntu.com/11026319/

HybridDog:
>Have you tried this with your current code? As in, with this pull request and not just your swamp mod?

l didn't test if it can improve speed yet.

LeMagnesium:
> l didn't test if it can be used to improve speed. But l think there's no big difference to the swamp trees spawning.

I didn't notice anything big, but when generating on a new map about 250 square nodes I didn't notice any lag, nor any difference between the moment the ground is generated and the trees appear.

paramat:
HybridDog, you shouldn't use update_map in your swamps mod, that's for non-mapgen object voxelmanip only and is much slower.
The tree growing functions in default are there for single trees grown from saplings long after mapgen, and therefore use the non-mapgen voxelmanip and are therefore slow. If someone wants to grow many trees they should not use the default functions but copy the lua code into their own mod, use a suitable voxelmanip and update the map once.
Your implementation is also incorrect, so this PR should not be merged.

HybridDog:
>You are missing calc_lighting() in your use of the mapgen-object voxelmanip, whch is probably causing the lighting problems you are then fixing with update_map().

when l decided to use the update_map l think using calc_lighting didn't work
l need to test it

>The tree growing functions in default are there for single trees grown from saplings long after mapgen, and therefore use the non-mapgen voxelmanip and are therefore slow. If someone wants to grow many trees they should not use the default functions but copy the lua code into their own mod, use a suitable voxelmanip and update the map once.

ok, sorry, l thought these tree functions were used for mapgen.

HybridDog:
>l need to test it

l did it and it didn't work, there are no tree shadows:
screenshot_20150508_194440

paramat:
I think that's because you need to generate the trees in the on-generated function, not after.

HybridDog:
>I think that's because you need to generate the trees in the on-generated function, not after.

l did it, at first the ground is set and written to map, then the trees are set with own vm objects but without update_map and after that the light is calculated.

paramat:
> lf l used the mapgen-object voxelmanip, the trees would be cut at the chunk corners, wouldn't they?

No =) the MOVM extends 16 nodes beyond the mapchunk, it overlaps other mapchunks, so any structure placed can extend up to 16 nodes into the neighbour mapchunk without being chopped.

HybridDog:
the trees are 21 nodes high

paramat:
Ah i just remembered something ... for lighting to work using the MOVM you need to 'set_lighting)' to zero then 'calc_lighting()', like this:

	local vm, emin, emax = minetest.get_mapgen_object("voxelmanip")
	local area = VoxelArea:new{MinEdge=emin, MaxEdge=emax}
	local data = vm:get_data()
...............
	vm:set_data(data)
	vm:set_lighting({day=0, night=0})
	vm:calc_lighting()
	vm:write_to_map(data)

paramat:
> the trees are 21 nodes high

No problem as long as the trees are not placed on ground higher than 47 + 16 - 21 = 42. Mgv6 terrain almost nver gets that high, and swamps are not at high altitude anyway, so no problem.

HybridDog:
https://github.com/HybridDog/sumpf/blob/master/sumpf/mapgen.lua#L208
they're generated higher and should support the ethereal mod

HybridDog:
>maybe l could update the map only for trees on the upper chunk corners

l think about another solution: l could make the trees in the top part have a lower height.

Issue #500, opened 2015-05-03
[Request] TNT API


Rui914:
The function of TNT can use in any other mod.
・Example
```
tnt.explode
tnt.boom
tnt.register_tnt(name,description,tiles,radius)
```

ShadowNinja:
[My nuke mod](https://github.com/ShadowNinja/nuke) has an API, and the TNT mod is based on it.

Heybye44:
They should merge it and add server protection if it doesn't have it. :+1:

Rui914:
Also, the explosion function of TNT also will should be to the API.

C1ffisme:
Maybe functions like: ```tnt.explode``` and ```tnt.boom```.

Actually, I think these functions already exist, lemme check the code...

Yeah, they're all there, but they're all local. And a ```tnt.register_tnt(name,description,tiles,radius)``` would be nice.

C1ffisme:
Maybe functions like: ```tnt.explode``` and ```tnt.boom```.

Actually, I think these functions already exist, lemme check the code...

Yeah, they're all there, but they're all local. And a ```tnt.register_tnt(name,description,tiles,radius)``` would be nice.

LeMagnesium:
I made this modification, it's not something that complicated, and I feel like the issue list is growing without issues being closed/solved, so if it helps :  #502

LeMagnesium:
`x_range` and `y_range`  would be enough (having x and z being different would lead to very strange shape). I will do it later, when #502 is closed or merged.

ShadowNinja:
> A tnt.register_tnt(name,description,tiles,radius) would be nice. [...] mese TNT [...]

We might as well just rename my `nuke` mod and merge it (without missiles).
I rewrote TNT to basically just be a minimal version of that mod with some slower things like drops (because it doesn't support big explosions).

C1ffisme:
I never knew you had missiles in your nuke mod!!! I wish I could get this kind of news...

Anyway, if you have x,y and z ranges, then you could make elipses (egg shapes).

LeMagnesium:
> then you could make elipses (egg shapes)
For eggs you only need x (because x = z) and y.

C1ffisme:
Yes, but say you wanted to make one that was shaped like a real egg, not like a flattened ball?

LeMagnesium:
![tnt blow](https://lut.im/v7pheSzq/AQmJhtD1)
[...]

PR #502, opened 2015-05-08
Created TNT mod's namespace

LeMagnesium:
- Created a namespace called `tnt` for the TNT mod, containing
   variables and functions of this mod to be used in other mods

(Was created for issue #500)

### DIFF
--- a/mods/tnt/init.lua
+++ b/mods/tnt/init.lua
@@ -7,13 +7,16 @@ if (not singleplayer and setting ~= true) or
 	return
 end
 
+-- Global TNT namespace
+tnt = {}
+
 -- loss probabilities array (one in X will be lost)
-local loss_prob = {}
+local loss_prob = {
+	["default:cobble"] = 3,
+	["default:dirt"] = 4
+}
 
-loss_prob["default:cobble"] = 3
-loss_prob["default:dirt"] = 4
-
-local radius = tonumber(minetest.setting_get("tnt_radius") or 3)
+tnt.radius = tonumber(minetest.setting_get("tnt_radius") or 3)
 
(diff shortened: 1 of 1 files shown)


LeMagnesium:
I think it should be done later. I will do it, but after this is merged (or not). It is a separated feature.

ShadowNinja:
This needs to use more localizations.  For example `local def = cid_data[cid]` does one hash lookup (get `cid_data` by local index, look up `cid` in `cid_data`) wheras `local def = tnt.cid_data[cid]` does *three* (look up `tnt` in `_G`, look up `cid_data` in `tnt`, look up `cid` in `cid_data`).

LeMagnesium:
Among the values, tables and variables I added to my localization PR, and after you pointed `cid_data`, I agreed this table is not useful to stock in the namespace (knowing that it contains content ids, and they are not needed outside of the mod). I still have a doubt about the last table in the namespace, containing the probabilities to be destroyed instead of dropped. I thought that any mod that would need to create its own blast system might want to use these values.
So to reduce calls, I will just include functions in the namespace, not variables or tables.

LeMagnesium:
Rebased after 4bd1094619f8001fa4418465a03aecbeb7f93b5b.

LeMagnesium:
I fixed variables which had become global (thanks @ShadowNinja ) and squashed my commits into one because.. well.. three commits for just 2 variables is a bit too much.

ObaniGemini:
Is anyone for ? Anyone against ?

MT-Modder:
:+1:  This is a very good idea.

sofar:
This is superseded/partially included by #862 (and others, before that).

PR #504, opened 2015-05-09
Change the registered name of the bed

Rui914:
Can register the bed from the other mods.

### DIFF
--- a/mods/beds/api.lua
+++ b/mods/beds/api.lua
@@ -1,5 +1,5 @@
 function beds.register_bed(name, def)
-	minetest.register_node(name .. "_bottom", {
+	minetest.register_node(":" .. name .. "_bottom", {
 		description = def.description,
 		inventory_image = def.inventory_image,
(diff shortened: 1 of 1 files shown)


PilzAdam:
This is not needed; beds can be registered by other mods if they use their proper mod-prefix in the name.

Rui914:
It is, registration is possible, but the recipe is not properly registered.

LeMagnesium:
> but the recipe is not properly registered.

What do you mean by this? You've not modified anything related to the recipes.

Rui914:
https://www.dropbox.com/s/ezh88iifjg03q0q/Carpets.zip?dl=0
In this, error in carpet_tools.

LeMagnesium:
Ok. First, it's not a craft recipe, it's a register_node which breaks, and second, you used the wrong prefix for your itemstring, it should be `carpet_tools` and not `carpet`.

PR #535, opened 2015-06-08
Rename README.txt to README.md

phase:
It's already in Markdown format, so why not have GitHub render it as such?


Calinou:
It is intentionally not named README.md to not confuse Windows users (who need to set a file association to open those files). Sadly, Windows does not acknowledge MIME types.

phase:
It works fine on my windows. All they have to do is say "open with"; or install Notepadd++.

PilzAdam:
Duplicate of #256.

PR #542, opened 2015-06-16
Add the listring functionality

Amaz1:
This makes chests, locked chests, the creative inventory and the craft grid work with https://github.com/minetest/minetest/commit/c977fbd9288d0e80019554411a2f1885f7760568

Bookshelves and vessel shelves also work nominally with it, but because both of those only allow one item per slot, shift+click seems to fail... I'm fairly certain this is an engine problem, so should I remove the listring code for those two items?


est31:
They should work agree, leave them in.

est31:
:+1:

Novatux:
:+1: :)

ShadowNinja:
:+1:

est31:
7361eb5e0586c5c3b4ac4f9d1474ed5d75ff1d6e

PR #543, opened 2015-06-17
Strengthen obsidian glass

tenplus1:
Obsidian glass should be harder to break than it is just now...

### DIFF
--- a/mods/default/nodes.lua
+++ b/mods/default/nodes.lua
@@ -1757,7 +1757,7 @@ minetest.register_node("default:obsidian_glass", {
 	is_ground_content = false,
 	sunlight_propagates = true,
 	sounds = default.node_sound_glass_defaults(),
-	groups = {cracky = 3, oddly_breakable_by_hand = 3},
+	groups = {cracky = 3},
 })
 
 


RHRhino:
makes sense :+1: 
(But why is the new group level=2 instead of oddly_breakable_by_hand? Has that changed or is that a "typo"?)

C1ffisme:
That sounds like a good idea, In fact, we should make it blast-proof too, so that you can use it for bomb shelthers!

tenplus1:
RHRhino: bope, not a typo...  If obsidian glass is strengthened then the ability to dig it by hand shouldn't be there...

rubenwardy:
IRL obsidian is pretty weak. But this isn't real life and obsidian in game is strong so obsidian glass should be strong too.

Jordach:
@rubenwardy - obsidian IRL can be very hard to break. It just depends on
the formation. Assume that MT uses the hardened form.

kilbith:
IRL, obsidian is hard but brittle. Great hardness doesn't mean that it's hard to break.

(I personally carved knives with obsidian)

rubenwardy:
Exactly, kilbith. Like how glass is hard but brittle. I think Jordach was saying that it may not be brittle, though.

HybridDog:
mese is harder than obsidian, isn't it?

paramat:
:-1: For cracky 2 and level 2, this makes it as hard as stone bricks and requiring a tool with a level high enough for obsidian.
However i would +1 just removing 'breakable by hand' so players can't just punch your windows out.

0-afflatus:
Again gameplay wins over reality. I would support the ability to make 'toughened glass'.

Does obsidian glass even exist in rl?

paramat:
Well obsidian is a 'volcanic glass' and exists in clear and coloured forms, but sheets of obsidian-infused glass i'm not sure about. It probably wouldn't be any tougher, i just like the idea of having a glass type not punchable.

paramat:
Needs squashing.

tenplus1:
paramat: krock squashed it for me, could you add it already :)

paramat:
Okay cracky 2 seems ok to me now to make it tougher than normal glass :+1:

ShadowNinja:
Obsidian glass was originally added as a compromise between the "obsidian shatters easily" crowd and the "MC uses hard obsidian so we should too" crowd.
IMO they should both be realistically fragile.

tenplus1:
The general idea was to make it so that players required a tool to break obsidian glass which the latest changes handle fine.

paramat:
Perhaps it should be cracky = 3? Otherwise its harder than stone.
Removing 'breakable by hand' seems enough change by otself.

tenplus1:
@paramat as it stands the change will work for the better and players will craft obsidian glass as an alternative so it's not easily broken by noobs.  Please add.

PilzAdam:
I'm not sure whether it should be `cracky=2` or `cracky=3`, but it should definitely not be breakable by hand.

paramat:
I'll +1 if it's changed to cracky = 3.

tenplus1:
@paramat - setting cracky to 3 still feels too quick to dig whereas 2 feels like it's obsidian glass and not normal glass being dug up with a pick.

PilzAdam:
@tenplus1 you can control the dig time by setting a `level` group, too.

tenplus1:
@PilzAdam - lol, this was originally added but deemed unsuitable at the time so it was removed.

paramat:
Yeah that was me who objected, the level 2 was making it insanely tough.
I know that cracky 3 is the same as normal glass, but removing breakable by hand is what makes the difference, it seems somewhat wrong to have glass tougher than stone. However i don't feel strongly about it if other devs are ok with cracky 2. It seems ShadowNinja would prefer cracky 3 also?

tenplus1:
The changes have been in effect on Xanadu server for a few weeks now and players are happy with the differences in glass types, actively using obsidian glass to make their builds tougher for noobs to get into.  I'd rather the cracky level stay at 2.

paramat:
The difference is only a slightly longer dig time with the same tool, only 'level' would make a significant difference in protecting against other players.

tenplus1:
@paramat - yes, a slightly longer digtime but you MUST use a tool, that's the main difference.

paramat:
Sfan5 had no opinion on this. ShadowNinja sounds like they would like a less tough option, PilzAdam is neutral on cracky but agrees to removing breakable by hand. So, you can get 2 approvals with cracky = 3 .. =)

paramat:
Thanks! :+1:

paramat:
@PilzAdam > I'm not sure whether it should be cracky=2 or cracky=3, but it should definitely not be breakable by hand.

I'll assume this is a +1.

paramat:
Needs rebasing. Also please could you add spaces either side of the 2nd '='?
@PilzAdam is your comment a +1 now this is ```groups = {cracky = 3},```?

PilzAdam:
@paramat I guess; I'm not really sure, but if others are comfortable with it then it's ok.

sofar:
I'm ok with removing the oddly_breakable_by_hand group.

tenplus1:
@paramat REBASED!

paramat:
Approval from nore http://irc.minetest.ru/minetest-dev/2016-04-11#i_4582023

paramat:
6e1a3297636bc0e20d623f0a3214144158fddb66

PR #548, opened 2015-06-30
Improved ores blocks textures

kilbith:
The new ones look more "voxel-ish" and have different patterns.

Before / after : 

old
new

dia_old dia_new


Calinou:
:+1:, they look cleaner.

MirceaKitsune:
:+1:

nerzhul:
:+1: for me

sfan5:
:+1:

asl97:
i didn't notice that the blocks use the same pattern. still, i prefer the old texture with it's roughness at the edge.

ExcaliburZero:
:+1: Those look a good bit better than the current ones.

C1ffisme:
:+1: They look smoother and cleaner, more like polished blocks instead of badly refined ones.

est31:
:-1: I hate them.

paramat:
Yes i like these clean looking textures.

C1ffisme:
The old ones look like someone threw the blocks off a skyscraper. The new ones look like they are just out of the factory.

kilbith:
To all : I'd like to change the diamond block texture. How do you like it ?

dia_old dia_new

paramat:
Thay diamond block is a huge improvement, the old one looks dirty.
Some comments on the other 3...
The gold block has bevel highlighting wrongly placed along the top of the texture, it is placed 3 pixels in from the top but 2 pixels in from the left side.
For all 3... they still look a little dull, perhaps they could be made to look more shiny, metallic and reflective somehow. Bevelled edges with strong highlights and shadows might help.

kilbith:
OK thanks, so I think I'll push it. @Calinou told me he likes it too.

kilbith:
@paramat Highlighting fixed.

C1ffisme:
Now diamond blocks look more like a rare, expensive material and less like lapis luzuli or some cheap blue stone. Good job!

pumarc:
:+1: i like.

Novatux:
I like the new diamond blocks, however I am not sure about the others (I am not against reducing the noise at the edge, but I reckon this is too much)

RealBadAngel:
i like them but the diamond block (neat) is just a bit to bright, it should be closer with saturation to the original one.

kilbith:
So I worked on them considering all your feedbacks (shininess, noise on edges, etc.)
How do you like that ?

screenshot_20150708_235359

screenshot_20150708_235447

Calinou:
Even better now.

est31:
still :-1: .

asl97:
ores :+1: 
diamond :-1: 
even though the diamond block is better now,  the pattern still look too perfect,  i suppose other people would be fine with it though

paramat:
Yes these are okay, a little noise on the edges is an improvement.
The diamond block, which i love (and prefer in it's original brightness, diamond should be very clear and bright) is actually from Gambit's PixelBOX pack (WTFPL), i was thinking it was your own.

kilbith:
@paramat I added [the credit](https://github.com/minetest/minetest_game/pull/548/files#diff-7d4ae2e3f49f3ad3c9345532654d2dcaR204) to Gambit for the diamondblock since I introduced it.

Some people don't like the original brightness so I found a good compromise with the current texture, although I can't satisfy everyone...

nerzhul:
I dislike the diamond too, it looks like a metal gear solid box but blue

paramat:
Oops i missed the credit in the commit, sorry.

gravgun:
@kilbith A little more noise on the gold block would be good I think

C1ffisme:
I don't know about the noise on the edge. It looks less polished. As if the factory worker lazily left it messy and told his boss it was "decorative".

est31:
@C1ffisme what about removing all color difference entirely? then thats "clean" in your eyes? As we don't have fancy graphical effects, we need noise at an element level to give it at least some sense of reality, and to make shapes distinguishable.

I strongly oppose all new textures except the diamond block in this PR. The border should remain strong, we don't need a second light border for the gold block (only on the top WTF). The new noise on the copper block (and in fact on all new textures) looks like ugly jpg noise, and there is a weird "H" letter visible on the left lower edge (less visible on the "worked on" image, but still there).

The diamond block is the only not totally ugly change. The old one was very dark.

kilbith:
@est31 Look closer the gold block, there's a 2nd (subtle) light border.

C1ffisme:
@C1ffisme No, but it does look slightly bent up and pushed around... Although, I suppose there is the fact that nothing made by human is perfect. (It's only human.)

gravgun:
@C1ffisme "nothing made by human is perfect", mainly because perfectness is a human concept.

Novatux:
Looks good

kilbith:
Will open an issue for the ores.

PR #559, opened 2015-07-08
Add some normalmaps with heightmap in alpha channel

RealBadAngel:
screenshot_20150709_000235
stone, desert stone, sands, cobble, all minerals and bricks.
screenshot_20150708_203041
screenshot_20150708_211506
screenshot_20150708_211615
screenshot_20150708_211706
screenshot_20150708_211910

screenshot_20150709_002816


PilzAdam:
Shouldn't this be done in a texture pack?

RealBadAngel:
No, those are for default textures only. Most of them fixes issues that autogen cannot (and never will) solve. Like bricks, ores for example.

kilbith:
@RealBadAngel There's a mistake with the copper pattern, it's not the same than the other ores (except mese). Otherwise :+1:

Calinou:
:+1:

HybridDog:
How would it look like if you used waifu2x instead of simple upscaling making every pixel come out of the texture?
http://waifu2x.udp.jp/
196f5b3c-eaf9-4dcd-cdfb-85819f1359b0

kilbith:
@RealBadAngel Can you add the normalmap for the bookshelf please ?

RealBadAngel:
@HybridDog i cant see how this can be usefull here.
@kilbith makin a good map takes time, will add them one by one (or others can do that too).

C1ffisme:
Could you make the normalmap for the brick a little less bumpy? It looks like the inside of a brick house would be real weird.

Novatux:
Looks good, although the brick could be a bit more smooth IMO

sfan5:
Looks good.

sfan5:
166b1c623add33a0ad4ba469da30035df5293bca

PR #564, opened 2015-07-10
Better bucket textures

ElementW:
Ported over from my commit Ombridride/minetest-minetestforfun-server@6aebcc4a4e5a1f01d4bcd70e0394d681a71a428b.
Fixes the rather ugly appearance the buckets have, at least a bit (I know those textures aren't _perfect_ either, but I think they look **much** nicer than the current ones).
Criticism welcome, will probably take the time to edit the textures and amend the commit.


asl97:
some people would prefer a screenshot or blown-up image of them since it's so small, just to show how they look like.

i prefer the thicker rim and shorter bucket like the current default texture.

here is a blown-up image:
buckets

RHRhino:
They have more details, nice! :+1: 
BTW which license? CC0?

gravgun:
@RHRhino CC0, WTFPL, Public Domain, whatever. I don't even care about credit (I don't consider my 16x16 (and below) artwork to be important enough to put restrictions on them).

kilbith:
OK for me.

C1ffisme:
Looks nice and shiny. :+1: !

C1ffisme:
Actually, it might be better to make a bucket texture and liquid overlays like the ore textures. That would save lots of time for people who make other bucket liquids.

tenplus1:
+1 for overlay idea...

gravgun:
@C1ffisme @tenplus1 Made into overlays.

kilbith:
The overlay is not necessarily a good idea.

First at all, it breaks the texture packs.
Second, the round shape of the liquids have to adjusted to the bucket. On HD textures it's not easy to dissociate a non-straight shape properly against a complete alpha background.

gravgun:
> dissociate a non-straight shape properly against a complete alpha background.

@kilbith you mean when one edits the texture? If then, people should get a real image editor, not a flaky one not supporting layers.
But also yes this would imply that other mods switch to overlays, which would, in cases where the texture is not overlaid properly (possibly by usage of a texture pack), have bad results.
A sustainable overlay idea would be to use a mask along with the base bucket texture, and use it on a "bucket content" texture before overlaying it on the bucket -- however it gets more complex with different texture resolutions.
Should I rollback the changes?

kilbith:
What I mean is : 
- You necessarily have to draw the liquid inside the bucket before dissociate it on another layer.
- On HD textures, pixels are small and gradients have an important disparity of tones. Thus it's not easy to cut the right area for making a proper dissociation (even on Gimp).
- Higher-res textures can also afford totally different shapes for each filled buckets.

> Should I rollback the changes?

Yes. Also C1ffisme's ideas are generally pure noise, he's not even a modder but have an opinion on literally everything.

gravgun:
Rolled back overlay changes.

C1ffisme:
@kilbith Actually, I have a texture pack not on github (Too bright and flashy) several games in several different points of completion, and a couple mods. But most I keep to myself, because they just aren't good enough for public consumption. Maybe an LAN server with friends, but not servers and everything.

paramat:
Yes overlays are too complex for this, best to keep it simple.

Novatux:
Looks good

paramat:
Our current textures are a bit sloppy and bent out of shape.
I'm happy with these but please can you correct the hue of the water and river water? Water is hue 209 and river water is hue 194. If you open the 'default water' and 'default river water' textures then 'flatten' them to remove alpha you can use those pixels as a palette of different brightnesses.

GunshipPenguin:
:+1:  to the new textures from me. I think they're a definite improvement.

sfan5:
IMO overlays are a good idea.
'Texture packs will break' is not a valid argument as that would hinder any kind of changes to how textures are used; texture pack creators will just have to update the packs.
Considering different textures for different buckets, this can be done no matter whether overlays are used or not.
> Also C1ffisme's ideas are generally pure noise, he's not even a modder but have an opinion on literally everything.

Attacking people on the issue tracker is not appreciated. @C1ffisme has the rights to have an opinion and to say it. In the same way, your opinion is not any 'better' than other opinions just because you contribute to Minetest more extensively. Don't be such an ass.

gravgun:
@paramat River water is now 190°-207°, and water 202°-219°.
@sfan5 Let's settle it: are overlays a good enough idea?

kilbith:
@sfan5 An opinion is necessarily better when based on experience. I'm not the only one to complain about the nonsenses of C1ffisme (see IRC and grep that nick).

About "attacking people", you should be impartial with some devs instead of bothering me (as usual)... Don't be such a white knight.

sfan5:
@kilbith Not being the only one complaining doesn't make it right.

@gravgun IMO you should use overlays as they add more flexibility.

PilzAdam:
:+1: for new textures (overlays).

kilbith:
@sfan5 Here's [an example](https://github.com/minetest/minetest_game/pull/486#issuecomment-90748117) of nonsense among hundreds that I'm referring about. This is assertive, leads by pure ignorance and rather noisy upon the tracker. I doubt you're aware of all that.

paramat:
Okay overlays are fine by me, as there is support in the team.
gravgun a matching hue would be nice but i'm not too bothered, those are close enough.

C1ffisme:
Kilbith, it's just how some people talk. What do you expect of people? Cursing, Shutting up and Worshipping you? You aren't the supreme ruler of the universe here. I'm not going to turn into anyone but me. So stop this. We don't have to be enemies. Especially on a website for collaboration.

About overlays, I'm just suggesting. When is suggesting a bad thing? When was it ever? And if people change there pull requests to match things I suggest, they just like the idea. If I had suggested something absolutely terrible, they wouldn't have changed the pull request to it.

HybridDog:
using an overlay frame image (like the one for fence inventory textures) makes it impossible to make the liquid in the bucket have different transparency at different pixels and every liquid would flow out of the bucket on the same side
but using a texture which is put onto the empty bucket texture works and if the bucket texture becomes changed, the texture of the mod doesn't need to be changed (except if the bucket gets a different aperture), e.g. https://github.com/HybridDog/sumpf/blob/master/sumpf/textures/sumpf_bucket_dirtywater.png https://github.com/HybridDog/sumpf/blob/master/sumpf/init.lua#L192

gravgun:
@HybridDog That's exactly was I had done before reverting my overlay changes...

paramat:
+1. For new textures, no overlays, let's keep it simple.

gravgun:
Deleted repo, cancelling PR.

You guys are damn too slow, fearing every change will make some players unhappy, which is why minetest_game as well as MT core are still bad as they are.

ShadowNinja:
:+1: on the textures.  I'm undecided on overlays (without them if you change the bucket texture every mod has to update theirs, but with them a bucket texture change could make the liquids not line up any more).

This PR's actually been O.K. to merge for a while (it got it's approvals) it just needs someone to actually commit it.  I can't reopen this since the repo's been removed though.

est31:
pushed: 0004c5e637fae407f5e95e589686a8a5f9a93420

paramat:
Thanks for merging, basic new textures had approval (Ekdohibs and myself) but was delayed over uncertainty about use of overlays.

HybridDog:
@gravgun l think freeminer doesn't wait so long to merge pull requests.

asl97:
paper thin bucket...... (sigh)
it lost it sense of depth......

RHRhino:
@HybridDog thats probably because the freeminer dev team is much smaller than minetest ...

gravgun:
@HybridDog I never mentioned freeminer nor ever contributed to them, so that's not a thing I'd compare with. IMHO Freeminer is even worse than MT since it doesn't evolve the modding way.

PR #565, opened 2015-07-12
Add mushrooms

kilbith:
This is the addition of mushrooms in a zone that was a bit miserable in terms of decoration (we only had flowers). The red one is lethal on eating. The brown one gives health.

@paramat is willing to do the mapgen stuff for those (spawning near the trees), after a discussion we had on IRC.

screenshot_20150712_124902


PilzAdam:
Insta-kill when using an item is not good. At least make it gradually, so the player has time to heal against it. I'd prefer to make it cost about 3 hearts.

kilbith:
@PilzAdam Done. Anything else ?

PilzAdam:
Making mushrooms able to grow in dark areas on dirt may be an idea worth considering. The regrowing group from the flowers mod can't be used in this case.

Speaking of the flowers mod: This should be in the flowers mod, shouldn't it?

kilbith:
`flora` group removed then. Mushrooms in `flowers` mod ? Obviously inconsistent (unless if we rename the mod to `flora`).

PilzAdam:
We can't rename mods due to dependencies. It makes enough sense to me to be in the flowers mod. default is already too bloated.

kilbith:
OK, this also prevents the PRs conflicts and rebasing.

paramat:
I like the idea of having mushrooms. I might prefer to have just one for simplicity, the brown one, which is the useful one (however if we have both we could possibly have weird player effects from the poison, like altered skybox and gravity, as well as damage).
Also you're correct this should be plantlike drawtype for consistency, not a meshnode.

kilbith:
@paramat This is a distinct group of flora placed in specific places. 2 mushrooms represent a minimum of diversity in that group (for comparison : we have 6 flowers and 5 tall grasses). The harmful poison is a danger inherent to a challenging game. Also the red one is so cute that I couldn't get rid of it :)

VanessaE:
Putting my argument here for the record.

I disagree with making the mushrooms use plantlike.  That drawtype works fine for grasses and other plants because they're all basically flat anyway.  I've never seen an instance where a mushroom can be drawn flat as done here - the textures are nice but a flat mushroom simply doesn't look good.  Mushrooms should be a low-poly mesh node; the one in plantlife modpack's mushroom mod works perfectly fine in my opinion.

Aside from stairs (which look exactly the same whether they're nodeboxes or meshes), what's the point of having meshes if minetest_game doesn't exploit the feature?

kilbith:
I strongly disagree for using a mesh and your opinion is biased, @VanessaE. 
- A mesh is not texture-pack-friendly nor shape-flexible (only one shape is determinated and not easy to modify). This is against the primary philosophy of Minetest.
- This is an abstract and simplistic world made of cubes, a fully realistic model is not suitable in a voxel-style game. This is the reason why MC kept that style. We are officially a MC inspired game.
- The plants are not basically flat, they don't grow strictly on the same rows in cross and the plants themselves aren't flat anyways (especially papyrus).
- Using a mesh here would be a slippery slope for using more mesh models (I don't want minetest_game heading to be a copy of your game, Dreambuilder).
- Mesh on stairs has been used only for increasing the performances, not for an aesthical reason at all.

paramat:
I'm not too bothered about having 1 instead of 2, lets see what others think.

paramat:
I've changed my mind. While plantlike is good for most plants, i don't like it for mushrooms. I would prefer an ultra-low resolution (minimum of 1/8th node microvoxels) nodebox mushroom, lets say 2 or 3 cuboids: stem, cap and perhaps smaller cap top. Defined as a nodebox it may well be easily textured just as slabs etc. are. OCD has nodebox mushrooms:

screenshot_1

C1ffisme:
I don't mind plantlike mushrooms. After all, the hills should be slanted and the sun should be round, but they aren't! So why would mushrooms be any different?

If we absolutely need good shaped mushrooms, we can use nodeboxes. But limit them to a stick and a plate/cube.

Calinou:
Plantlike mushrooms are fine. I agree with adding mushrooms in the game, we need more stuff to play with and to even look at.

kilbith:
@paramat Multiple nodeboxes add and hide unnecessary vertices and it's bad for performances. And I'm repeating myself for the _n_ time : any hardcoded model is not texture pack friendly nor KISS.

I hope your change of mind is not the consequence of a certain influence (you were pro-flatness usually).

kilbith:
Well in fact it's useless to debate. You just don't consider my comments.

paramat:
I do. Best to keep this open, with more comments you may find there is a majority support for plantlike mushrooms, although i would prefer 3D, plantlike is acceptable and certainly consistent. No-one influenced my change of mind. For most plants i still prefer flat textures with plantlike drawtype.

kilbith:
Sorry don't take that personally, just that I'm convinced that plantlike is the way to follow. I like realism but not excessively.

PilzAdam:
I prefer plantlike drawtype, since it grants texture packs much more freedom. It also looks decent enough.

Rui914:
This mushrooms looks good!
This adds color to minetest_game.
Also, it's cute!

paramat:
+1 looks good.

Novatux:
Looks good !

GunshipPenguin:
:+1:  I actually really like the look of plantlike mushrooms.

Novatux:
@paramat : could you please add mapgen support for those mushrooms?

Novatux:
https://github.com/minetest/minetest_game/commit/0cf4fd4c1800030a3555305c73fc2269c5b79a3d

Rui914:
???
```
not_in_creative_inventory=1
```

PilzAdam:
@Rui914 good catch, that should be changed.

Rui914:
#577

kilbith:
@PilzAdam It's like the tall grasses and dry grasses, I assumed that the wild spawned elements shouldn't be in the inventory.

PilzAdam:
grass_2 to 5 are not in the creative inventory because you always have grass_1 (the other ones drop grass_1, too).

kilbith:
`dry_grass_1` is out of the inventory : https://github.com/minetest/minetest_game/blob/master/mods/default/nodes.lua#L949-967

But I agree for the mushrooms in the invo.

paramat:
Novatux yes will add to mapgen.

PR #568, opened 2015-07-14
stairs model v3

VanessaE:
The v2 model still has some UV mapping glitches (namely, misalignment on the sides of the stairs).  Here's a freshly-made one derived from blender's default cube.

Before (look near the crosshairs; screenshot by RBA):

After:


kilbith:
Can we have a proper before / after comparison please ?

VanessaE:
Model is not needed - the v2 model is fine.  The misalignment issue is an artifact of the handmade cobble normalmap not being able to tile correctly from the cobble over to autogenerated one being used on the stair.  Does not happen with normalmaps disabled.

PR #571, opened 2015-07-15
Node definition for stairs and slabs

Rui914:
Added node definition for stairs and slabs.

### DIFF
--- a/mods/stairs/init.lua
+++ b/mods/stairs/init.lua
@@ -3,18 +3,19 @@
 
 stairs = {}
 
+local recipeitem_things = {"tiles","use_texture_alpha","alpha","light_source","sounds","groups"}
+
 -- Node will be called stairs:stair_<subname>
-function stairs.register_stair(subname, recipeitem, groups, images, description, sounds)
-	minetest.register_node(":stairs:stair_" .. subname, {
-		description = description,
+function stairs.register_stair(subname, recipeitem, groups_or_def, images, description, sounds)
+	local recipe_def = table.copy(minetest.registered_nodes[recipeitem]) or {}
+	local node_def = {
+		description = recipe_def.description.." Stair",
+		is_ground_content = false,
 		drawtype = "mesh",
 		mesh = "stairs_stair.obj",
-		tiles = images,
 		paramtype = "light",
 		paramtype2 = "facedir",
 		is_ground_content = false,
-		groups = groups,
-		sounds = sounds,
 		selection_box = {
 			type = "fixed",
 			fixed = {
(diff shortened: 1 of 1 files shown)


mods/stairs/init.lua
```diff
@@ -4,8 +4,8 @@
 stairs = {}
 
 -- Node will be called stairs:stair_<subname>
-function stairs.register_stair(subname, recipeitem, groups, images, description, sounds)
-	minetest.register_node(":stairs:stair_" .. subname, {
+function stairs.register_stair(subname, recipeitem, groups, images, description, sounds, def)
```
HybridDog: instead of adding def as an additional parameter to the function, you could use the third one (groups) as def if images isn't given. Then you could define the groups, tiles (images), description, sounds and more in the third parameter (def (or groups))

Rui914:
Updated.
How about this?

HybridDog:
l like it, maybe for a few things of the node_def the stuff could be copied from the node definition of the recipe item, e.g. groups, walkable and use_texture_alpha

Rui914:
Updated.
But, still WIP.
I need review.

mods/stairs/init.lua
```diff
@@ -4,17 +4,20 @@
 stairs = {}
```
HybridDog: you could add here a table of things which should be copied from the recipeitem, e.g. {"alpha", "groups", …(and more from http://dev.minetest.net/register_node#Node_definition)} and set the things in the nodedef to the ones of the recipeitem instead of writing each one into the nodedef (https://github.com/minetest/minetest_game/pull/571/files#diff-d85e6749f6707aa8b7c089afb33ccdecR17)

Rui914:
```
local recipe = minetest.registered_nodes[recipeitem] or {}
local table = {
	alpha = recipe.alpha,
	group = recipe.group,
	...
}

...

for n,i in pairs(table) do
	node_def[n] = i
end
```
This like?

HybridDog: no, l mean a list of indices

```
stairs = {}
local recipeitem_things = {"alpha", "groups", "walkable", …}
…
local recipe_def = minetest.registered_nodes[recipeitem]
for _,i in pairs(recipeitem_things) do
    nodedef[i] = recipeitem_things[i]
end
…
```

Rui914: I understand.

Rui914:
Updated.
By the way,
```
if groups_or_def and groups_or_def.description then
                                                                 ↑
```
Should do judged by "description"?

HybridDog:
>Should do judge by "description"?

maybe it's better to test if images exists, some node might have a description group

Rui914:
```
	if not images then
		for n,i in pairs(groups_or_def) do
			node_def[n] = i
		end
	else
		node_def.groups = groups_or_def or node_def.groups
	end

	node_def.tiles = images or node_def.tiles
	node_def.sounds = sounds or node_def.sounds
	node_def.description = description or node_def.description
```
OK?

HybridDog:
yes

Rui914:
Updated.

PR #577, opened 2015-07-17
Remove not_in_creative_inventory in the some nodes

ghost:
Remove not_in_creative_inventory=1.

PR #578, opened 2015-07-17
Remove not_in_creative_inventory on the default:dry_grass_1

ghost:
Remove not_in_creative_inventory=1.
And grass node like.


kilbith:
You should move that change into https://github.com/minetest/minetest_game/pull/577 and change its title.

PR #610, opened 2015-08-05
Prevent door duping. Fixed invalid condition.

### LINKED ISSUE #609: Door duplication bug
1. Place a door.
2. rip the lower and the upper half appart (for example with pistons)
3. open/close the door-parts
4. there are now two brand-new doors

https://www.youtube.com/watch?v=f4hc8jGQgoU

So solve this, the upper half should vanish (instead of repairing itself).

btw this must not be solved by changing the mesecon/piston code, because the door mod should be able to handle partially obstructed doors (which it does when they are not used before digging)

wischi-chr:
This fixes #609 and prevents door duping

### DIFF
--- a/mods/doors/init.lua
+++ b/mods/doors/init.lua
@@ -129,7 +129,7 @@ function doors.register_door(name, def)
 
 	local function on_rightclick(pos, dir, check_name, replace, replace_dir, params)
 		pos.y = pos.y+dir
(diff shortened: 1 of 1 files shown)


Ekdohibs:
Looks good

ShadowNinja:
35de5241f457a88928c74eec0a74bc1638c615dd

PR #618, opened 2015-08-09
Better jungle wood and jungle tree top textures

VanessaE:
Jungle tree wood candidate from kilbith
Jungle tree top derived from the existing image combined with a candidate image from @LazyJ

See also issue #611.


kilbith:
Orange-ish core & mauve-ish ring on trunk, really ? Is that radioactive ? Have you smoke ?

This repo is getting to be a cesspit where people dump their poop. I can't no longer be associated with that stream of braindamaged craps. Do whatever you want, I don't care.

PR #624, opened 2015-08-11
Vessels: Just a little clean up code

Rui914:
Improvement of readability


PilzAdam:
Don't remove the comma at the end of the last element in tables. It makes it much easier (and produces nicer diffs) to add new elements to it.

PR #645, opened 2015-08-21
Check hp in explosion of TNT

Rui914:
Don't kill again the player that already dead.


PilzAdam:
Shouldn't this be properly handled by the engine?

Rui914:
@PilzAdam 
Yes.
But, I don't understand C and C++.
Should I submit new issue?

Rui914:
https://github.com/minetest/minetest/pull/3090

PR #650, opened 2015-08-25
change mineral textures

HybridDog:
change mineral textures


paramat:
Don't worry about providing normalmaps, they are complex to make and apparently need to be high resolution (gimp has a filter to create normalmaps from textures).
Please can you post before / after screenshots?

paramat:
These seem to be the ones from #313, the coal is okay, diamond and gold are excellent. Copper needs editing as it has messy use of alpha around the edges, perhaps you could remove the translucent pixels and make it shinier and more metallic, in the way diamond and gold are?

kilbith:
This is total BS...

The ores aren't supposed to be overloaded in stone like that. Rather discreetly inlaid to blend well within the stonish environment and not being too "showy" to the eyes of the heedful miners.

Look at PixelBOX for examples.

HybridDog:
>I feel coal and copper need editing as they have messy use of alpha around the edges, perhaps you could remove the translucent pixels?

l actually wanted to make a smooth transition between mineral and stone texture.
>Please can copper be shinier and more metallic, in the way diamond and gold are?

l would rather like shaders to do reflections.
>The ores aren't supposed to be overloaded in stone like that. Rather discreetly inlaid to blend well within the stonish environment and not being too "showy" to the eyes of the heedful miners.

sorry, maybe l made it unknowingly to get more success in mining

paramat:
Okay no problem. I don't like the use of alpha to blend, so my opinion is to use the diamond and gold textures only. If HybridDog and other devs are okay with this i can make a new PR.

HybridDog:
the gold maybe has too much saturation

paramat:
Gold and Diamond look good to me, if you open a new PR with those 2 i will +1.

HybridDog:
The diamond texture is from the diamonds mod. l think minetest took its diamond textures from an old version of the mod.

kilbith:
From [Cisoun's Texture Pack](https://forum.minetest.net/viewtopic.php?id=16) (WTFPL) - gold texture : 

capture d ecran_2015-09-21_21-35-17 default_gold

PR #669, opened 2015-09-09
Tidy sethome code, add global functions, round coords to 1 decimal

tenplus1:
Tidied code, added global functions so that mods can get/set/go with home positions, also rounded coords to 1 decimal place.

Example: https://github.com/tenplus1/inventory_plus/blob/master/home_gui.lua

### DIFF
--- a/mods/sethome/init.lua
+++ b/mods/sethome/init.lua
@@ -1,65 +1,80 @@
+
+sethome = {}
+
 local homes_file = minetest.get_worldpath() .. "/homes"
 local homepos = {}
 
 local function loadhomes()
-    local input = io.open(homes_file, "r")
-    if input then
-		repeat
-            local x = input:read("*n")
-            if x == nil then
-            	break
-            end
-            local y = input:read("*n")
-            local z = input:read("*n")
-            local name = input:read("*l")
-            homepos[name:sub(2)] = {x = x, y = y, z = z}
-        until input:read(0) == nil
-        io.close(input)
-    else
-        homepos = {}
-    end
+	local input, err = io.open(homes_file, "r")
+	if not input then
+		return minetest.log("info", "Could not load player homes file: " .. err)
+	end
+
+	-- Iterate over all stored positions in the format "x y z player" for each line
+	for pos, name in input:read("*a"):gmatch("(%S+ %S+ %S+)%s([%w_-]+)[\r\n]") do
+		homepos[name] = minetest.string_to_pos(pos)
+	end
+	input:close()
 end
 
 loadhomes()
 
+sethome.set = function(name, pos)
+	local player = minetest.get_player_by_name(name)
+	if not player or not pos then
+		return false
+	end
+
+	local data = {}
+	local output = io.open(homes_file, "w")
+	if output then
+		homepos[name] = pos
+		for i, v in pairs(homepos) do
+			table.insert(data, string.format("%.1f %.1f %.1f %s\n", v.x, v.y, v.z, i))
+		end
+		output:write(table.concat(data))
+		io.close(output)
+		return true
+	end
+	minetest.log("action", "[sethome] unable to write to homes file")
+	return false
+end
+
+sethome.get = function(name)
+	return homepos[name]
+end
+
+sethome.go = function(name)
+	local player = minetest.get_player_by_name(name)
+	if player and homepos[name] then
+		player:setpos(homepos[name])
+		return true
+	end
+	return false
+end
+
 minetest.register_privilege("home", "Can use /sethome and /home")
 
-local changed = false
-
 minetest.register_chatcommand("home", {
-    description = "Teleport you to your home point",
-    privs = {home=true},
-    func = function(name)
-        local player = minetest.get_player_by_name(name)
-        if player == nil then
-            -- just a check to prevent the server crashing
-            return false
-        end
-        if homepos[player:get_player_name()] then
-            player:setpos(homepos[player:get_player_name()])
-            minetest.chat_send_player(name, "Teleported to home!")
-        else
-            minetest.chat_send_player(name, "Set a home using /sethome")
-        end
-    end,
+	description = "Teleport you to your home point",
+	privs = {home = true},
+	func = function(name)
+		if sethome.go(name) then
+			return true, "Teleported to home!"
+		end
+		return false, "Set a home using /sethome"
+	end,
 })
 
 minetest.register_chatcommand("sethome", {
-    description = "Set your home point",
-    privs = {home=true},
-    func = function(name)
-        local player = minetest.get_player_by_name(name)
-        local pos = player:getpos()
-        homepos[player:get_player_name()] = pos
-        minetest.chat_send_player(name, "Home set!")
-        changed = true
-        if changed then
-        	local output = io.open(homes_file, "w")
-            for i, v in pairs(homepos) do
-                output:write(v.x.." "..v.y.." "..v.z.." "..i.."\n")
-            end
-            io.close(output)
-            changed = false
-        end
-    end,
+	description = "Set your home point",
+	privs = {home = true},
+	func = function(name)
+		name = name or "" -- fallback to empty player name if nil
+		local player = minetest.get_player_by_name(name)
+		if player and sethome.set(name, player:getpos()) then
+			return true, "Home set!"
+		end
+		return false, "Player not found!"
+	end,
 })
--- a/mods/sethome/init.lua
+++ b/mods/sethome/init.lua
@@ -26,7 +26,7 @@ sethome.set = function(name, pos)
 	end
 
 	local data = {}
-	local output = io.open(homes_file, "w")
+	local output, err = io.open(homes_file, "w")
 	if output then
 		homepos[name] = pos
 		for i, v in pairs(homepos) do
@@ -36,7 +36,7 @@ sethome.set = function(name, pos)
 		io.close(output)
 		return true
 	end
-	minetest.log("action", "[sethome] unable to write to homes file")
+	minetest.log("action", "Unable to write to player homes file: " .. err)
 	return false
 end
 
@@ -70,7 +70,7 @@ minetest.register_chatcommand("sethome", {
 	description = "Set your home point",
 	privs = {home = true},
 	func = function(name)
-		name = name or "" -- fallback to empty player name if nil
+		name = name or "" -- fallback to blank name if nil
 		local player = minetest.get_player_by_name(name)
 		if player and sethome.set(name, player:getpos()) then
 			return true, "Home set!"


tenplus1:
er... looked like Rui914 was thinking the same thing although saving the sethome positions when one is changed is a lot better than saving on server shutdown (incase of crash)...

HybridDog:
>although saving the sethome positions when one is changed is a lot better than saving on server shutdown (incase of crash)

l think it should be configurable, if the setting isn't set (nil), it takes another setting which tells if the server runs on a pc with accumulator.

tenplus1:
How often to players actually use /sethome ? rather than every n seconds or on shutown which is flakey at best this is the simplest option...

HybridDog:
>How often to players actually use /sethome ?

Guests come and leave, they maybe don't even know the sethome command,
admins, moderators and people who play long on servers hardly use it,
newbies which play there for a short duration is the people that uses it most.
>rather than saving every n seconds or on shutown which is flakey at best this is the simplest option

l don't like saving every n seconds, but saving one time when the server shuts down isn't flakey l think if it saves it even if the server crashed or was stopped via ctrl+c.

rubenwardy:
> isn't flakey l think if it saves it even if the server crashed or was stopped via ctrl+c.

That's not possible, eg: you can't run code or recover from segfaults, as the program state is undefined.

HybridDog:
ok, but it works if it's a simple crash caused by a lua mod

rubenwardy:
Not really - you can't guarantee that the on_shutdown() command won't cause a Lua error as well.

rubenwardy:
If Minetest supported 10,000s of players, I'd use a buffered system that would write to the file every 5 seconds if there has been a change. However, MT only supports 20, so such a system is overkill. The current method is fine. If we supported 10,000s of players, we'd probably use a database to store this data, anyway.

HybridDog:
ok, how about compressing the file using minetest.compress?
https://github.com/minetest/minetest/blob/master/doc/lua_api.txt#L2294-L2300

rubenwardy:
It won't make that much difference - you're talking about a few kilobytes. Unless @VanessaE has a different experience.

everamzah:
What do you mean by "MT only supports 20" @rubenwardy?

rubenwardy:
I'm missing a really there - I'm talking about the number of online players. It's an approximate number, and depends on the hardware.

PilzAdam:
Saving every 5 seconds if a dirty flag is set, as @rubenwardy suggested, would be the best solution. The number of players is not really the problem; one malicious client could be enough. It could bombard the server with any commands it wants, so nothing should cause direct file IO on the server.

est31:
I had trouble with position rounding on the xanadu and clip public servers, I always had to place a dirt node, step on it and then type sethome in order to not spawn in the floor when I typed home. But I don't know whether I have trouble with current code, haven't been on these servers for a long time (one doesn't exist even).

HybridDog:
>Saving every 5 seconds if a dirty flag is set, as @rubenwardy suggested, would be the best solution. The number of players is not really the problem; one malicious client could be enough. It could bombard the server with any commands it wants, so nothing should cause direct file IO on the server.

You could also save it if players didn't do it for 5 seconds:

local num=0
local function save_homes()
	num = num+1
	minetest.after(5, function(oldnum)
		if num ~= oldnum then
			return
		end
		-- save them
		[…]
	end. num)
end

mods/sethome/init.lua
```diff
@@ -2,64 +2,59 @@ local homes_file = minetest.get_worldpath() .. "/homes"
 local homepos = {}
 
 local function loadhomes()
-    local input = io.open(homes_file, "r")
-    if input then
+	local input = io.open(homes_file, "r")
```
t4im: Second return value is the error message, which should either be propagated or at least printed to the log, if input is nil.

mods/sethome/init.lua
```diff
@@ -2,64 +2,59 @@ local homes_file = minetest.get_worldpath() .. "/homes"
...
+	privs = {home = true},
+	func = function(name)
+		local player = minetest.get_player_by_name(name)
+		if not player then
+			-- just a check to prevent the server crashing
+			return false
+		end
+		if homepos[name] then
+			player:setpos(homepos[name])
+			minetest.chat_send_player(name, "Teleported to home!")
```
t4im: Chat commands should return "true" when succeeding and allow a second return value as message to be printed, which makes the extra `chat_send_player`s redundant.

mods/sethome/init.lua
```diff
@@ -2,64 +2,59 @@ local homes_file = minetest.get_worldpath() .. "/homes"
...
+		else
+			minetest.chat_send_player(name, "Set a home using /sethome")
```
t4im: Same here, just:
```lua
return false, "Set a home using /sethome"
```

mods/sethome/init.lua
```diff
@@ -2,64 +2,59 @@ local homes_file = minetest.get_worldpath() .. "/homes"
...
+	privs = {home = true},
+	func = function(name)
+		local player = minetest.get_player_by_name(name)
+		if not player then
+			-- another check to prevent the server crashing
+			return false
+		end
+		homepos[name] = vector.round(player:getpos())
+		minetest.chat_send_player(name, "Home set!")
+		local output = io.open(homes_file, "w")
```
t4im: Error message should be handled.
Also as the others pointed out: never do IO unless you have to.
At the very least an early return should be added above, which you have to do even if you write lazy with dirty flag.

mods/sethome/init.lua
```diff
@@ -2,64 +2,59 @@ local homes_file = minetest.get_worldpath() .. "/homes"
...
+		for i, v in pairs(homepos) do
+			output:write(v.x.." "..v.y.." "..v.z.." "..i.."\n")
```
t4im: You have 7 concatenations per write here and then loop that through all saved positions, passing it to `:write()` each time.
If it weren't in a loop a `string.format` would be most appropriate, but since this can add up to hundreds of entries for a server (which would mean thousands of concatenations you're doing here) i'd rather go for a `table.concat` here ending in a single `:write()`-call.

mods/sethome/init.lua
```diff
@@ -2,64 +2,59 @@ local homes_file = minetest.get_worldpath() .. "/homes"
...
-            local y = input:read("*n")
-            local z = input:read("*n")
-            local name = input:read("*l")
-            homepos[name:sub(2)] = {x = x, y = y, z = z}
-        until input:read(0) == nil
-        io.close(input)
-    else
-        homepos = {}
-    end
+			local x = input:read("*n")
```
t4im: It's faster in lua to read a complete file in and then operate on it via string operations, than having to read it number for number in a loop.
A single match operation per line might be enough to replace all this with a one-liner.

mods/sethome/init.lua
```diff
@@ -2,64 +2,59 @@ local homes_file = minetest.get_worldpath() .. "/homes"
...
-    end,
+	description = "Teleport you to your home point",
+	privs = {home = true},
+	func = function(name)
+		local player = minetest.get_player_by_name(name)
+		if not player then
+			-- just a check to prevent the server crashing
+			return false
+		end
+		if homepos[name] then
```
t4im: Hah. Good, you replaced that.
I'm always amazed how often you can find things like 
`minetest.get_player_by_name(name):get_player_name()`
in minetest mods (even in minetest_game)

mods/sethome/init.lua
tenplus1: lol 4wim, time fo put your code where your mouth is... :)

t4im: > lol t4im, time fo put your code where your mouth is... :)
 
```lua
for pos, name in input:read("*a"):gmatch("([%d%s%.]+)%s(%a+)[\r\n]") do
	homepos[name] = minetest.string_to_pos(pos)
end
```
`string.gmatch` is an iterated `string.match`

pinkysnowman: t4im, i tried that bit of code with no luck btw, but if its a tas bit off well idk im not good with strings yet :)

paramat:
> I had trouble with position rounding on the xanadu and clip public servers, I always had to place a dirt node, step on it and then type sethome in order to not spawn in the floor when I typed home.

Position can't be rounded because standing on a node has a y value containing a .5.
@tenplus1 this needs improving.

paramat:
Please squash commits.

PilzAdam:
@paramat there is no need for non-coredevs to squash their commits. The core dev who merges the PR can do it easily.

tenplus1:
The 'ok dis is final' is the only one that needs applied, the rest are unecessary...

paramat:
Although devs can do it, contributors should squash their commits, we are understaffed so it's better the contributor does it.

tenplus1:
I'm struggling to learn github as it is and have to ask help from users who do know how it works, and as far as I'm aware github website has a button already in place for admin staff to "Confirm and merge" pulls which would be great if it would appear for my own pulls to make things easier than the command line.

paramat:
No problem, i understood core mapgen before basic git use :)
I actually don't know how to edit commits when merging =/

tenplus1:
Admin privs get you this: https://help.github.com/articles/about-pull-request-merge-squashing/

tenplus1:
Squashed by Krock... please add :)

tenplus1:
@paramat - how many devs are reviewing minetest-game pulls ? it seems only textures are being changed.

paramat:
Ok will try to review this. The other MTGame devs are very elusive recently, you have to catch them on dev channel and ask for reviews.

t4im:
@tenplus1 there seem to be still unaddressed issues. An ignored error and wasteful I/O code.
I even wrote you a snippet to fix the latter.

paramat:
^ Please address any issues t4im has raised.

tenplus1:
@t4im pinky tried your code snippet which didn't work, what other issues are you referring to ?

t4im:
Why wouldn't it work? Maybe you need to change the pattern a bit to match the file format (newlines maybe?), but the general idea is still valid. And if that's too complicated, you can still replace it with a multi-liner.

Both issues are named within this issue:
https://github.com/minetest/minetest_game/pull/669#discussion-diff-39928170
https://github.com/minetest/minetest_game/pull/669#discussion-diff-39928417
https://github.com/minetest/minetest_game/pull/669#issuecomment-227855605

tenplus1:
OK, the error in question HAS been addressed, the homepos = {} is already defined at the start of the mod itself, and as for the wasteful io, I'd rather keep this code readable which is still faster than previous sethome mod, saves memory with reduced decimal placed and adds well needed global functions to the game...  @paramat please add this pull as-is...

t4im:
> OK, the error in question HAS been addressed

No, it still is not. The error message is still being thrown away on line 8.

> I'd rather keep this code readable which [...] saves memory

You are keeping it error prone, verbose, expensive and in my opinion also less readable.
`minetest.string_to_pos(pos)` is self explanatory for each entry.

But since you are clearly not interested in fixing these issues, just remove your off-topic changes of the `loadhomes` function and i will withdraw these 2 complaints and fix them myself in a later PR.

Already tested the snippet and it works perfectly, once the pattern is slightly changed to match the file format.

mods/sethome/init.lua
```diff
@@ -1,65 +1,82 @@
...
+	io.close(output)
+	return true
+end
 
-local changed = false
+sethome.get = function(name)
+	if homepos[name] then
+		return homepos[name]
+	end
+	return nil
```
t4im: This is redundant. Just return homepos[name], either it's already `nil` or not.

mods/sethome/init.lua
```diff
@@ -1,65 +1,82 @@
...
 
-minetest.register_privilege("home", "Can use /sethome and /home")
+sethome.set = function(name, pos)
+	local player = minetest.get_player_by_name(name)
+	if not player or not pos then
+		return false
+	end
+
+	local data = {}
+	local output = io.open(homes_file, "w")
```
t4im: This needs proper IO handling. When there were problems with opening the file (permission issues for example) it will crash at `:write` later, but hide, similar to the case in line 8, the error message explaining to the user, what went wrong.

mods/sethome/init.lua
```diff
@@ -1,65 +1,82 @@
...
-                output:write(v.x.." "..v.y.." "..v.z.." "..i.."\n")
-            end
-            io.close(output)
-            changed = false
-        end
-    end,
+	description = "Set your home point",
+	privs = {home = true},
+	func = function(name)
+		local pos = minetest.get_player_by_name(name):getpos()
```
t4im: Would crash if the player isn't found (the other cases handle it well though).

t4im:
The three new issues need addressing though, 2 of them can crash.

tenplus1:
t4m, you are being pedantic with these errors, the /sehome error if  player isn't found wouldnt happen because you would need a player to enter the command, also the permissions and for homes file wasn't in the mod to begin with.

t4im:
> you are being pedantic with these errors, the /sehome error if player isn't found wouldnt happen 

1. Unless that's [exactly](https://github.com/ShadowNinja/minetest-irc_commands/blob/99afcc794b5029e742b63befe1a0e6d346410e5b/init.lua#L114) what a mod does. A command by an off-game user would get a name, but not find an online player to that name. The command should send an error, because sethome for an irc-user is useless, but it should not crash the game for everyone.
2. It is reasonable to assume, that under certain circumstances a disconnect of a user might be processed before his reliable chatcommand. In thise case you wouldn't want the game to crash either.

> the permissions and for homes file wasn't in the mod to begin with.

You rewrote that code. There is no good excuse to intentionally reimplement bugs from previous versions, that can crash the game.

mods/sethome/init.lua
```diff
@@ -1,65 +1,89 @@
...
-                output:write(v.x.." "..v.y.." "..v.z.." "..i.."\n")
-            end
-            io.close(output)
-            changed = false
-        end
-    end,
+	description = "Set your home point",
+	privs = {home = true},
+	func = function(name)
+		name = name or "" -- fallback to empty player name if nil
```
t4im: What do you needs this for? Does `minetest.get_player_by_name(name)` return anything for that?

tenplus1:
t4im - which part is redundant ?

t4im:
@tenplus1 added explanation to the line comment https://github.com/minetest/minetest_game/pull/669/files#diff-76b4c3462c3911215a58c8916ea17235R77

mods/sethome/init.lua
t4im: Just needed a small change in the match string. The one up there was written down on the fly without testing, so it did e.g. not consider negative signs in front of coordinates or special chars in nicknames. But that's easy to fix:

`"(%S+ %S+ %S+)%s([%w_-]+)[\r\n]"`

t4im:
Please also don't forget to remove the (mostly whitespace) changes around `loadhomes()`, unless you want to fix the mentioned issues here yourself. I have a PR ready to push with a repaired loadhomes() function, where i "[put my code where my mouth](https://github.com/minetest/minetest_game/pull/669/files#r40065155)" is ;-)

t4im:
see #1175

tenplus1:
@t4im - if I don't add the ' name = name or "" ' line then the next line crashes game due to a 'bad argument #1 to 'get_player_by_name' (string expected, got nil)' error...  It's required...

also... couldn't you wait until this pull gets added before issuing game#1175 pull ? Now whichever one gets included the other breaks.

tenplus1:
@t4im - if 'name' turns out to be nil due to glitch then the mod crashes, name = name or "" stops this from happening...  I *could* add a further IF that checks and returns another error, but the current way is simpler, smaller, and STILL returns an error if it ever happens.

As for your #1175 pull, if either of our pulls get accepted then the other breaks, couldn't you have waited or even asked me to add your code to this single pull ?

t4im:
> if 'name' turns out to be nil due to glitch then the mod crashes

`minetest.get_player_by_name(nil)` doesn't crash, so that is not a problem, even if `name` ever became `nil`
It only crashes, if you don't check the player object, which you already do. :-)

> if either of our pulls get accepted then the other breaks, couldn't you have waited or even asked me to add your working code to this single pull

I did ask you first to resolve these things, but you ignored them several times. So now i solved it for you, so that this PR doesn't get delayed further (you are welcome). The only "conflict" is the whitespace changes for the `loadhomes()` function, as i already explained twice, there is no merrit in keeping whitespaces changes in a PR that are replaced anyway.

tenplus1:
I have to disagree t4im,  minetest.get_player_by_name(nil) does indeed crash in 0.4.14 dev builds, I just ran it 3 times and yes it does...  What version of minetest are you running ?

t4im:
> minetest.get_player_by_name(nil) does indeed crash

Oh, you're right. Nevermind that then.

tenplus1:
Explain whitespace changes please ?

t4im:
Your `loadhomes()` function for the most part only changed leading spaces/tabs and an unnecessary table assignment, so by reverting that, there would be no conflict at all.

tenplus1:
So this pull is ready to go ?? no more changes ??

t4im:
I asked you to remove the whitespaces changes of it, because there is no point to them. Not to merge my code from a different PR, that is independent from this one, because you did not want to address loading I/O issues.

So now you just force everyone to review my changes as part of your code and everyone yours *again* because of it. Which might just delay it a little further.

Meanwhile you keep ignoring the issues that are still open with your code for days:
https://github.com/minetest/minetest_game/pull/669/commits/0beef70676a25c0118c0d596cb80469a8dfb8ae4#r68559822

Which is the repeating reason this PR is being delayed so much in the first place.
Why make it harder for the maintainers to merge the code?

I would ask you to remove my changes to the `loadhomes()` function, so that this PR can get along faster and #1175 doesn't need to wait for it nor the other way around.

t4im:
> I'm struggling to learn github

I didn't consider this yet, but perhaps that is the core of the issue we're having.
There is a tab at the top "Files changed"



It should show you any active notes, next to the code, that is not considered "outdated" by github. This should usually include anything that is still a valid concern in the code.

I'm sorry if the problem just was, that you never saw the raised problems.
Maybe this helps in the future

tenplus1:
t4im - I don't understand most of what you're trying to say in the line comments cause it isn't being specific enough when pointing to a block of code...  Also, my original pull was to add global functions and make smaller save files...  What would have been a simple change has been taken over with your comments and when I DO make changes you still manage to find more ways for me to add or change or remove or whatever...  

Let the damn thing get pulled into the game already and THEN add your own amendments cause this is seriously annoying me now from what should have been a simple pull request.

paramat:
Yeah perhaps it's best t4im cleans this up later, that's more efficient than trying to explain their intended changes for someone else to do. Will review.

t4im:
I'm sorry that you are frustrated; i can even relate. I'm not doing this to annoy you.
But it is in your power to change this.
There is nothing wrong with ignoring suggestions by reviewers and solving it your way.
But if you add new code, then you have to deal with any new review comments about problems you might have introduced with it.
And we reviewers are hardly to blame for those, are we?

We could ignore them, but that doesn't like a good solution either:
Either your PR will be reviewed later and still declined/delayed by someone else, or issues that could have been prevented with maybe 1-2 additional lines, are added to the code base.
You would also be asking maintainers regularly to review unfinished code and mention these instead of letting them focus on getting PR's in.

So i honestly believe we're doing you and the minetest community a favor, even though i can fully understand, that you are frustrated by being pointed out so many problems. But that's why we point them out to you (and each other), so that they can be fixed quickly and the PR can get merged in safely and faster.

I'm doing my best to stick to important issues, like crashes, misbehavior and significant memory or cpu costly code. But i also will point out a few easy to fix things and best practices, like redundant code or simplification of code, when i'm already looking over it. (I wouldn't block a PR for those necessarily)
You haven't seen me nitpick about spelling mistakes, more commentary, better variable naming or whitespace corrections, have you? I'm just verbose; but i try to be verbose with reasons.

If you have suggestions on how to do handle programming issues in your PR's better, i'm open for suggestions. But maybe you should start seeing the chances of being reviewed by a professional coder instead of seeing only the frustration that comes with it. ;-)

t4im:
[Whitespaces](https://en.wikipedia.org/wiki/Whitespace_character) refer commonly to tabs and spaces. A whitespace change/commit refers to a change that involves changing mostly these, for example adding indentation or converting spaces to tabs.

Your `loadhomes()` was essentially just whitespace changes. Removing these would have resolved any minor conflict during merging with my PR. But even if you would not have done it, it would have been ok (it would just have been unnecessary). Copying my code over, after you kept insisting to not address the issues, was unnecessary and uncalled for. Please remove it again, so that the maintainers can proceed with merging both. Because believe it or not, i would like this over with, just as much as you do. ;-)

t4im:
Alright, I/O is ok now. With that the issues would finally be addressed.
Odd to take credit for someone elses work, but whatever.
The code seems ok now.

paramat:
@tenplus1 it's up to you if you are happy to revert whitespace changes or not, i don't blame you if you don't want to, and this PR has been waiting 9 months.
The code seems ok and has been tested thoroughly by yourself so :+1: with or without reverting whitespace changes, let me know if you are to do this or not.

paramat:
tenplus1 perhaps best not to attempt reverting whitespace changes, that can be tricky because it has to be exact to have any benefit. Conflicting PRs can be rebased.

t4im:
@paramat there are no whitespaces to revert, it was about asking him not to just take code and credit from #1175 because of whitespace "conflicts" after being difficult about it for so long. But the code is ok now, no pending code issues that would need attending.

paramat:
Ok.

PR #672, opened 2015-09-11
water mist

pinkysnowman:
Added water mist particle effects to enhance waterfalls and other water features.
screenshot_20150911_192515
screenshot_20150911_192504


VanessaE:
:+1:

VanessaE:
Some tweaks to make it even better:  http://pastebin.com/t4y28SDQ

adjusted the expiry times on the two spawner calls, and increased the particle count on the second (just flowing down a hill) one.

kilbith:
Nice, but not until we have the Irrlicht particles for obvious performance strangling.

mods/default/functions.lua
```diff
@@ -346,3 +346,59 @@ minetest.register_abm({
...
+--
+-- Water mist 
+--
+
+minetest.register_abm({
+	nodenames = {"default:water_flowing"},
+	interval = 5,
+	chance = 1,
+	action = function(pos, node, active_object_count, active_object_count_wider)
+		if pos.y < 0 then return end
```
PilzAdam: Why disable it underground?

pinkysnowman:
it is disabled as i have found some lakes to have only one layer of water_source on older builds and also caves under bodies of water can cause massive amounts of lag from the sheer amount of flowing water

PilzAdam:
Adding water_source to neighbors may help with performance.

pinkysnowman:
unfortunately i t cant be added, as it has some particles on flowing water that's just flowing over land.

VanessaE:
Yep, I can confirm the performance issues.   :frown:  

Too bad, too, as this PR adds a very nice effect.

pinkysnowman:
i have removed effects from flowing water and left the waterfall effects. I am still testing code for performance and will update asap!

paramat:
I feel this is best as a mod since it's a special visual feature that's triggered in rare circumstances.
An ABM on flowing water creates a processing load that in most worlds, most of the time, is doing nothing, i don't think this should be forced on everyone by default, for example i would rather not use it to improve performance.
Particles in Minetest are currently heavy, this feature creates 33 per node with collision detection.
This looks good under a waterfall, but in many situations, like a small stream of water running into a pool, spray like this would be unsuitable.

HybridDog:
you could make semi transparent particles

paramat:
Translucent textures are very buggy in Minetest due to no z-sorting, not recommended.

C1ffisme:
This looks pretty awesome, but as said before it can be laggy.

The particles could be a bit less bright, I think.

And a few of the particles should look a bit more like drips of water... That way it looks like water is splashing everywhere.

PilzAdam:
:-1: from me; I agree with @paramat.

mods/default/functions.lua
```diff
@@ -346,3 +346,40 @@ minetest.register_abm({
...
+	neighbors = {"default:water_source"},
+	interval = 5,
+	chance = 1,
+	action = function(pos, node, active_object_count, active_object_count_wider)
+		local posa = {x=pos.x, y=pos.y+1, z=pos.z}
+		local posu = {x=pos.x, y=pos.y-1, z=pos.z}
+		local na =  minetest.get_node(posa).name
+		local nu =  minetest.get_node(posu).name
+		if na == "default:water_flowing" then
+			if nu == "default:water_source" then
```
ShadowNinja: This indentation gets very deep.  You can reduce it by two levels by combining and inverting the conditionals like so:
```Lua
if na != "default:water_flowing" or nu != "default:water_source" then
        return
end
minetest.add_particlespawner...
```

ShadowNinja:
@paramat, @PilzAdam: The texture appears to be opaque.

mods/default/functions.lua
HybridDog: What does ```!=``` do?

pinkysnowman: != is doesn't equal in c/c++ i think SN  meant ~= but same thing just
different lang

ShadowNinja: You could do that too and increase the performance by a tad.

paramat:
Sure, i was responding to HybridDog's comment.

paramat:
:-1:

sfan5:
Very nice effect but it's too laggy/buggy in it's current state :-1:

paramat:
Closing due to 3 :-1:

PR #685, opened 2015-09-26
Flowers: Check whether under node is liquid.

ghost:


paramat:
Thanks for this, my code here may well need improving, i will consider improving it based on your PR.

paramat:
Looking at this PR i realised that documentation for 'on_place = function(itemstack, placer, pointed_thing)' is missing from lua_api.txt, which is why i ended up using 'after_place_node()'.
So i will rewrite my code, thanks for the help.

Rui914:
Should I close this PR?

paramat:
No, not  yet =)
I just tested your pull and found a problem, it places 2 waterlilies one above the other. Maybe you could fix this?
I tried rewriting my code but with no success.

Rui914:
Update.
But, this needs to fix coding-style.

paramat:
I tested this, it still has problems. I have now rewritten the code to work with river water.

Rui914:
Update.
How about this?

kilbith:

paramat:
Will test soon.

paramat:
Sorry this still has problems. Often a second waterlily will appear under the one on the surface, usually it disappears quickly but occasionally it will remain. Waterlilies can be placed on land and will then disappear after a while. The current code i wrote is a little hacky but behaves better. Thanks for your attempts though.

0-afflatus:
Rui's code certainly looks a lot cleaner.
Some obvious questions: 
This is about placing waterlilies on the surface of river water, yes? (I'm assuming this doesn't affect spawning)
And is it desirable that they should not be placeable anywhere else - on land, under or above existing lilypads?
Is this a generic problem for any node that can float / be placed on water - like seaweed? and potentially boats, buoys, little floating candles and rubber ducks etc.?

I'm thinking any item in group 'floats' should treat water as buildable_to.
Could that be done in lua?

PR #692, opened 2015-10-04
Default/book: textarea -> table

ghost:
screenshot_20151004_221819-min
#690

### DIFF
--- a/mods/default/craftitems.lua
+++ b/mods/default/craftitems.lua
@@ -30,7 +30,8 @@ local function book_on_use(itemstack, user, pointed_thing)
 		formspec = "size[8,8]"..default.gui_bg..
 			"label[0.5,0.5;by "..owner.."]"..
 			"label[0.5,0;"..minetest.formspec_escape(title).."]"..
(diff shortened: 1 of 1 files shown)


tenplus1:
+1

kilbith:

paramat:
+1 for the feature but i can't judge implementation another dev needs to.

Amaz1:
The only problem I see with this (which is an excellent fix!) is that lines are no longer wrapped, so if you want to be able to read things, you need to press enter after every line, not really a big deal though.

paramat:
Approval http://irc.minetest.ru/minetest-dev/2015-10-18#i_4427013

paramat:
3fb6ee5bdf7a8940d7e25c96f41436a49fea32fc

Ferk:
> lines are no longer wrapped

Wouldn't this break previous books?

Maybe we could try and do line wrapping manually?

Or actually make textarea scrollable if too big in the engine... which seems like the right thing to do. After all, I can't imagine a more appropiate use of textarea than blocks of text like the ones in books.

PR #717, opened 2015-10-31
Wrap book text

Rui914:
Note: This is ugly hack. This should support by engine.


PilzAdam:
This depends on ```font_size``` and other stuff; it should be properly done in the client. :-1:

C1ffisme:
IMHO someone just needs to properly improve the entire minetest formspec system. It needs all these hacks to make it work properly.

PR #736, opened 2015-12-07
Add a search field to the creative inventory

kilbith:
You'll no longer waste your time browsing several pages to find something.
Obviously gives a significant help to the creative activity.

screenshot_20151129_222037

screenshot_20151129_221146

### DIFF
--- a/mods/creative/init.lua
+++ b/mods/creative/init.lua
@@ -1,11 +1,16 @@
 -- minetest/creative/init.lua
 
 creative_inventory = {}
-creative_inventory.creative_inventory_size = 0
 
 -- Create detached creative inventory after loading all mods
-minetest.after(0, function()
-	local inv = minetest.create_detached_inventory("creative", {
+creative_inventory.init_creative_inventory = function(player)
+	local player_name = player:get_player_name()
+	creative_inventory[player_name] = {}
+	creative_inventory[player_name].size = 0
+	creative_inventory[player_name].filter = nil
+	creative_inventory[player_name].start_i = 1
+
+	local inv = minetest.create_detached_inventory("creative_" .. player_name, {
 		allow_move = function(inv, from_list, from_index, to_list, to_index, count, player)
 			if minetest.setting_getbool("creative_mode") then
 				return count
@@ -35,19 +40,29 @@ minetest.after(0, function()
 			end
 		end,
 	})
+
+	creative_inventory.update(player_name, nil)
+	--print("creative inventory size: "..dump(creative_inventory.creative_inventory_size))
+end
+
+function creative_inventory.update(player_name, filter)
 	local creative_list = {}
-	for name,def in pairs(minetest.registered_items) do
-		if (not def.groups.not_in_creative_inventory or def.groups.not_in_creative_inventory == 0)
-				and def.description and def.description ~= "" then
-			creative_list[#creative_list+1] = name
+	local inv = minetest.get_inventory({type = "detached", name = "creative_" .. player_name})
+
+	for name, def in pairs(minetest.registered_items) do
+		if not (def.groups.not_in_creative_inventory == 1) and
+				def.description and def.description ~= "" then
+			if (filter and def.name:find(filter, 1, true)) or not filter then
+				creative_list[#creative_list+1] = name
+			end
 		end
 	end
+
 	table.sort(creative_list)
 	inv:set_size("main", #creative_list)
 	inv:set_list("main", creative_list)
-	creative_inventory.creative_inventory_size = #creative_list
-	--print("creative inventory size: "..dump(creative_inventory.creative_inventory_size))
-end)
+	creative_inventory[player_name].size = #creative_list
+end
 
 -- Create the trash field
 local trash = minetest.create_detached_inventory("creative_trash", {
@@ -66,10 +81,12 @@ local trash = minetest.create_detached_inventory("creative_trash", {
 })
 trash:set_size("main", 1)
 
-
 creative_inventory.set_creative_formspec = function(player, start_i, pagenum)
+	local player_name = player:get_player_name()
+	local filter = creative_inventory[player_name].filter or ""
 	pagenum = math.floor(pagenum)
-	local pagemax = math.floor((creative_inventory.creative_inventory_size-1) / (6*4) + 1)
+	local pagemax = math.floor((creative_inventory[player_name].size - 1) / (6*4) + 1)
+
 	player:set_inventory_formspec(
 			"size[13,7.5]"..
 			--"image[6,0.6;1,2;player.png]"..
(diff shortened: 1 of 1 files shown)


Jordach:
:+1: This makes sense - why should I (depending on the set of mods), sort
through nearly 30 pages for one block or item? I would like to see tabs -
eg, tools, blocks, misc. This would help in searching faster.

mods/creative/init.lua
```diff
@@ -35,19 +40,29 @@ minetest.after(0, function()
...
-	for name,def in pairs(minetest.registered_items) do
-		if (not def.groups.not_in_creative_inventory or def.groups.not_in_creative_inventory == 0)
-				and def.description and def.description ~= "" then
-			creative_list[#creative_list+1] = name
+	local inv = minetest.get_inventory({type = "detached", name = "creative_" .. player_name})
+
+	for name, def in pairs(minetest.registered_items) do
+		if not (def.groups.not_in_creative_inventory == 1) and
+				def.description and def.description ~= "" then
+			if (filter and def.name:find(filter)) or not filter then
```
kaeza: This `find` call searches for regular expressions instead of plain substrings. If this is intended, you have to wrap the `find` call in `pcall` and fail gracefully in case of errors. If not intended, change to `find(filter, 1, true)`.

mods/creative/init.lua
```diff
@@ -152,18 +185,18 @@ if minetest.setting_getbool("creative_mode") then
...
-	
+
 	function minetest.handle_node_drops(pos, drops, digger)
 		if not digger or not digger:is_player() then
 			return
 		end
 		local inv = digger:get_inventory()
 		if inv then
-			for _,item in ipairs(drops) do
+			for _, item in pairs(drops) do
```
kaeza: Why the change to `pairs`?

kilbith: `ipairs` is completely useless when you don't (need to) iterate over tables with sequential numeric indices. It is the case here.

mods/creative/init.lua
kilbith: It is intended, yes. Sometimes I don't bother to type the whole name to find something, this is more convenient IMO.

Also how it can have errors ? There's always a filtering even with one letter.

kaeza:
```
# lua
LuaJIT 2.0.2 -- Copyright (C) 2005-2013 Mike Pall. http://luajit.org/
JIT: ON CMOV SSE2 SSE3 fold cse dce fwd dse narrow loop abc sink fuse
> print(("s"):find("%"))
stdin:1: malformed pattern (ends with '%')
stack traceback:
        [C]: in function 'find'
        stdin:1: in main chunk
        [C]: at 0x00401ec0
> print(("s"):find("%", 1, true))
nil
> print(("s"):find("s", 1, true))
1       1
>
```

mods/creative/init.lua
kaeza: `ipairs` makes it clear to the programmer that the table in an array.

kilbith: You're nitpicking here... don't use knowingly something slower when you don't need to.

mods/creative/init.lua
kilbith: You're right, will change that.

mods/creative/init.lua
kaeza: >slower

```
> iter,c,n,t = pairs,0,5,{"a","b","c"} for it=1, n do s=os.clock() for x = 1, 100000 do for a, b in iter(t) do end end c=c+os.clock()-s end print("Average:", c/n)
Average:        0.020999999999992
> iter,c,n,t = ipairs,0,5,{"a","b","c"} for it=1, n do s=os.clock() for x = 1, 100000 do for a, b in iter(t) do end end c=c+os.clock()-s end print("Average:", c/n)
Average:        0.0032000000000039
```
Still don't see a reason for the change, but won't argue further.

C1ffisme:
This is especially useful on laggy servers, I can't wait to see this (hopefully) merged.

JBBgameich:
I like Jordans Idea of tabs! But i think we need a search too. I also would like to see a better way of getting the items from the crative inventory.  An idea would be if you leftclick on an item, you get 1 of it, and when rightclick on an item you get a full stack. 
-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

C1ffisme:
One problem about search bars, though (I'm not sure if you can do anything about this though. :( ) : Every time I use one, I'm used to pressing the enter key to search. But this closes my formspec, and I have to type again. Not a problem for me, but to people who point their finger straight downwards and poke each key individually to type, this can get annoying.

Jordach:
@C1ffisme It appears the Unified Inventory doesn't have this issue. Perhaps
fiddling with text inputs might work.

kilbith:
Kinda surprised that none of the maintainers reacted yet...

@PilzAdam @ShadowNinja @Ekdohibs @sfan5  -> a review please ?

PilzAdam:
+1 for the feature, but I currently have no time to look at the code.

mods/creative/init.lua
HybridDog: >Still don't see a reason for the change, but won't argue further.

it takes 556.25% more time

HybridDog:
>[the enter key] closes my formspec

maybe you only need to change the order of the formspec elements,
What happens if you press tab and then enter?

mods/creative/init.lua
kaeza: So in my case, `ipairs` is 5.5 times faster than `pairs`, and in HybridDog's case, `ipairs` is 3.8 times faster. My benchmark is also with LuaJIT, BTW.

But let's change it anyway because...reasons! (not specifically directed to you HybridDog)

Ekdohibs:
:+1: for the feature, code looks good but I haven't reviewed it in detail.

mods/creative/init.lua
kilbith: Changed it back to interrupt the bikeshedding.

0-afflatus:
:+1: subject to code review; it's a no-brainer really.

PilzAdam:
The problems with the enter key that @C1ffisme pointed out are limitations of the engine. I'm adding the 'need-engine-change' label as a reference, but it shouldn't block this PR.

ShadowNinja:
Engine change: minetest/minetest#516

kilbith:
@PilzAdam O.K. for merging then ? Please do, otherwise nobody else will bother to (not even for reviewing).

kaeza:
>otherwise nobody else will bother to (not even for reviewing).

Offtopic for this discussion, but the issues/PRs keep piling up in this repo with basically nobody ("collaborators" in Github speak) saying yay or nay or even bother to review code, much less to bother merging in the cases where there's a clear agreement. Maybe there's a need to cycle game maintainers?

HybridDog:
>Maybe there's a need to cycle game maintainers?

Actually the repo's master branch is used for the dev version of minetest, isn't it?
So how about adding secondary maintainers merging PRs and if they are unwanted by enough primary ones, the commits get reverted?

sofar:
Reviewed the code, tested it, and seems fine. It creates a proper detached inventory.

I don't see why "needs engine change" is needed at all, it is certainly not a blocking requirement for this patch, and IMO should *not* prevent this from getting merged. The change works, is usable and functions exactly like unified_inventory's search field, so I doubt that anyone would be confused to begin with.

:+1: from me

paramat:
Seems to have enough approval and review, will merge soon (tomorrow).

kilbith:
Thanks you @paramat and @sofar. I appreciate your consideration.

paramat:
b1e2ece63842ff5228798feb9259398c75524f8b

PR #751, opened 2015-12-22
changed the formspec method

pinkysnowman:
switched to table.concat() for faster loading

### DIFF
--- a/mods/creative/init.lua
+++ b/mods/creative/init.lua
@@ -70,29 +70,30 @@ trash:set_size("main", 1)
 creative_inventory.set_creative_formspec = function(player, start_i, pagenum)
 	pagenum = math.floor(pagenum)
 	local pagemax = math.floor((creative_inventory.creative_inventory_size-1) / (6*4) + 1)
-	player:set_inventory_formspec(
-			"size[13,7.5]"..
-			--"image[6,0.6;1,2;player.png]"..
-			default.gui_bg..
-			default.gui_bg_img..
-			default.gui_slots..
-			"list[current_player;main;5,3.5;8,1;]"..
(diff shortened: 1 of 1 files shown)


kaeza:
Conflicts with #736, and is not needed for an operation that is not executed very often.

asl97:
table.concat is slower than joining a bunch of string **without needing to store them in a variable***

unlike the sethome, the amount of args doesn't grow ether, it will always be the same amount of args

there no reason for this, this is slower than the current code.

pinkysnowman:
I was under the impression ".."  is slower

kilbith:
:-1: `..` is already a concatenation meant for plain strings, you're basically using needlessly more ressources with `table.concat` (which is useful to use when your values are necessarily in an array and that is likely to be dynamically redefined).

The best way for optimizing this is to avoid the concatenation at all, i.e. make a long, continuous string (but you lost the readability).

Also rather bad-style deontologically to open a PR conflicting short after another.

pinkysnowman:
Fuck you cock knob,  i  don't need to hear your shit.  I  have my issues
and coding helps me fucking cope,  maybe next time ill come cut your arm
leg and wrist instead of my own!  Or cut off that pitiful thing you call a
dick.

pinkysnowman:
And you guys wonder why noone gives a fuck about deving this engine

asl97:
@pinkysnowman now that is just rude, what kilbith said is valid

@kilbith
* if 'newline' doesn't effect formspec, `[[]]` could be use instead of long continuous string.
* pinkysnowman might not have known about that PR

est31:
I've removed a particularly rude comment.

0-afflatus:
I have to say I find use of terms like 'deontologically' somewhat egregious.

PR #752, opened 2015-12-22
speed up saving of homefile

pinkysnowman:
switched to table.concat() and only use output:write() one time for speed

### DIFF
--- a/mods/sethome/init.lua
+++ b/mods/sethome/init.lua
@@ -57,7 +57,7 @@ minetest.register_chatcommand("sethome", {
         	local output = io.open(home_file, "w")
 			local data = {}
 			for n, p in pairs(homepos) do
-				table.insert(data,(p.x.." "..p.y.." "..p.z.." "..n.."\n")
+				table.insert(data,(p.x.." "..p.y.." "..p.z.." "..n.."\n")) 
 			end
 			output:write(table.concat(data))
 			io.close(output)
--- a/mods/sethome/init.lua
+++ b/mods/sethome/init.lua
@@ -54,11 +54,13 @@ minetest.register_chatcommand("sethome", {
         minetest.chat_send_player(name, "Home set!")
         changed = true
         if changed then
-        	local output = io.open(homes_file, "w")
-            for i, v in pairs(homepos) do
-                output:write(v.x.." "..v.y.." "..v.z.." "..i.."\n")
-            end
-            io.close(output)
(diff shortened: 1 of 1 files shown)


kaeza:
Conflicts with #736, and is not needed for an operation that is not executed very often.

mods/sethome/init.lua
```diff
@@ -54,11 +54,13 @@ minetest.register_chatcommand("sethome", {
         minetest.chat_send_player(name, "Home set!")
         changed = true
         if changed then
-        	local output = io.open(homes_file, "w")
-            for i, v in pairs(homepos) do
-                output:write(v.x.." "..v.y.." "..v.z.." "..i.."\n")
-            end
-            io.close(output)
+        	local output = io.open(home_file, "w")
+			local data = {}
```
asl97: why the indent?

mods/sethome/init.lua
```diff
@@ -54,11 +54,13 @@ minetest.register_chatcommand("sethome", {
...
+			for n, p in pairs(homepos) do
+				table.insert(data,(p.x.." "..p.y.." "..p.z.." "..n.."\n")
```
asl97: missing closing bracket?

mods/sethome/init.lua
pinkysnowman: I think my editor did that opps lol

asl97:
@kaeza it does?  i don't see how this would conflict with #736.

kaeza:
@asl97 see edit. I blame OctoDroid for bringing the wrong pull to top :P

mods/sethome/init.lua
pinkysnowman: @kaeza just because the code is not ran alot doesn't mean that it can't be
improved upon ;)

asl97:
@kaeza you can delete your post yourself by clicking the X button on the top right corner.

mods/sethome/init.lua
asl97: @pinkysnowman imo, it would be better to leave a reply for the pull request on the pull request rather than in the note

pinkysnowman: Sorry on my mobile if i reply it does that

kaeza:
#747

mods/sethome/init.lua
ShadowNinja: This is indented with a mix of spaces and tabs.

PR #759, opened 2015-12-27
Place tree logs as decorations.

sofar:
We can vary the landscape a bit more by placing "fallen logs"
around the various forests. These decorations are quite fast
and will provide some gameplay value but are still more rare
than the corresponding trees, so they don't provide free
materials.

I've manually put the schematic as lua tables since these log
schematics are only 8 blocks. We vary the log lengths between
2 and 4 blocks by making the end blocks have a lower chance
of appearing.

Amount is varied by fill_ratio, except for acacia trees where
we reduce the scale, so that acacia logs show up near places with
acacia trees consistently.

### DIFF
--- a/mods/default/mapgen.lua
+++ b/mods/default/mapgen.lua
@@ -967,7 +967,7 @@ end
 function default.register_decorations()
 	minetest.clear_registered_decorations()
 
-	-- Apple tree
+	-- Apple tree and log
 
 	minetest.register_decoration({
 		deco_type = "schematic",
@@ -988,7 +988,33 @@ function default.register_decorations()
 		flags = "place_center_x, place_center_z",
 	})
 
-	-- Jungle tree
+	minetest.register_decoration({
+		deco_type = "schematic",
+		place_on = {"default:dirt_with_grass"},
+		sidelen = 80,
+		fill_ratio = 0.0015,
+		biomes = {"deciduous_forest"},
+		y_min = 1,
+		y_max = 31000,
+		schematic = {
+			size = { x = 3, y = 3, z = 1},
+			data = {
+				{ name = "air", prob = 0 },
+				{ name = "air", prob = 0 },
+				{ name = "air", prob = 0 },
+				{ name = "default:tree", param2 = 12, prob = 191 },
+				{ name = "default:tree", param2 = 12 },
+				{ name = "default:tree", param2 = 12, prob = 127 },
+				{ name = "air", prob = 0 },
+				{ name = "flowers:mushroom_brown", prob = 63 },
+				{ name = "air", prob = 0 },
+			},
+		},
+		flags = "place_center_x",
+		rotation = "random",
+	})
+
+	-- Jungle tree and log
 
 	minetest.register_decoration({
 		deco_type = "schematic",
@@ -1003,7 +1029,33 @@ function default.register_decorations()
 		rotation = "random",
 	})
 
-	-- Taiga and temperate coniferous forest pine tree
+	minetest.register_decoration({
+		deco_type = "schematic",
+		place_on = {"default:dirt_with_grass", "default:dirt"},
+		sidelen = 80,
+		fill_ratio = 0.01,
+		biomes = {"rainforest", "rainforest_swamp"},
+		y_min = 1,
+		y_max = 31000,
+		schematic = {
+			size = { x = 3, y = 3, z = 1},
+			data = {
+				{ name = "air", prob = 0 },
+				{ name = "air", prob = 0 },
+				{ name = "air", prob = 0 },
+				{ name = "default:jungletree", param2 = 12, prob = 191 },
+				{ name = "default:jungletree", param2 = 12 },
+				{ name = "default:jungletree", param2 = 12, prob = 127 },
+				{ name = "air", prob = 0 },
+				{ name = "flowers:mushroom_brown", prob = 127 },
+				{ name = "air", prob = 0 },
+			},
+		},
+		flags = "place_center_x",
+		rotation = "random",
+	})
+
+	-- Taiga and temperate coniferous forest pine tree and log
 
 	minetest.register_decoration({
 		deco_type = "schematic",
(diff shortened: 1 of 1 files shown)


paramat:
The idea is good and works well on flat ground, but as you pointed out this has problems with placement on slopes http://i.imgur.com/lVy8tCH.jpg http://i.imgur.com/kN2ifWM.png the way the log and dirt projects out of a slope is ugly. This is perhaps best as a mod, which can then do some scanning of ground to find a suitable place.

paramat:
Maybe it would work better with logs 3 nodes long with a single central dirt node uunderneath? This would stop the dirt projecting, then have probability on each end of the log, or just one end, to create 2 node logs.

sofar:
Doing this entirely in lua isn't something I'd prefer at this point. Moretrees already does this, so there's almost no benefit. If we can't do this as a decoration, I'll just pass on it. Looking at generate_nyancats() makes my head hurt.

But let's see if a max log length of 3 and some ground level tweaking may help.

sofar:
Shortened to max 3 logs. I've slightly increased the chance that the logs will be 2 or 3 long, instead of 1.

sofar:
None of the logs I can find on cliffs with the 3x logs look weird to me. Yes they're logs on cliffs, but they don't stick out as bad as the one screenshot I posted earlier.

VanessaE:
Actually, it's the trunks mod in plantlife modpack that does the fallen logs, via biome_lib's functions.

sofar:
@VanessaE and after reading that code, I really wouldn't want to do this in lua...

VanessaE:
Yeah it's a bear to do any kind of terrain modification from Lua unless you're using something like biome_lib.  While it does most of the hard work, it can't do it all. :-)

C1ffisme:
> this has problems with placement on slopes
I feel like there should be a flag in ```minetest.register_decoration()``` for this, like "only_place_on_ground". This might cause problems with larger builds, but for small objects like this, it is almost essential.

paramat:
Nope. Code to recognise flat ground will be far too complex for the decoration API.

paramat:
>  If we can't do this as a decoration, I'll just pass on it.

Yes i agree.
WIll test this.

sofar:
I've looked at mg_decorations.cpp, and I think it's possible to scan the lowest yslice of the schematic and do some testing on the vm where it's going to be placed. We could add a flag to the data array and have some mgflag "BLOCK_MUST_EXIST" flag set in each schematic block where we want it to test.

that would allow us to create a schematic like this:

```
size = { x = 3, y = 2, z = 1}
data = {
    { name = "group:dirt", block_must_exist = true },
    { name = "group:dirt", block_must_exist = true },
    { name = "group:dirt", block_must_exist = true },
    { name = "default:tree" },
    { name = "default:tree" },
    { name = "default:tree" },
}
```

thus allowing the schematic to only ever to be placed on a flat surface...

I think i'ts possible, and would be generic enough. Scanning only the bottom yslice wouldn't be too much of a performance hit either.

sofar:
BTW having that sort of testable mechanism would allow the default map decorator to place entire houses, shrubs and other content that has issues if they're not placed on flat ground. That would seem worthwhile.

paramat:
Interesting will consider, but it may not be needed enough to be worth the effort, the decoration API is meant to be as simple as possible for placing just basic flora, logs are a more specialised use.

C1ffisme:
@paramat 6 blocks is too big for the decoration API? We aren't building entire villages here, but this just seems ridiculous.

paramat:
That wouldn't ensure flat ground, as it could be the centre of a depression or a hole.

sofar:
right

paramat:
Now testing, apart from some logs levitating on the side of cliffs i quite like it. Apart from savanna the fill ratio should be a little lower. I don't know why these happen:

screenshot_20160104_020916

paramat:
Since the schematic has a z dimension of 1 the 'place centre z' flag may not be needed, it might even help the odd placement.
Another idea, perhaps the dirt node underneath is not needed? (maybe it is) the levitating logs would look better without them.
One type of mushroom on top would be nice, whichever is more realistic for growing on logs.
The occasional log on a cliff looks okay and what you would expect in a forest, logs can fall down cliffs and get lodged in cracks.
The per-node force-placement is false by default so you only need specify it when true.
When we're happy with the schematic it should be converted into a file using the method shown here https://github.com/paramat/saveschems
There you can see the conventions for a lua table schematic, for 'air' not force placed the convention is:
```
{name="air", prob=0},
```

sofar:
Good points, I'll go and experiment a little bit with air at the bottom as well.

I would think that schematic loading from lua is faster if external parsing isn't done from file and it's just parsed from a table... I know how to save schematics but when they're this small they just clutter?

sofar:
![cavegen]
(http://i.imgur.com/6aLgWR6.png)

air as bottom layer works fine. however (see image) I do have a floating log here over a cave. Is cavegen active before or after decorations?

sofar:
Then again, here's a floating tree.

![bug]
(http://i.imgur.com/AHbikpu.png)

sofar:
@paramat, I'm inclined to say that floating decorations are a mapgen bug :^)

paramat:
> I'm inclined to say that floating decorations are a mapgen bug

Did removing 'place centre z' help with the odd placement of logs?

Yes i'm fine with defining small schematics as tables, i guess making them files is just a habit because i was asked to do that by other MTGame devs before i was a MTGame dev. So it's fine as it is.

Decorations are placed after cavegen but sometimes caves can overgenerate from neighbour mapchunks and remove ground from underneath. The occasional floating decoration doesn't bother me it's cute.

I think the logs look better without the dirt node underneath when floating, smaller schematic too.

sofar:
- dirt nodes underneath are all removed
- mushrooms are placed on a log with low probability
- brown mushrooms on apple, jungle tree
- red mushrooms on pine
- no mushrooms in savannah's since that's a dry biome, so nothing on acacia logs
- reduced log fill rates by 50% or more.

paramat:
Nice will test again

paramat:
A perfectionist request, please could every air node in th eschematics be:
```{ name = "air", prob = 0 },```

paramat:
Another tip, for decorations that use 'fill ratio' instead of noise, use:
```sidelen = 80,```
instead of 16, it speeds up placement, there's no need to subdivide the mapchunk when there's no variation by noise.

sofar:
ok, changed air to be ```prob = 0``` and ```sidelen = 80``` for when ```fill_ratio``` is used.

paramat:
Will review in-game again soon.

paramat:
Approved by sfan5 http://irc.minetest.ru/minetest-dev/2016-01-05#i_4494718

paramat:
I''ll probably approve this soon just want to test first.

paramat:
:+1:

sofar:
that *totally* explains the floating apple tree image above!

paramat:
Ahhhh.

sofar:
retesting... can't hurt ;)

sofar:
once this is in I'll patch Aspen trunks with red mushrooms, I'm thinking.

sofar:
Scanned over a few km's of landscape, and can't see any more floating apple trees either.

paramat:
:+1: Will merge soon.

paramat:
57939c6c3f78530cde376a603ff35e157b0749be

PR #766, opened 2016-01-01
on_burn callback for nodes to handle burning

cmdskp:
Allows nodes to define on_burn(position, igniter_node_name, action_string) to handle different(future) burn actions (E.g. allows nodes to change burnt nodes to a charred/burnt version or reduce chance of removal, rather than just remove equally).

Optimised by removing duplicate call to find_node_near before removing flammable nodes.
Improved readability with meaningful variable names and some added comments.
Provided fire.features.on_burn_callback (Thanks, Celeron55!) to allow mods to save recreating functionality with their own burning ABMs by testing fire.features.

### DIFF
--- a/mods/fire/init.lua
+++ b/mods/fire/init.lua
@@ -3,7 +3,12 @@
 -- Global namespace for functions
 
 fire = {}
-
+fire.features = {
+	on_burn_callback = true, -- When groups.flammable node about to be burnt, calls:
+				 -- node.on_burn(position, source_node_name, action_string)
+				 -- `action_string` = "remove"
+				 -- Return: true, stops the node being removed.
+}
 
 -- Register flame nodes
 
@@ -172,10 +177,10 @@ minetest.register_abm({
 	interval = 3,
 	chance = 2,
 	catch_up = false,
-	action = function(p0, node, _, _)
-		minetest.remove_node(p0)
+	action = function(position, node, _, _)
+		minetest.remove_node(position)
 		minetest.sound_play("fire_extinguish_flame",
-			{pos = p0, max_hear_distance = 16, gain = 0.25})
+			{pos = position, max_hear_distance = 16, gain = 0.25})
 	end,
 })
 
@@ -191,13 +196,11 @@ if minetest.setting_getbool("disable_fire") then
 		interval = 7,
 		chance = 2,
 		catch_up = false,
-		action = function(p0, node, _, _)
-			minetest.remove_node(p0)
+		action = function(position, node, _, _)
+			minetest.remove_node(position)
 		end,
 	})
-
 else
-
 	-- Ignite neighboring nodes, add basic flames
 
 	minetest.register_abm({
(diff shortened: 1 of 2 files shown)


everamzah:
I like to leave fire enabled and this would make me feel a lot better about doing so.

mods/fire/init.lua
```diff
@@ -172,10 +177,10 @@ minetest.register_abm({
 	interval = 3,
 	chance = 2,
 	catch_up = false,
-	action = function(p0, node, _, _)
-		minetest.remove_node(p0)
+	action = function(position, node, _, _)
```
asl97: imo, a pull request about adding an on_burn callback shouldn't mess with unrelated stuff like variable renaming.

asl97:
imo, a pull request about adding an on_burn callback shouldn't mess with unrelated stuff like variable renaming.

cmdskp:
Including clearer, meaningful variable names was necessary to ensure understanding as part of the new feature (e.g <p> to <flammable_position>).  For code consistency, similar variables were renamed.

Reverting those renames to the previous p0 and p, before the pull request, could introduce easily missed errors on a single character missed and would require additional testing passes.

asl97:
specifically, what i meant was that a commit with a name like `(add) on_burn callback for nodes to handle burning` shouldn't be touching unrelated parts of the code but that just my opinion which doesn't hold much if any weight on whether the maintainers would merge this.

cmdskp:
IMHO, overall codebase quality should always be made consistent with newer changes, esp. in this case, where the existing variable names were prone to cause confusion and errors in the new(and existing) logic that are related by being part of a single unit file.

mods/fire/init.lua
```diff
@@ -3,7 +3,12 @@
 -- Global namespace for functions
 
 fire = {}
-
+fire.features = {
```
asl97: what's the point of the fire.features?

asl97:
which i agree, however, imo, such changes should be in their own commit so that the git history would be searchable and meaningful, it would also mean smaller manageable diff which does what the commit name say and nothing more.

mods/fire/init.lua
cmdskp: It provides a means for other mods to determine if they can use the newer Fire feature on_burn callback, by checking this (as opposed to using a version number).

This allows mods to save adding a custom ABM for burning nodes that the mod wishes to turn into a charred version(for example).  It also provides easy and meaningful feature checking of the Fire API present, as opposed to using a version number check.  It was suggested by celeron55 as a better alternative.

asl97: most mod just support version of minetest (and minetest game) which have the features they need and crash when it doesn't.

iirc, there no way to override an abm and trying to support older version without this feature by adding an abm would just make it an abm race to see which get the node first

the only way is for them to override the abm in the code on their server when using an older version,

the only thing useful about this which i can think of is it let mod print a pretty message before exiting instead of crashing

cmdskp:
That is true with extra time & resources.  There would be additional work and tests to ensure the one job is split into two correctly after the renames happened during development to ensure clarity of testing & understanding and correct functioning.

mods/fire/init.lua
cmdskp: Here's an example to help see the benefit to it:

A mod wishes to add lava causing rails(they have wooden tie/sleepers) to burn into twisted/burnt rails that lose there function, but return some steel ingots on digging them up.

Currently, if that mod sets groups.flammable on the rail, the default Fire mod will completely remove the rail.  So, the mod has no choice but to create it's own ABM to trigger on rails(which, by default, aren't flammable and normally left untouched by Fire) to transform them.  This means an extra ABM which could be avoided now with the features.on_burn_callback check.

The mod can check that and add it's own custom ABM to burn nodes(that normally aren't) to a burnt version or if the features feature is available on the server (from default game being upgraded), the mod can now instead handle the transform with the callback without needing to add an ABM any more, improving server performance.

I hadn't added to the game_api.txt before it was requested later - I'd included comments to the Fire mod itself for ease-of-access and folder copying into various games.  It's useful to explain in the Fire lua, for those who just want to check out what it does, without looking for a higher-level document that they may not be aware exists.

PR #767, opened 2016-01-01
Added Fire API section for on_burn

cmdskp:
See feature Fire on_burn pull-request: https://github.com/minetest/minetest_game/pull/766


asl97:
this should be together with #766 in a single pull request

cmdskp:
Sorry, attempting to edit it in with #766 now...will close after success.

PR #771, opened 2016-01-05
Aspen trees.

sofar:
Adds a birch-like tree to the default_game. Aspen was chosen
on purpose instead of birch, as several birch tree mods already
exist and choosing a different name avoids any conflicts.

Schematics were made for both normal and sapling version,
assuring saplings will not be able to grief. The shape of
the Aspen is "fanning out at the top" and provides an easy
tree to walk under, but still a somewhat thick cover.

Textures were painted from scratch, except tree_top and _wood
variants, which are color modified versions of the pine_wood
variants. Appropriate licenses have been chosen for the new
textures.

The leaf texture is light enough to contrast other trees, but
dark enough to offset against our light default grass color. The
leaves are drawn in the typical minetest default fashion with plenty
of transparancy, but enough definition to suggest that you're
seeing something that looks like leaves. The placement of leaves
in the schematic also suggests the top of the tree is sparse
and you can see the sky through the leaves.

Sapling texture is both traditional and different, with lush
green leaves and a well-defined stem, but slightly stick-like
and skinny, as these plants tend to grow up first, then out.

Add stairs, slabs for these wood types as well.

Mapgen will place these trees in both coniferous and deciduous
forests, but the seed is somewhat different and will provide some
variation to the otherwise mostly monotonous forest landscapes.
The placement of Aspen is patchy, which means that large areas
of these two biomes will not have any Aspen, but when you do find
them, they will be interspersed with either Pine trees or Apple
trees.


sofar:
Obligatory screen shot.

![Aspen Trees]
(http://i.imgur.com/b8HOzw2.jpg)

MoNTE48:
Nice! +1

kaeza:
:+1:

kilbith:
:ok_hand:

Ekdohibs:
Looks good!

sofar:
Lowered the perlin noise scale to remove most of the double trees - density was way too high. With 0.02 I can't find any immediate double trees and everything looks appropriate.

Fixer-007:
+1

RealBadAngel:
thats what tigers like the most. +1

paramat:
Nice pale wood, it complements the existing wood colours. Good textures.

paramat:
What horizontal size is the aspen, 5x5? this affects the placement offset:
```minetest.place_schematic({x = pos.x - 4, y = pos.y - 1, z = pos.z - 4},```
For 5x5 it should be pos.x - 2, pos.z - 2.

**Comment by sofar** [comment]
@paramat Yes, I specifically spend a lot of time comparing texture "texture", color, hue, lightness, style and shapes. I also assured to put all the woods and saplings side-by-side to make sure none of them appear too similar (which would be confusing) and provide an appealing contrast, yet something you'd likely combine in a craft.

I find the other woods to be quite colorful, so for this one I chose to desaturate the texture significantly.

**Comment by sofar** [comment]
@paramat 5x5, yes, so that's wrong. I'll fix the placement and re-verify.

**Comment by sofar** [comment]
Didn't mention this in the PR but Aspen trees are 6-8 trunks long, so end up 7-9 blocks tall. That's slightly higher than apple trees, and lower than Pine trees.

**Comment by sofar** [comment]
@paramat looking at it, I also do not need to randomly rotate it, so that parameter can be removed.

Will resubmit later when I get a chance to test.

**Comment by sofar** [comment]
if #759 gets merged I'll update this to include Aspen trunks as well.

**Comment by paramat** [comment]
759 will be merged so you can start updating.

Just to check the structure of your schematics:

The mapgen tree should use per-node force-place on all trunk and branches except the lowest trunk node (the root in the ground) because that leaves ugly holes when players remove mapgen trees.

The sapling-grown tree should use per-node force-place on the second trunk node up from base only, this is the node that replaces the sapling.

Y-slice probability should only be used on slice 2 and higher (3rd slice up), otherwise it's possible to skip either the not-force-placed root node (mapgen tree) or the force-placed sapling-replacing node (sapling tree).

**Comment by sofar** [comment]
I made my schematics as a modification to save_schems, so I can publish them, and you can review. I'll post this in a bit.

**Comment by sofar** [comment]
Here is the diff against saveschems:

https://gist.github.com/sofar/42ab373e34146d41e638

I created them with all the caveats you quoted, so they should be all ready to go. I reformatted the tables so it's really easy to read them.

**Comment by sofar** [comment]
Ok, I'll wait until #759 gets pushed before I update this branch, as I need to rebase and it's not an easy rebase. I have it done, but I'll keep it local.

I ended up reducing the Aspen logs dramatically compared to the others, and making them 2-3 long instead of 1-3 long. This allows me to place both red and brown mushrooms on top of them, which seems the most balanced option - otherwise you'd end up with either all red mushrooms on trunks only in pine forests, or brown in apple tree forests.

**Comment by paramat** [comment]
Ok the schems are fine, nice format too.

**Comment by paramat** [comment]
About tree location.

I would prefer these in deciduous forest only, even if that's unrealistic, i have a 'thing' about pure coniferous forests, maybe from experiences as a child. I also think they look better pure as a sea of pointy tops. Appletree forests however feel like they need some variation.

Currently aspen trees are denser where other trees are denser, not ideal, here's the trick: use the same seed, octaves, persistence but make the scale negative, then aspen is denser where appletrees are less dense.

**Comment by sofar** [comment]
no problems with just having them in apple tree forests, as they're most abundant and do need the variation most. I like the idea of the inverse scale, will see if that works.

**Comment by sofar** [comment]
Had to significantly change the apple tree and log placement, as the biome was already over populated as it was, and there was barely space for more stuff.

I reduced offset/scale to create space in the least densely areas mostly, leaving dense apple forests mostly unaffected. This creates enough room for aspen forests that are (from what I've seen) 90% aspen or more.

I left the aspen trees and logs less dense than apple trees, so they nicely complement and weave a bit into eachother, and provide a bit of a clearing in between, although barely. At no point have I seen anything that looks like a field.

Have a look, your taste is likely a bit different.

**Comment by sofar** [comment]
A Typical transition zone from Aspen to Apple

![image]
(http://i.imgur.com/a3cs82V.jpg)

**Comment by paramat** [comment]
I agree that appletrees needed to create more space, so the 2 distributions are more equal or equal. I'll test some parameters.

**Comment by paramat** [comment]
Also of course the apple and aspen logs will now need to be placed by noise to locate them near the corresponding trees.

**Comment by sofar** [comment]
That was done in the last patch. Without that change, it's an obvious uncorrelated mess of logs in areas where they are out of place.

**Comment by red-001** [comment]
@sofar the aspen texture doesn't look right to me, it feels too repetitive and very regular. I made two alternative textures that looks better in my option.

**Comment by sofar** [comment]
Your screenshots have bumpmapping enabled. With it disabled you have odd aliasing since you seemed to have copied a row of pixels to the adjacent row? Your texture has an odd repeating notch dead center twice...

I think your intent is fine, but you didn't accomplish what you wanted.

**Comment by sofar** [comment]
FYI it's entirely possible that you have a hard time finding aspen, just like it is possible you have a seed where you have no deserts, or no snow biomes.

In general though, look for apple trees and scan around those, and you should not have much trouble.

**Comment by paramat** [comment]
I prefer sofar's texture, clean, simple and the regularity is suitable for a blocky world.
Bumpmapping makes textures look really bad.

**Comment by paramat** [comment]
```	-- Apple tree and log

		deco_type = "schematic",
		place_on = {"default:dirt_with_grass"},
		sidelen = 16,
		noise_params = {
			offset = 0.04, -- was 0.04
			scale = 0.02, -- was 0.02

	-- Aspen tree and log

		deco_type = "schematic",
		place_on = {"default:dirt_with_grass"},
		sidelen = 16,
		noise_params = {
			offset = 0.01,
			scale = -0.02,
```
^ My suggestion.

1/3 area is pure appletrees, sometimes dense and dark.
2/3 area is a varying mix, areas of pure aspen are extremely rare.
The aspen-rich areas are not very dense, for a more open feel.

I found an area that had the 2 extremes next to each other:
Dark dense appletree forest.
Less dense area of mostly aspen with a few appletrees scattered.

screenshot_20160111_022653

^ Overview of 2 extremes.

screenshot_20160111_022948

screenshot_20160111_022851

paramat:
Please could the aspen trunk be a node or 2 shorter? Currently it extends to within 1 node of the tree top which looks a little odd, the top half of the leaves volume would be thinner branches which are represented by leaf nodes.

sofar:
The seed change is only a very minor change from my initial post, so I have no issues with that and will incorporate the values you suggest in the next revision.

sofar:
Lowered trunk, changed seed values. Commit message was updated to reflect all the changes we've done to this patch set.

HybridDog:
>Bumpmapping makes textures look really bad.

l guess even those pixel bumpmapped aspen tree textures would look good if minetest had a good lighting system.

sofar:
Normalmaps need to be made tailored to each texture to make it look good. Since minetest_game does not ship normal textures anymore, it's accepted that bumpmapping makes stuff look bad, and not relevant for this thread.

Please note I'm deliberately posting screenshots with bumpmapping disabled so you can actually see the textures I made, and not something that is distorted.

C1ffisme:
Not a bad idea. :+1:

paramat:
Okay very nearly ready, one more thing, the noise offset and scale for aspen logs needs to be an exact division of the trees' offset and scale, to ensure logs have the same distribution as the trees.
Since for appletree you have divided by 20  how about the same for aspen?:
```
+			offset = 0.0005,
+			scale = -0.001,
```
1 log per 20 trees (roughly a grid of 4 x 4) is plenty, i did notice aspen logs were a little too common.

sofar:
@paramat - yeah, reasonable and fine by me. Pushed exactly those seed changes.

paramat:
:+1:

paramat:
@sfan5 @Ekdohibs @PilzAdam @ShadowNinja

Ekdohibs:
:+1:

sfan5:
:+1: more trees are nice I guess..

paramat:
6267f26027350f0b80d091044ddfc1177aa6fa47

0-afflatus:
Very nice @sofar

PR #777, opened 2016-01-10
New design of the creative inventory (with tabs)

kilbith:
Here's a  cute new design geared toward simplicity and space efficiency.

Notes : 
- The inventory size per page remains the same (24 slots).
- Tabs added to ease the sorting.
- Less concatenations on the formspec string, it should slightly increase the performances.
- Some clean-up in the fields handler to improve its efficacy.

screenshot_20160113_114517

More screenshots : http://imgur.com/a/KwTbY


mods/creative/init.lua
```diff
@@ -132,39 +136,42 @@ minetest.register_on_player_receive_fields(function(player, formname, fields)
...
 
-	if fields.creative_prev or start_i >= inv_size then
-		start_i = start_i - 4*6
-	elseif fields.creative_next or start_i < 0 then
-		start_i = start_i + 4*6
-	end
-
-	if fields.search or fields.clear then
+	if fields.tabs then
+		for _, idx in pairs(fields) do
```
kaeza: Why not just use `tonumber(fields.tabs)`?

kaeza:
:+1:

mods/creative/init.lua
kilbith: Right, thanks.

rubenwardy:
:+1:

I use the the craft grid in creative mode to test that my recipes work, but giveme would work just as well.

mods/creative/init.lua
```diff
@@ -41,23 +41,34 @@ creative_inventory.init_creative_inventory = function(player)
...
 end
 
-function creative_inventory.update(player_name, filter)
-	local creative_list = {}
+function creative_inventory.update(player_name, filter, tab_idx)
+	local creative_list, tab_cat = {}, {}
 	local inv = minetest.get_inventory({type = "detached", name = "creative_" .. player_name})
 
-	for name, def in pairs(minetest.registered_items) do
+	local idx_category = {
```
kaeza: This could be moved outside the function.

rubenwardy: The [1], [2] is implicit in:

```lua
idx_category = {
    minetest.registered_items,
    minetest.registered_nodes,
    minetest.registered_tools,
    minetest.registered_craftitems
}
```

mods/creative/init.lua
```diff
@@ -41,23 +41,34 @@ creative_inventory.init_creative_inventory = function(player)
...
+		[1] = minetest.registered_items,
+		[2] = minetest.registered_nodes,
+		[3] = minetest.registered_tools,
+		[4] = minetest.registered_craftitems
+	}
+
+	for k, v in pairs(idx_category) do
```
kaeza: You don't need a for loop. Suggestion:
```lua
-- If index out of range, show default ("All") page.
local tab_cat = idx_category[tab_idx] or idx_category[1]
```

mods/creative/init.lua
kilbith: Done.

mods/creative/init.lua
kilbith: Applied, thanks.

mods/creative/init.lua
```diff
@@ -80,41 +91,34 @@ local trash = minetest.create_detached_inventory("creative_trash", {
 })
 trash:set_size("main", 1)
 
-creative_inventory.set_creative_formspec = function(player, start_i, pagenum)
+creative_inventory.set_creative_formspec = function(player, start_i, pagenum, tab_idx)
```
kaeza: `tab_idx` should default to 1 ("All") if not specified. Add this after this line:

```lua
tab_idx = tab_idx or 1
```

kilbith: Applied.

NathanSalapat: I like the design.

tenplus1:
+1

MoNTE48:
+1

paramat:
Looks good, but the crafting grid is needed to craft an item that is 'not in creative inventory'.

kilbith:
@paramat I can understand, but for the rare times you need to craft such item, is it really worth having such a cumbersome element ?

NathanSalapat:
Could the craft grid be added to its own tab?

C1ffisme:
Looks a lot better than the old one, :+1: .

Fixer-007:
Please add crafting grid here just for sure :}

kilbith:
**Update :** crafting tab added.

screenshot_20160112_005337

rubenwardy:
I'd suggest making "All" default.
Either by making it left most, or setting the default current index.

kilbith:
> I'd suggest making "All" default.

This is already the default tab.

paramat:
Nice.

Dragonop:
:+1:

RHRhino:
:+1:  it looks great @kilbith 
Could the selected tab also have another color?

Jordach:
:+1: Noice.

Ekdohibs:
:+1:

paramat:
Please could you recreate the gap that separates the top row of the inventory from the lower rows?
The gap is not particularly important to me but i think it should be consistent with the standard inventory. Other than that :+1:

kilbith:
@paramat Done. Preview updated.

paramat:
Thanks, will merge soon.

paramat:
6d79d157c30ceb5a85fc20c703a72794f7c254e9

0-afflatus:
Excellent!

PR #783, opened 2016-01-12
New trapdoor recipe.

sofar:
Related: #779, @535e611e

Trapdoors are square, not rectangular. The rectangular recipes are
already used for glass panes, walls, doors, but no square recipes
existed until #779 got fixed.

This course of events suggests that it was a mistake to pick the
original trapdoor recipe. Instead, we propose to use a similar
2x2 recipe for the trapdoor, which is square. The recipe can be
made in any of the 4 corners of the crafting grid.

### DIFF
--- a/mods/doors/init.lua
+++ b/mods/doors/init.lua
@@ -530,11 +530,10 @@ doors.register_trapdoor("doors:trapdoor_steel", {
 })
 
 minetest.register_craft({
-	output = 'doors:trapdoor 2',
+	output = 'doors:trapdoor',
 	recipe = {
-		{'group:wood', 'group:wood', 'group:wood'},
-		{'group:wood', 'group:wood', 'group:wood'},
-		{'', '', ''},
+		{'group:wood', 'group:wood'},
+		{'group:wood', 'group:wood'},
 	}
 })
 


Jordach:
Considering that a trapdoor looks a bit like a slab, why not use slabs to
create the trapdoor? The only difference between a slab and trapdoor is
that a slab doesn't have hinges. Just a small suggestion, considering slabs
have no other uses other than decoration.

kilbith:
> why not use slabs to create the trapdoor?

Agree with that. Besides, the 2x2 of `group:wood` is already massively used in mods to craft e.g. wood tile or workbench.

sofar:
Yeah, this is exactly the type of feedback I'm looking for.

you're suggesting something like:

```
slab slab
slab slab
```

?

red-001:
@sofar I thought he meant one slab to one trapdoor, but if four slabs gave four trapdoors that would be good too.

sofar:
Right now it takes 6 planks to make 2 trapdoors. If you use slabs, it takes 3 planks to make 6 slabs.

Current:
    6 planks -> 2 trapdoors
Proposed 1 slab = 1 trapdoor
    6 planks -> 12 trapdoors (6x)
Proposed 4 slabs = 1 trapdoor
    6 planks -> 3 trapdoors (1.5x)

a 6x increase in the amount of trapdoors you get is way too many.

Jordach:
(I mean I wouldn't mind using an steel ingot to make a door hinge as part
of the recipe 1 slab + 1 steel ingot -> one trapdoor...same could be
applied for doors, bar the steel door...)

4 slabs isn't bad, but. Problem is - where does that excess wood go from
making those slabs?

Same problem for stairs. They're missing some of the original material.
(I'm kinda thinking Sam has a hidden bandsaw and a set of tools in his
arms.)

C1ffisme:
> Problem is - where does that excess wood go from making those slabs?

This problem happens with sticks too.

And stairs.

And chests.

And lots of other things.

Though in this case, I would favor gameplay over realism. I don't want to have my chests filled up with extra sticks, chests, stairs and trapdoors.

sofar:
Excess materials are hardly a problem in minetest. Stacks go to 99, and most materials are so abundant that throwing out remainder materials is a problem of minute proportions.

We could make the trapdoor recipe take 6 slabs, in which case 3 wood -> 6 slabs -> 1 trapdoor? that would equal the current cost of the wooden trapdoor.

Jordach:
Auke, sounds good. Especially with the 6 slabs -> 1 trapdoor.

sofar:
Given the fact that there is no slab group, it would be really hard to make a trapdoor out of aspen wood. I'm not in favor of making slab groups at the moment, so I'm dropping this patch for now.

Jordach:
@Auke Kok - start a new PR for having custom slab groups, either with a
function to allow in-place group table overrides, or directly within the
registration of stair and slabs.

sofar:
@Jordach Yes, I contemplated that, and that should be a separate PR indeed.

Auke:
@Jordach Ehm. Please don't use my user name in this conversation. I do get notifications via e-mail, because you did so. Thanks.

Jordach:
I can't tell the actual usernames from email replies, my bad.

PR #785, opened 2016-01-15
Move character.png to textures.

sofar:
This file is obviously out of place, as the engine does not
load textures, ever, out of a models folder.


Jordach:
Probably an issue since Taoki's implementation of models.

I'm reasonably sure that it loads from both - testing might be needed.

everamzah:
screenshot_20160115_100217
This is what appeared when I removed models/character.png outright.

sofar:
@everamzah but did you move it to ```../textures/``` ?

everamzah:
No, I deleted the file outright as a test.  Normally, I move it to textures too, because of the console spam "Fail to load character.png" or whatever.

sofar:
```
3727  16:31:22.381774 open("/home/sofar/git/minetest/character.png", O_RDONLY) = -1 ENOENT (No such file or directory)
```

Irrlicht is somehow attempting to read from ```PWD```. Irregardless, nothing is reading the texture from the ```models/``` folder, so this commit should be merged.

paramat:
Terminal messages became more verbose recently which is why we are seeing it now.
Not sure moving is correct though, perhaps it is only looking in the texture pack for an optional override skin.

sofar:
@paramat The strace shows it never looks for this in ```models```. So this should really go into ```textures```.

the error comes from irrlicht looking in ```$PWD```, which is unrelated, and I don't think this will fix that.

paramat:
If it never looks in models how does it currently find it?

MoNTE48:
Can a path be prescribed in the model?
In any case, +1. All the texture should be in one place.

C1ffisme:
I imagine, and I may be wrong since I haven't looked too closely at Minetest's code, that Minetest imports all resources (models, textures, sounds, schematics) from folders that are known to have said resources. It does not care if a sound is in the textures file, or if a model is in the schematics file. It simply loads it all in, though somewhere somebody probably put a log to tell people "Hey, I loaded this in!" without saying "from a file it shouldn't be in.".

That's just my hypothesis.

paramat:
:+1: from me and sfan5 http://irc.minetest.ru/minetest-dev/2016-02-21#i_4540322 will merge soon.

paramat:
3fcb7821d78aef7c647a7e2687ff1a6328eb108b

Issue #738, opened 2015-12-08
[Suggestion] Change doors to meshnodes


C1ffisme:
Since we've changed stairs to mesh nodes, changing doors would also be a better idea too, in my mind. Not only would we save the rendering of the 4 vertices around the middle of the door, but they would open and close cleaner, and not have to use ABMs, etc. It would also solve every door duplication bug known. And we would only need one node per door, not two.

The only disadvantage I can see is changing old doors into new doors. But, that is fixable by first adding the new doors, then adding an ABM in the version after that, and after that version remove the old doors. This may give problems to those who might upgrade from 0.3.x or 0.4.13 and below going directly to 0.5.0, but we are, after all, removing some mod support. And doors is a mod.

Jordach:
:+1: With this, you could even rotate the door to hang upside down with the
6D facedir. Would look amazing for space station type builds, or for bat
people... ;)

0-afflatus:
:+1:

C1ffisme:
I'm glad at least one of my suggestions gets a bit of positive comments. Thank you.

tenplus1:
It would also be handy if all doors opened by default (by anyone) and you could use a key tool to LOCK the door for owner only or check for protection and open only for those inside a protected area.  This would save having to re-create doors for each instance.

0-afflatus:
@tenplus1 :+1:
"This message will self-destruct in ten seconds ..."

sofar:
@tenplus I wrote a modification that adds keys that work for lockable doors and chests, so that may either conflict or suit your needs. 

https://forum.minetest.net/viewtopic.php?f=9&t=13732
https://github.com/sofar/minetest_game/tree/keys

sofar:
I'm tempted to say that this would drastically improve the door code. Currently door code is convoluted and difficult due to the double-node handling code.

tenplus1:
I was thinking of a formspec idea where you could use a key tool on a normal door and lock it to owner, lock to protected or even lock it to shared use...

sofar:
@tenplus1 doors need a good API first, and then it would be trivial to write something like that in a mod.

tenplus1:
agreed, lets hope the meshnodes are the first port of call :)

sofar:
https://github.com/sofar/newdoor

Here's an initial stab at a new door based on mesh nodes.

[...]

PR #789, opened 2016-01-16
New mesh door models, and extensive door API

sofar:
This patch replaces the default door nodes with a new mesh model
and nodes.

Two new models were added that are 2 blocks high. One for left-hinge
and one for right-hinge doors. This allows us to make a single texture
fit on both models. The alternative would have been 1 model and 2
unmapped textures, which is more work for mod developers.

Doors work exactly like the old doors, including ownership, breaking
doors, opening and closing.

Under the hood, we can prevent the top part of the door from being
obstructed by placing an invisible node. This prevents liquids from
flowing through doors or people placing sand or other blocks in the
top half. The door code automatically places and removes these as
needed.

Metadata is used to store door state, just like the old version.

A doors API is added, it allows mods to use the API to open/close or
toggle door states without worrying about sounds, permissions and
other details. This is intended for e.g. mesecons. This API allows
mods to manipulate or inspect doors for players or for themselves.

In-game old door nodes are automatically converted using an ABM and
preserve ownership and orientation and state.

TNT blows up all doors and trapdoors except for the steel ones,
who can survive a blast.

Thanks to red-001 red-001@users.noreply.github.com for some
of the conversion code, cleanups, and extra textures.


MoNTE48:
-0.5, this broken compatibility with Minecraft and Minetest texture packs...

tenplus1:
+1

Jordach:
It would take less than two minutes to merge the textures into the UV layout. Not really that difficult with GIMP or Photoshop.

-----Original Message-----

sofar:
@MoNTE48 The UV layout is really easy to understand, and allows for more complex door textures than before, including different door inside/outside textures.

mods/doors/README.txt
```diff
@@ -40,8 +43,20 @@ Steel trapdoor textures by sofar (CC-BY-SA-3.0)
   doors_trapdoor_steel.png
   doors_trapdoor_steel_side.png
 
+Door 3d models by sofar (CC-BY-SA-3.0)
+  door_a.obj
+  door_b.ob
```
kaeza: j?

mods/doors/init.lua
```diff
@@ -1,435 +1,440 @@
...
-	end
-	if not def.selection_box_bottom then
-		def.selection_box_bottom= box
+--]]
+
+-- our API object
+doors = {}
+
+-- private data
+_doors = {}
```
kaeza: If it's private, please make local.

mods/doors/init.lua
```diff
@@ -1,435 +1,440 @@
...
+_doors.registered_doors = {}
+_doors.registered_trapdoors = {}
+
+-- returns an object to a door object or nil
+doors.get = function(pos)
```
kaeza: Minor nitpicking, but can the `foo.bar = function(...)` form just die already? `function foo.bar(...)` does exactly the same and looks prettier <3

mods/doors/init.lua
```diff
@@ -1,435 +1,440 @@
...
+	if state % 2 == 1 then
+		state = state - 1
+	else
+		state = state + 1
 	end
-	if not def.sound_open_door then
-		def.sound_open_door = "doors_door_open"
+
+	local dir = minetest.get_node(pos).param2
+	local transform = {
```
kaeza: This could be moved outside the function so it's initialized only once.

mods/doors/init.lua
```diff
@@ -1,435 +1,440 @@
...
+		minetest.sound_play(def.sound_close, {pos = pos, gain = 0.3, max_hear_distance = 10})
+		minetest.swap_node(pos, {name = string.sub(node.name, 1, string.len(node.name) - 5), param1 = node.param1, param2 = node.param2})
+	else
+		minetest.sound_play(def.sound_open, {pos = pos, gain = 0.3, max_hear_distance = 10})
+		minetest.swap_node(pos, {name = node.name .. "_open", param1 = node.param1, param2 = node.param2})
+	end
+end
 
-function doors.register_trapdoor(name, def)
+doors.register_trapdoor = function(name, def)
```
kaeza: Why?

mods/doors/init.lua
sofar: Good one, yes, this one is supposed to be ```local```

mods/doors/init.lua
sofar: sorry, I'm happy with either form, but unless there's a style guide, there will just be someone else nitpicking the other way.

I'm not opposed to your change, this file originally mixed both forms, hence me changing all the forms to the same style, which is an improvement.

mods/doors/init.lua
sofar: sure, will do

sofar:
@kaeza pushed an update addressing all your comments, including the ```function foo()``` change.

mods/doors/README.txt
sofar: ah, lol, didn't see that.

mods/doors/init.lua
kaeza: I guess it's like the eternal war about where to put the braces in C-like languages.

This is just personal opinion, but I (generally) use `function foo.bar()` when I mean to use `foo` as a namespace, while using `foo.bar = function()` if `foo` is an object/record/whatever. Sort of implicit documentation or something.

Did not mean that comment as "do it or else..."; as I said, it was "minor nitpicking". Either way, thank you for making it consistent.

est31:
See [discussion in #738](https://github.com/minetest/minetest_game/issues/738#issuecomment-172275279): I like the idea of the api and so on, but I don't like the new meshes, they have two door faces, the old only had one.

sofar:
@est31 I just looked at ```backface_culling```, and it appears that ```mesh``` nodes don't support this, otherwise I would enable it.

sofar:
@est31 I've updated this patch to enable backface culling as is added in this PR:

https://github.com/minetest/minetest/pull/3578

With that patch, the visual appearance is as you expect, and you do not see the double door.

sofar:
![culling]
(http://i.imgur.com/qqMEVZa.png)

Notice the glass door is (intentionally) NOT culled, but the wooden door is.

est31:
Really nice :)

sofar:
Updated this branch with the necessary settings to enable backface_culling for the 3 doors that need it. This is dependent on minetest/minetest@3578 fix, which I think is now nailed down and ready for merging.

sofar:
Rebased this as there was a small merge conflict due to 2 trapdoor changes being merged.

With the minetest backface_culling change for mesh and other nodes as of 2 days ago, this now can be cleanly applied and works as the screenshot shows.

red-001:
Texture names need to be change to avoid unnecessary conflicts with other mods and texture packs.

sofar:
@red-001 ack, will do.

sofar:
Renamed all the textures to avoid texture pack conflict.

paramat:
I need to test this but seems a good PR. One thing, can we have 'glass-edge textures' halfway down on both sides of the glass door like with the obsidian glass door? I'm thinking the bar is needed on both sides as the imagined handle would be located here. It also makes it clear the glass doors are made from 2 sections of glass, curently the glass door is in 2 parts on one side and 1 part on the other.

paramat:
Testing, nice. The obsidian door has backface culling enabled, perhaps it should be the same as the glass door? Is it possible to make the 'dug' sounds of the glass doors the breaking glass sound?

sofar:
@paramat the obsidian door texture could get some refining, and the idea of a double handle bar visual is probably a good idea. Maybe @red-001 wants to look at it?

paramat:
I recommend adding the standard glass-edge texture halfway down the glass door where it's missing, then for both glass doors add a metallic (for a modern look) horizontal handle on one side, just as the wooden doors have. Maybe i'll do the texture myself ..

paramat:
Here's an attempt, wide flat handles for a modern look:
doors_door_glass2
doors_door_obsidian_glass2

sofar:
I like the ideas sofar, I'll take your textures and change the culling setting, and refresh this PR later today.

sofar:
@paramat, taking your textures verbatim. Made obsidian door no longer backface_culled.

red-001:
@paramat I did that but it look strange when you look at it from an angle but just be my option.

red-001:
@sofar `door:hidden` should have an empty `on_blast` callback so TNT can't destroy it.

sofar:
@red-001 good catch. I see that I had anticipated that, and written all the code to remove it in case the bottom door does get blasted and needs to be removed (a non-protected door), so this is a very simple change.

sofar:
@red-001 what do you mean "inside". I'm not going to change the mesh model at this point. The whole idea is to keep it basic.

paramat:
> I did that but it look strange when you look at it from an angle but just be my option.

Might be nice to remove the glass-edge horizontal from both sides so the doors are single sheets of glass, with the handle placed directly into the glass. Also thinking of adding hinges, will provide some alternative textures.

paramat:
doors_door_glass3
doors_door_obsidian_glass3
These might look better, removed central divider, added hinges, moved handle closer to door edge because it is so wide.

red-001:
@paramat yes that should look better.

sofar:
Yeah, these are good. It's going to be fun watching people run into the obsidian doors :)

Let's not try to overdo this too much, though. We can always revisit these textures at a later point in time.

sofar:
pushed the 2 new door textures from paramat.

sofar:
FYI - here is the corresponding code changes needed for mesecons_doors:

https://github.com/Jeija/minetest-mod-mesecons/pull/249

@Jeija

sofar:
I've pushed a minor update that updates the ```on_blast()``` methods to return an itemstack where this is needed. This will help a change in #806 to make TNT'd doors actually drop the doors as items. Otherwise doors are lost in the blast.

paramat:
To avoid an ABM with interval5 for door conversion how about converting old doors to new on opening? if this is possible and fairly lightweight.

sofar:
@paramat Not as simple as it seems, as I have to catch things like on_punch, screwdriver use, etc. It also likely breaks mesecons. 

The ABM is safer and more efficient: Once all the doors are converted it's never going to fire anymore.

paramat:
Okay no problem.

sofar:
@paramat with minetest/minetest@47464c9344a1a4817ad6e4c6ec44526df305b1f6 merged this can now be merged as well.

@ShadowNinja @sfan5, who else can +1 this so it can be merged?

paramat:
Added line comments.

Ekdohibs:
:+1:

mods/doors/init.lua
```diff
@@ -1,435 +1,444 @@
...
+			end,
+			state = function(self)
+				local state = minetest.get_meta(self.pos):get_int("state")
+				if state % 2 == 1 then
+					-- open
+					return true
+				else
+					-- close
+					return false
+				end
```
kilbith:
``` Lua
return state % 2 == 1
```

mods/doors/init.lua
```diff
@@ -1,435 +1,444 @@
...
+			toggle = function(self, player)
+				return _doors.trapdoor_toggle(self.pos, player)
+			end,
+			state = function(self)
+				local name = minetest.get_node(pos).name
+				if string.sub(name, -5) == "_open" then
+					return true
+				else
+					return false
+				end
```
kilbith:
``` Lua
return name:sub(-5) == "_open"
```

mods/doors/init.lua
```diff
@@ -1,435 +1,444 @@
...
+	},
+}
+
+function _doors.door_toggle(pos, clicker)
+	local meta = minetest.get_meta(pos)
+	local state = meta:get_int("state")
+	local def = minetest.registered_nodes[minetest.get_node(pos).name]
+	local name = def.door.basename
+
+	if clicker ~= nil then
```
kilbith:
``` Lua
if clicker then
```

mods/doors/init.lua
```diff
@@ -1,435 +1,444 @@
...
-				local pn = placer:get_player_name()
-				local meta = minetest.get_meta(pt)
-				meta:set_string("doors_owner", pn)
-				meta:set_string("infotext", "Owned by "..pn)
-				meta = minetest.get_meta(pt2)
-				meta:set_string("doors_owner", pn)
-				meta:set_string("infotext", "Owned by "..pn)
+			if def.protected then
+				meta:set_string("doors_owner", placer:get_player_name())
+				meta:set_string("infotext", "Owned by " .. placer:get_player_name())
```
kilbith: Please localize the placer name.

mods/doors/init.lua
```diff
@@ -1,435 +1,444 @@
...
+
+function doors.register(name, def)
+	-- replace old doors of this type automatically
+	minetest.register_abm({
+		nodenames = {"doors:"..name.."_b_1", "doors:"..name.."_b_2"},
+		interval = 5.0,
+		chance = 1,
+		catch_up = true,
+		action = function(pos, node, active_object_count, active_object_count_wider)
+			local l = tonumber(string.sub(node.name, string.len(node.name)))
```
kilbith:
``` Lua
local l = tonumber(node.name:match("(%d)$"))
```

sofar: :match() is more expensive, it has to iterate over the string

mods/doors/init.lua
sofar: ack

mods/doors/init.lua
sofar: ack

mods/doors/init.lua
sofar: sure

mods/doors/init.lua
sofar: sure, yeah

sofar:
I need to test the updates before I push, which will be later today.

mods/doors/init.lua
kilbith: Then use : 

``` Lua
local l = tonumber(node.name:sub(-1))
```

kaeza: You can use -1 instead of `string.len`. Also, `node.name:sub(...)`, please. If it's really performance critical, localize.

sofar: thanks

sofar:
All tested well after the latest changes.

paramat:
Seems well reviewed now so :+1: will merge soon.

paramat:
f600a9f645af40d22c8eb7c17aff89507b71816e

proller:
this pull brokes all mods using doors api.
protector (fixed)
landrush
and other

sofar:
@proller I'll investigate, I had looked at homedecor and xdecor, but not those yet.

PR #793, opened 2016-01-19
Improve Chest appearance - opening chests.

sofar:
Adds a mesh node that shows when a chest is opened. The chest
stays visibly open until closed. While opening and closing, a sound plays
(a new sound effect is added for both open and close).

A second person inspecting the chest will trigger a second sound
open. However, only after the last player closes the chest, does the
chest actually visually close and is the sound close played.

While it's possible that a server shutting down may cause chests
to appear open, this does not affect the chests' working matter,
and opening or closing them should reset the node state.

Old chests are converted to the new style by opening. The first open
action performs the conversion, after which the conversion will not
repeat. The conversion removes the node formspec to assure that this
is the case (well-tested).

If the lid of the chest is obstructed, the sounds continue to play,
but the lid isn't opened. Obstructed means that a node is present
above the chest lid, however, we ignore several node types like signs,
torches and wallmounted nodeboxes (typically signs) since they don't
pose any major obstruction in almost any case, and are typically
found above chests. Additionally, the selection box of the opened
chest does not include the lid, and so one can still interact with
e.g. a sign behind an open lid.

screenshot


sofar:
Oh, you can also texture stuff on the inside of chests, to suggest something is in there. Open and Close textures are identical, so only 1 texture is needed per chest.

MoNTE48:
It's very cool and very beautiful!

mods/default/nodes.lua
```diff
@@ -1422,115 +1410,200 @@ local function has_locked_chest_privilege(meta, player)
...
+	def.visual = "mesh"
+	def.paramtype = "light"
+	def.paramtype2 = "facedir"
+	def.legacy_facedir_simple = true
+	def.is_ground_content = false
+
+	if def.protected then
+		def.on_construct = function(pos)
+			local meta = minetest.get_meta(pos)
+			meta:set_string("infotext", "Locked Chest")
```
kilbith, line 1838: You're already setting this infotext inside `after_place_node`.

Jordach: Looks good. But. The inside of the chest *isn't* black.

https://www.youtube.com/watch?v=JHfmoG4MUIQ

I've already had user controlled animated nodes. Tsk. Tsk.

Secondly, this could be done with a single mesh.

Not to mention /clearobjects does like to delete things. You don't seem to
replace it once deleted. (Unlidded chests look wrong entirely.)

https://github.com/Jordach/big_freaking_dig/blob/master/mods/deco/init.lua#L133
, (You need one line of node meta for lid state, can be stored between
server restarts, etc)

https://github.com/Jordach/big_freaking_dig/blob/master/mods/deco/init.lua#L940
, (complete with yaw -> degree rotations in place, see mesh .blend - I also
don't see node rotation),

https://github.com/Jordach/big_freaking_dig/blob/master/mods/deco/init.lua#L1082
, (Helper to set the animation frames correctly.).


You're doing it wrong, sort of. Never sort of re-invent the wheel with
existing worlds with chests, and rotations.

paramat:
Inside of chest black by defaults seems best to me, it represents the magic space and void within that can contain 1500m^3 of stone etc.

mods/default/nodes.lua
sofar: @kilbith yes, on purpose - check the on_construct and on_after_place code path of the locked chest. I believe it can go, but the intent was to prevent people from opening the chest as it was placed and gained access.

sofar: @Jordach BFG's chest is an entity with animation. This implementation isn't animated on purpose. Hence /clearobjects is irrelevant for this implementation, since the chest always is a mesh node, open or not.

I purposely chose to make a very restrained and simple implementation that uses nodes and not entities, as I believe that adding an object in the game for the purpose of an animation is a bit over the top for minetest_game. The code of this change is pretty much the same as what e.g. the basic doors do, which is fitting, relatively simple and doesn't spawn entities everywhere.

Because we want to use a single bitmap for both open and closed chests, I decided to make a closed and an open mesh and UV unwrap them in the same way, such that chests can use a single texture. This is preferable since people making textures will be able to replace the skin more easily - there is only one and it's easy to understand. This is why there's a second mesh.

Black was deliberately chosen - I personally think it's a good trade off and fitting. Texture pack makers can draw anything they want on the two exposed black surfaced.

Give it a try, though, don't just look at the code and assume it just breaks everywhere.

C1ffisme:
:+1:

mods/default/nodes.lua
```diff
@@ -1387,32 +1387,20 @@ minetest.register_node("default:torch", {
...
-local function get_locked_chest_formspec(pos)
+local function get_chest_formspec(pos)
 	local spos = pos.x .. "," .. pos.y .. "," .. pos.z
 	local formspec =
 		"size[8,9]" ..
 		default.gui_bg ..
 		default.gui_bg_img ..
 		default.gui_slots ..
-		"list[nodemeta:" .. spos .. ";main;0,0.3;8,4;]" ..
+		"list[nodemeta:" .. spos .. ";default:chest;0,0.3;8,4;]" ..
```
kaeza: Why the name change?

mods/default/nodes.lua
```diff
@@ -1422,115 +1410,200 @@ local function has_locked_chest_privilege(meta, player)
...
-	is_ground_content = false,
-	sounds = default.node_sound_wood_defaults(),
+local function chest_lid_obstructed(pos)
+	local above = { x = pos.x, y = pos.y + 1, z = pos.z }
+	local def = minetest.registered_nodes[minetest.get_node(above).name]
+	-- allow ladders, signs, wallmounted things and torches to not obstruct
+	if def.drawtype == "airlike" or
+			def.drawtype == "signlike" or
+			def.drawtype == "torchlike" or
+			(def.drawtype == "nodebox" and def.paramtype2 == "wallmounted") then
```
kaeza: Also, `def.walkable`?

kaeza:
While I like the feature, using `show_formspec` is subject to perceptible lag (compared to client-side prediction of `formspec` meta field). Many players don't like locked chests for the same reason.

I give half a :+1:.

PilzAdam:
I don't like that this breaks texture packs. Can you create the single texture via `[combine`?

sofar:
@PilzAdam I gave that serious thought.

However, using ```[combine``` to use the same textures to UV wrap the open model and the closed model at the same time is far worse than a single-texture for both chest states. There may be only 6 faces in the closed model, but there are 12 in the open model, and only 4 of the open model faces are 16x16px... Even if it is possible it is bad form to do so.

While this change does break texture packs, this is the easiest thing to fix for them - all they need to do is load the template frim minetest-game and copy-paste the 3 texture pack textures - about 2 minutes work at most.

Of course, I have to make sure that I'm not using any of the previous texture names, so that at least it shows the correct texture from minetest_game, and not a 16x16px tile from a texture pack (checking now...)

sofar:
@kaeza with a nodedef formspec we can't easily and reliably detect player's closing chests, and so sounds are impossible. There is no other way to do sounds currently.

sofar:
@PilzAdam removed conflicting texture names. If people use a texture pack, now there's a way that texture pack makers can anticipate and add skins for the new chest without any breakage.

mods/default/nodes.lua
sofar: Interesting idea, I'll have to play with the concept, but what nodes would you like to see not obstructing the lids in particular?

kilbith:
Alternatively, for not breaking the TPs, you can split each external face in its own mesh object and apply an UV-map per face (the stairs model have that).

paramat:
Would it be possible to delay slightly the closing of the chest so that the player briefly (<= quarter-second) gets to see it open, and sees the change from open to closed? I'm concerned about how the open chest is mostly hidden behind the inventory.

sofar:
@kilbith yes but stairs has the same texture on each model's face, not different textures. I still think this choice is a simpler solution that is easier to understand for texture makers and people reusing the code flow.

sofar:
@paramat I considered that, not without on_globalstep() hook, which I'm very resistant to do. ```minetest.after()``` is too unreliable to do something "shortly after" since it can easily take 5 seconds even if you tell it "do something in 1 second"....

my lightning mod does something like that, it's not pretty and not something I would prefer to put in minetest_game.

Jordach:
Ok, wasn't expecting nodes, but that's fine.

kilbith:
@sofar Wrong, the stairs model can be used with different textures on faces (https://github.com/minetest/minetest_game/pull/576). If you split your mesh objects in the good order, you couldn't break the TPs at all without even changing the `tiles = {}` part.

paramat:
Okay.

sofar:
@kilbith I hear you, I can understand that people are hesitant to change things.

There are however issues with the old method that I think are problematic. First, there's no method to have different sides, bottoms or back. People may want to texture a hinge on the rear, or asymmetric handles on the sides, or otherwise. That's not possible with the current
textures, as top-bottom and left-right-back are shared textures.

Second, I'd like to start moving minetest_game forward, and stop pretending like minetest_game needs to be backwards compatible like this. The attachment to texture packs seems an arbitrary restriction. Perhaps I can offer to submit PR's to texture pack maintainers to convert and add the new chest textures for them?

paramat:
Now tested, lovely feature and good sounds. When close to the chest or at an angle it is visible behind the formspec, which works well visually.
Just once i had some error messages appear of ths form:
```
2016-01-20 19:27:08: WARNING[Main]: GUIFormSpecMenu::drawList(): The inventory location "nodemeta:-143,2,126" doesn't exist
```

kilbith:
@sofar We have a strong culture of backward compatibility for not alienating our users and contributors (insofar as possible). And I repeat : breaking the TPs isn't a necessity here.
- http://wiki.tesnexus.com/index.php/Splitting_meshes_in_Blender
- http://wiki.blender.org/index.php/Doc:2.4/Manual/Modeling/Meshes/Editing/Vertices#Separate

Just split your mesh objects and keep 6 tiles in the nodedefs.

sofar:
@kilbith it's not breaking the TP. It's adding a new texture.

red-001:
@kilbith I don't think there is even one texture pack maker out there who wouldn't be able to merge a few textures in gimp or photoshop in less then five minutes. And if @sofar is right and there is no way to have separate front and back textures without breaking compatibility (till someone merges a few textures) I think most players and texture pack makers would be ok with that.

paramat:
I prefer the simpler (and probably lighter and faster) texture implementation that is in this PR currently, even if it forces texture packs to be updated. TPs are always being updated anyway due to new nodes so we shouldn't be scared of breaking them. So i would like this PR to be judged without consideration of TP breakage.

sofar:
@paramat if you're interested to see what is needed to make the open/close change visible to the user, then this change would accomplish that effect:

https://gist.github.com/sofar/d5c2c5e36ff6bd9bc33e

This works fairly reliable on a localhost type server. I haven't attempted to try this with more network delay, so no idea if it's terrible or not.

```minetest.after``` is pretty bad, and the resolution is unusable for anything serious. I'm wondering if we should just add ```minetest.after_ticks``` as Implemented as it may be useful for other similar things (as said I use it for lightning sky box effect).

**Comment by sofar** [comment]
@paramat after minetestt/minetest#3593 gets merged or fixed, I think we can do the animation thing as well. With the improved timer, this should be more than reasonable.

**Comment by paramat** [comment]
I find that when closing the chest, sometimes the closing happens 'just' after the formspec disappears, a tiny amount of extra delay would be good. I'm not sure about delaying the formspec appearing so the opening is visible, depends how reliable 'after ticks' is, it would be a very short delay because players will want to access the chest inventory fast. By short delay i'm thinking 1/8th second.

**Comment by sofar** [comment]
@paramat with the fix to ```minetest.after()```, we won't need the ```after_ticks()``` and can just use the former. I have done a test and it functions really well. I used a delay of 0.2 and 0.1, and both work, so that's about the ball game number needed. Additionally, I delay the display of the formspec initially, so you also see the chest open before the formspec appears.

I fully intend to add this, it just looks too good. It even makes the lid close "snap" sound play at exactly the right time.

**Comment by paramat** [comment]
I see i got 'ticks' confused with improvements to 'after'. I will probably +1 this but would certainly prefer these tiny delays to be added. Even just the sounds and occasionally seeing the open lid in the background really improves the game.

**Comment by sofar** [comment]
@paramat I will, for sure, merge that code.

But if I merge that code in this patch, then it will cause chests sometimes not to open for 5+ seconds, since that is how bad ```minetest.after()``` is right now. So I really don't want to break things in this patch.

**Comment by paramat** [comment]
Yes i understand.

**Comment by sofar** [comment]
Updated this patch:

- with the fixes in game to ```minetest.after```, we can now use a very small timer to delay formspec visibility (0.2sec) reliably. This will make chest opening and closing clearly visible and unobstructed for the player.
- minor indentation fixes on really long lines.
- renamed a local variable.

**Comment by sofar** [comment]
Make sure to try this PR in combination with the latest ```HEAD``` of ```minetest```

**Comment by paramat** [comment]
Nice will test soon.

**Comment by paramat** [comment]
Timing is fine, especially good when it closes.
One thing to consider, for simplicity how about the chest opening even if there is a node above? it's weird but so is a chest visually not opening but still making sounds.

**Comment by sofar** [comment]
Opening the lid if there's a full size node above it wouldn't be too weird, but it becomes awkward when there's a slab or another semi solid node there.

I opted for playing most all of the sounds to give players in the area some feedback as to what's going on around them. If I would omit sounds, it would be too trivial for someone to peek into other people's chests since nobody can see or hear that. Making the opening sound always play will alert players of that. Plus we shouldn't shy away from adding more functional audio to nodes. Furnaces are completely silent. Beds don't make any sounds, even tools that break don't have sounds :)

**Comment by 0-afflatus** [comment]
I love this idea. Long time coming. Yes it will be a PITA having to redo my chest textures, but I think it will be worth it.

**Comment by paramat** [comment]
That's fine, so i support this but haven't reviewed the code in detail.

**Comment by sofar** [comment]
@0-afflatus let me know where your textures are and I'll make them for you.

**Comment by 0-afflatus** [comment]
Nice one @sofar https://github.com/0-afflatus/grailtest/tree/master/mods/equipment/textures

Probably needless to say, but I've been a bit useless at backporting my ideas due to time constraints, but if anyone else wants to, go ahead. Most of my textures are 32x32px so probably not much use to MTG, but should all be appropriately licensed.

**Comment by sofar** [comment]
@0-afflatus 

![]
(http://i.imgur.com/i7GXkUn.png)

![]
(http://i.imgur.com/oVRK73H.png)

**Comment by 0-afflatus** [comment]
Thank you. I've just realised this means I can take the hinges off the sides now.

**Comment by sofar** [comment]
Yes, I noticed that too, they don't look so bad per se, but you may want to tweak it further.

**Comment by ShadowNinja** [comment]
I think the black inside looks like a bug.  If you want a "mysterious deep dark chest" look add a border of wood around the black and make the edges lighter.

**Comment by sofar** [comment]
@ShadowNinja textures are easily fixed. A 1px wood border may help but it may just get people to complain that now they rest of the box on the inside ... I'll leave it open for a few days to see how others react.

**Comment by paramat** [comment]
The black seems fine and suitable to me, anything else might look awkward or suggest an empty chest. We could always discuss and change it later.

**Comment by C1ffisme** [comment]
@paramat I was thinking of making just a 1px border, and the inside of the border is still just black.

**Comment by Ekdohibs** [comment]
It looks good, **but** has the problem that minetest.show_formspec is laggy, and thus the chest suffers the same problems as locked chests on laggy servers. So, half a :+1: too. Also, what happens to already placed chests that have a "formspec" field?

**Comment by sofar** [comment]
@Ekdohibs Old style chests get converted automatically when they're opened. The first time they are opened they won't visually change shape, but on subsequent openings they will.

**Comment by sofar** [comment]
FYI here's how it looks with a 1px brown edge. Note it's the same edge as the rest of the chest's edging.

It looks fairly decent, actually. I might actually prefer it. @paramat?

![]
(http://i.imgur.com/ptM4h84.png)

**Comment by C1ffisme** [comment]
That looks good. Not too glitchy, but not too much like the chest is empty.

:+1:

**Comment by Xanthin** [comment]
The edge fits well and I like the whole chest improvement. :+1: 

I was just curious to test it ingame and while doing I wondered what will happen if a player 1 (P1) opens a chest and player 2 (P2) comes by and digs it away. (not too crazy thought, isn't it? :))

The debug was flooded with this messages:
```
2016-02-12 01:41:30: WARNING[Main]: GUIFormSpecMenu::drawList(): The inventory location "nodemeta:1,6,-37" doesn't exist
2016-02-12 01:41:30: ERROR[Main]: InventoryMenu: The selected inventory location "nodemeta:1,6,-37" doesn't exist
```

When P2 digs the chest that P2 has opend, the chest formspec for P1 gets blank while P2 gets an open chest item.



After P1 exits from the formspec the chest gets placed again, but the open chest is still in P2's inventory. If the chest is opend again, the formspec is still blank.

It's a way to easy duplicate chest then.
(somehow I managed to open a locked chest from P1 with P2 in this procedure, but unfortunaley I wasn't able to reproduce it again)
In general the above didn't work for me with a chest with items inside and locked chests.

**Comment by sofar** [comment]
@Xanthin ahh, I made a typo in the drop list for the open chest - which is supposed to return a closed chest. Fix incoming.

**Comment by sofar** [comment]
@Xanthin I'm fairly certain that the warnings displayed and the formspec going blank is actually expected behavior and unpreventable.

**Comment by sofar** [comment]
It's the client that displays the warning messages, not the server. The warning messages are correct.

**Comment by sofar** [comment]
I'm looking more deeply in the duplication issue.

**Comment by sofar** [comment]
Ah, easy fix. I'll make opened chests undiggable.

**Comment by paramat** [comment]
Yes i like the 1px border, i had the wrong idea of how it would look.
It looks like the texture in the screenshot has very dark brown for the 'space' with a slight noise, i thnk it is better a featureless dark brown, as any noise will suggest a flat surface instead of a void.

**Comment by sofar** [comment]
@paramat hmm, that was not intentional, I'll repaint it black before I repush the textures with border

**Comment by sofar** [comment]
Just a minor adjustment: I've renamed the "chest closed" mesh to "cube.obj". It's such a generic shape that it can be used for other purposes and may be useful, so it should have a generic name.

**Comment by kilbith** [comment]
@Ekdohibs is right that `minetest.show_formspec` is laggier than a simple node metadata setter. You have no way to use the latter instead ? If not, I'm not too keen on this PR.

**Comment by sofar** [comment]
@kilbith There's no callbacks usable to change the node to the open state with a node formspec in time, so this patch completely relies on that.

**Comment by sofar** [comment]
To put it bluntly, with a nodemeta formspec-based chest you can't even play sounds when a chest is opened.

If we had client-side Lua, then this could be done lag-free....

**Comment by paramat** [comment]
Perhaps best leave this until after release due to possible problems that may arise.

**Comment by sofar** [comment]
@paramat that's entirely reasonable.

**Comment by Ferk** [comment]
What happens if a user disconnects while he had the chest formspec open?
Would the open chest remain?

**Comment by sofar** [comment]
@ferk, yes, but it's fixable by someone else opening and closing the chest, and only visual.

**Comment by sofar** [comment]
Rebased due to conflicts.

**Comment by sofar** [comment]
rebased, retested and still looks OK.

**Comment by paramat** [comment]
No longer sure about this, i like the opening animation and the sounds, however i am concerned about the complexity and the way it is focussed on a top-opening lid.

In my moonrealm mod i have just added my own storage nodes by copying the current chest code, i did not have to add mesh files and i am free to make my storage side-opening instead of top-opening. The top-obstruction code in this PR does not apply to my storage nodes. Because mine are side-opening and don't want an animation i would probably the use cubic mesh for open and closed modes, but using such a mesh is unnecessary.

So essentially, this is fancy and looks good, but is more complex and less flexible.
For modders to create their own alternative types of opening storage would involve creation of meshes, which is beyond most people's skills (mine too).
This is related to what celeron55 writes in the Minetest roadmaps:

"Particularly, Minetest is not about making fancy-looking things at the cost of freedom."

"Don't mind the lesser visuals and take full advantage of them in how they make creating unimagined things easier. There are enough games in this world that focus on visuals by taking away from everything else, or by making everything else harder to do."

**Comment by sofar** [comment]
@paramat, those are very weak arguments.

> For modders to create their own alternative types of opening storage would involve creation of meshes, which is beyond most people's skills (mine too).

This patch doesn't change that, at all.

> "Particularly, Minetest is not about making fancy-looking things at the cost of freedom."

That's the philosophy for `minetest`, but a bit out of context for `minetest_game`. We spend hours debating pixel colors, so aesthetics are not off the table at all.

> "Don't mind the lesser visuals and take full advantage of them in how they make creating unimagined things easier. There are enough games in this world that focus on visuals by taking away from everything else, or by making everything else harder to do."

This argument would hold true if the patch would *make things harder to do*, instead it doesn't make anything harder to do (in the gameplay sense, since nothing changes for the player), and makes things easier to do (in the coding sense, since it adds a simple chest API).

> Some things in MTGame are fairly crude and super-simple, but these characteristics are welcome when modding, i would always prefer ease of modding to fancy features.

Nobody mods the default chests. Everyone, overrides them entirely or simply adds new chests.

As for complexity, it would be trivial to modify the code here to allow passing an e.g. different mesh, which would allow e.g. an opening drawer type of storage node, or a kitchen cabinet swinging door, etc. Would that be preferable perhaps?

**Comment by C1ffisme** [comment]
> That's the philosophy for `minetest`, but a bit out of context for `minetest_game`. We spend hours debating pixel colors, so aesthetics are not off the table at all.

I agree with you completely, @sofar. How many grass textures have we had now?

@paramat If you want this behaviour to copy, why not publish a mod with storage objects that don't open?

**Comment by sofar** [comment]
Note: I'm not hugely attached to this code per se. If this doesn't get merged I will attempt to make it a mod that overrides the chests, but I think having a simple storage API may well be worth it (especially considering there are a ton of storage nodes out there).

Perhaps that should come first?

**Comment by paramat** [comment]
> This patch doesn't change that, at all.

If i were to make a side-opening storage, i would need to make a mesh. However i think most modders would just give up on that and just use a non-changing cube.

> That's the philosophy for minetest, but a bit out of context for minetest_game. We spend hours debating pixel colors, so aesthetics are not off the table at all.

It's the philosophy for MTGame too, 'freedom' meaning the ease and flexibility of modding. Debating textures and aesthetics is fine and not 'fancy-looking things at the cost of freedom'.

> makes things easier to do (in the coding sense, since it adds a simple chest API).

Easier to make one particular type of chest, harder to make alternative storage. When i made my moonrealm storage it would have been a pain to modify your code, and it would involve an unnecessary mesh file. The existing chect code is already simple to mod, just copy it and add a new texture, so this chest API is barely easier.

> Everyone, overrides them entirely or simply adds new chests.

Yes as i did, by copy-pasting code, and it was very welcome that the chest code was short and simple and didn't involve meshes.

> it would be trivial to modify the code here to allow passing an e.g. different mesh, which would allow e.g. an opening drawer type of storage node, or a kitchen cabinet swinging door, etc. Would that be preferable perhaps?

No, because that still restricts modders to a set of designs by someone else, MTGame should as much as possible be easy enough to mod freely into anything a modder can imagine.

Some things are already complex, like the player model and boats, and i wouldn't suggest removing those, however we should be very careful about adding new complex things. If these new chests had particularly useful new features maybe there would be more justification, but it's only non-essential visual fancyness and sounds, which i admit i do like a lot.

**Comment by sofar** [comment]
>> Everyone, overrides them entirely or simply adds new chests.
>Yes as i did, by copy-pasting code

Completely still entirely 100% possible, and just as easy as before. And now people have the choice to (1) use the old code for simple chests, AND
(2) use the new code without writing their own meshes.

Just because boilerplate code is changed doesn't mean it becomes unavailable.

**Comment by paramat** [comment]
Where is the old chest code available in the code? Not many modders would search back through Github history or think of doing so.

Concerning mesh doors, before the change these were simple nodeboxes involving simple node manipulations. If i wanted to make 3 or 4-node high doors i could stack these nodeboxes and extrapolate the code to work.
Now the code is much longer and more complex, i can barely understand it as a modder of 3 years and 50 mod releases. Making a new mesh is also beyond most modders' ability, mine too.

Investigating whether a mesh would be good for a door was cool, but the problems kept arising and more complex code was added to overcome these, but there is very little advantage to them. Having the selectionbox cover the whle door instead of half a door is one minor improvement, as is the door moving more 'as one' instead of one half sometimes lagging visually.
But these do not justify the large amount of complex and hard to understand code needed and the problems keep coming. I feel we should have given up earlier and reverted them before release.

Celeron55 resists meshes unless they are absolutely essential, such as player model and boat, the approach of MTGame is to keep things to the existing drawtypes and to nodeboxes if possible. The engine is the place for fancy stuff because modders don't need to understand that code. Subgames need to be fairly easily moddable and understandable, needing perhaps GImp and nothing else, certainly not blender.

All this is exactly what c55 means by his roadmap comments i quoted.
I think these new chests should be a mod. The animation is not important enough to justify the complexity, the extra code and the use of meshes.

**Comment by DonBatman** [comment]
@paramat I understand what you are saying but disagree. Modders are quite resourceful and the community is very friendly. I don't have a clue how to use blender but was easily able to get models I needed. 

Also if a modder wants to do something like make the old style chests then they ask on irc or the forum. Many people are there and willing to help.

I would like to see opening chests. Would be a nice addition. The only question I have about it is what happens to existing chests that are stacked. For example I have a storage room on a server with chests stacked 4 high.

**Comment by benrob0329** [comment]
@DonBatman I completely agree.

**Comment by paramat** [comment]
I'm fairly sure stacked chests will still work, just without the animation.

**Comment by C1ffisme** [comment]
@paramat Besides that, we are on Github, which means anyone can access an older version at any time. (Unless Github somehow deletes it's entire database, in which case, the code still exists in minimal.) Not only that, but it's fairly easy to code, as long as you've learned the documentation well enough.

**Comment by paramat** [comment]
You are a very talented, valuable and appreciated contributor to MTGame, i get a good vibe from you and it's fun to work with you, you have added good things and have worked hyperactively.

Some think that's enough to become a game dev, but it's important to have the right approach to MTGame because you will have influence over what is merged. i wish i could support you being a game dev, we need the help, but i can't yet.

From obsessive reading of chat logs i have seen many messages from you that give the impression you want to make extreme and radical changes to MTGame, you jokingly warn us we might regret making you a dev, a few times you have openly admitted your approach is too extreme for us. No offense at all and it doesn't change your complete awesomeness as a contributor, i just feel you, erm, need to be kept under control.

**Comment by ExeterDad** [comment]
I for one wish we had more Sofars around here.

**Comment by sofar** [comment]
Well, you're absolutely right about one thing: I'm not here to support the status quo, and if I would be bound to maintaining the status quo, I would be long gone before you know it.

I also do not mind having serious discussions about how far I can take changes, if the conversation isn't had, then nobody is clear that things even *can* be changed.

If we're not raising the bar, we are falling behind.

This discussion is far more valuable than the actual patch.

**Comment by Ferk** [comment]
Actually, with this implementation of register_chest, even if a modder was able to make its very own mesh he wouldn't be able to use it.

While I really like this idea, I have to agree with @paramat on the lack of flexibility, a lot of the parameters are hardcoded.

But I think the register_chest function can easily be adapted to be able to create chests with other drawtype / visual than "mesh", or other meshes than "cube.obj" / "chest_open.obj".

It would be a bit more trouble to make the open chest conditional.. but even that would be doable. Something like this:

    -- ... add here common chest functions

    if def.open_chest then
        local def_opened = table.copy(def)
        def_opened.drawtype = def.open_chest.drawtype  or "mesh"
        def_opened.visual = def.open_chest .visual or "mesh"
        def_opened.mesh = def.open_chest .mesh or "chest_open.obj"
        def_opened.tiles= def.open_chest.tiles or def.tiles
        -- ... add here other chest_open specific stuff
        minetest.register_node("default:" .. name .. "_open", def_opened)
    end

    def.drawtype = def.drawtype  or "mesh"
    def.visual = def.visual or "mesh"
    def.mesh = def.mesh or "cube.obj"
    -- ... add here other closed chest specific stuff
    minetest.register_node("default:" .. name, def)

Then you can make the swap_node logic only happen when there exists a def.open_chest

**Comment by sofar** [comment]
I'm contemplating what to do with this PR, just haven't made up my mind on it yet.

**Comment by paramat** [comment]
I'd be happier if the chest states were nodeboxes, if that's possible, i'm not sure a nodebox can extend beyond the cube though, this would mean a less elegant vertical lid.

Your efforts to improve MT are much appreciated and of course you should do what you want and no-one will dictate that. Some of your submissions are too complex for MTGame, but there are always mods and subgames for those.

**Comment by sofar** [comment]
> i'm not sure a nodebox can extend beyond the cube

yes, but you lose the ability to make a nice UV map.

**Comment by paramat** [comment]
have you seen the 2 roadmap posts by Celeron55? These perfectly describe what Minetest is and the stuff that needs to be done. We do want improvement and change in certain areas, perhaps you could help out with VAEs? Other subgames are also much wanted.
http://c55.me/blog/?p=1491
https://forum.minetest.net/viewtopic.php?f=7&t=9177

**Comment by paramat** [comment]
I seem to remember if a nodebox extends larger than the cube the texture is tiled or something inconvenient. Otherwise it would have been good to use larger nodeboxes for the new doors.

**Comment by sofar** [comment]
> I seem to remember if a nodebox extends larger than the cube the texture is tiled or something inconvenient. 

That's what I meant with "lose the ability to make a nice UV map"...

> We do want improvement and change in certain areas, perhaps you could help out with VAEs?

Well, not VAE's, but I am working on a scriptable entity API (taking a break this week to do other things and some semi-vacation stuff next week, but it'll come back on my desk soon enough).

**Comment by benrob0329** [comment]
I think that having a general storage node API that can optionally have an "open" mesh would be the best way to go.

**Comment by paramat** [comment]
Sounds good.
Mejh i've been too grumpy about doors, sorry, i don't regret the change.
For this PR i'm neutral.

**Comment by numberZero** [comment]
:+1:
Really, currently MT lacks many minor features that make the game impressive. Opening chests were one of such things. (and ambient sounds are another one)

And +1 for chest API. But there should be more flexibility: for example, there may be a choice for opening direction (upwards, forwards, none [block change may still happen, but no checks on surrounding nodes]) and possibility to use custom mesh (both for opened and closed chest). Also there may be a configuration option for opening direction of default chests (upwards for nicer look, none for maps with weirdly stacked chests)

**Comment by paramat** [comment]
Sorry about my grumpy comments here, i removed them.

**Comment by yyt16384** [comment]
> If the lid of the chest is obstructed, the sounds continue to play,
but the lid isn't opened.

I think some players will find it confusing and consider it a bug.

Something offtopic:

> Concerning mesh doors, before the change these were simple nodeboxes involving simple node manipulations. If i wanted to make 3 or 4-node high doors i could stack these nodeboxes and extrapolate the code to work.

Is it still possible to do it now with meshes, considering that nodes with selection box that extend more than 1 node don't work?

**Comment by paramat** [comment]
I must admit i'm tempted by this, but am cautious because of the lack of dev time, and because sofar seems to have little time to dev recently. Maybe reconsider after 0.4.15 or if/when sofar has more time.

**Comment by DS-Minetest** [comment]
shouldnt a part of the lock go up, too? it looks like someone cut the upper part away

**Comment by sofar** [comment]
@DS-Minetest that would complicate the model a lot. I prefer keeping it this simple,

**Comment by paramat** [comment]
Yeah it's fine.

**Comment by octacian** [comment]
I personally think that it is more important to work on making actual animations available. That way the transition would be smooth. I'm not sure how hard that is, so maybe my thought is quite far put. Will still be nice though.

**Comment by Ferk** [comment]
I don't think it needs animations, it's consistent for it to open just like the door and trapdoor do.
It also would be too complex or hacky to add them (either switching to an entity for the transition or implement mesh animations for nodes in the engine)

What I think this needs is opinions from more developers to know which direction should this be going.

**Comment by paramat** [comment]
The direction is fine, and sofar has stated he wants to keep it simple with no animations, and i agree, this is complex enough. Perhaps we'll merge this after 0.4.15.

**Comment by octacian** [comment]
@paramat true. I guess I would prefer something to nothing.

**Comment by sofar** [comment]
Rebased.

**Comment by tobyplowy** [comment]
@sofar @paramat Oh man This wold be a Amazing feature (is this Still Being worked on)

**Comment by tobyplowy** [comment]
@sofar @paramat Oh man This wold be a Amazing feature (is this Still Being worked on)

**Comment by sofar** [comment]
Rebased. This is still current and working well.

**Comment by sofar** [comment]
Ergh, ignore that, I must have broken something on rebase... fixing.

**Comment by sofar** [comment]
Need more time to fix locked chest conversion, which may have been broken for a while somehow (probably a rebase killed it a while back, and I didn't spot the problem earlier).

**Comment by paramat** [comment]
I suggest we consider and possibly merge this after release and early in the next cycle, due to issues that may arise as they did with doors.
This needs more input from other devs, i am still neutral.

**Comment by rubenwardy** [comment]
I agree with @paramat 
I like the concept

**Comment by tobyplowy** [comment]
@paramat will this be in the next update?

File: mods/default/nodes.lua [hunk 4]
```diff
@@ -1686,125 +1674,226 @@ local function has_locked_chest_privilege(meta, player)
...
+	local pn = player:get_player_name()
+
+	local pos = open_chests[pn].pos
+	local sound = open_chests[pn].sound
+	local swap = open_chests[pn].swap
+	local node = minetest.get_node(pos)
+
+	open_chests[pn] = nil
+	for k, v in pairs(open_chests) do
+		if v.pos.x == pos.x and v.pos.y == pos.y and v.pos.z == pos.z then
```
SmallJoker: `vector.equals(v.pos, pos)`

paramat:
tobyplowy, please stop asking pointless questions like that, i can't forsee the future :) If you just follow this thread it will be obvious how this is progressing.

tobyplowy:
@paramat sorry but I'm really hyped for this I just want it so bad!

paramat:
I quite like what this does and have no objections, but am still neutral.

sofar:
Sorry, vacation has killed my time available, this is still in my queue.

JurajVajda:
I think the metal lock in front of the chest should stand out more. Right now its too flat. In 3d it would give the chest more real visual. Also the lock should be attached to the top part of the chest when opening. Both from above makes more sence for locked chest but is welcome also for regular chests too.

sofar:
> Suggestion:
> I think the metal lock in front of the chest should stand out more. Right now its too flat. In 3d it would give the chest more real visual. Also the lock should be attached to the top part of the chest when opening. Both from above makes more sence for locked chest but is welcome also for regular chests too.

Suggesting something that is clearly going to be a copy of minecraft isn't productive. I don't want to create more complex models, either.

JurajVajda:
I dont suggest to copy minecraft just suggested to make the model more natural, it's not like minecraft have invented chests so it could have been a more detailed model wihin the minetest styleguide and vision and if it means to have simple model then be it...i am completely fine with that. I just remembered discussions from the past about what is missing in MT and why its not that smoot as MC or whatever therefore the suggestion. These kind of details makes a big difderence many times when good executed and shows difference and creativity rather than opening block with a chest texture over it. But as im saying i am fine with that as it is now and im excited to see it in MT regardless. +1

numberZero:
@sofar what’s up with this?

DS-Minetest:
This would break the chest textures of all texture packs. What about doing something with materials if possible?

tobyplowy:
@DS-Minetest with that kind of reasoning were never going to get anywhere
 
(If a new update Comes Out And you don't make textures for the new blocks and items your Texture Pack is gonna be Broken Anyways (out-of-date) so no point into Keeping compatibility in this situation)

paramat:
This has 1.5 approvals so won't need much more approval to be merged, as stated before i have no objections and it works well, i'm just feeling rather neutral.

DS-Minetest:
> This has 1.5 approvals so won't need much more approval to be merged...

But it still has the `WIP` label.

I hope, this is not a that stupid question: Where does the .5 come from?

paramat:
Ekdohibs gave half a +1 above.

tobyplowy:
No offence and I'm not trying to post useless comments but seriously why is this taking so long?
Can someone Please explain?

paramat:
Usual reason, devs are very busy and have to prioritise, the PR needs some work before it is mergeable.

> Need more time to fix locked chest conversion, which may have been broken for a while somehow 

> this is still in my queue.

paramat:
sfan5 is +1 http://irc.minetest.net/minetest-dev/2017-04-23#i_4893287

sofar:
Alright, I'll try and make the time to fix the issues.

sofar:
Rebased. Several small bugs and of course the major bug were fixed:

- conversion of all chests is now an LBM. This is reliable and doesn't rely on players clicking the chests to convert it. (tested)
- sound volume was fixed for all sounds (wasn't properly positional)
- rebased to use new permission changes

Tested pre-patch conversion and both chests convert OK now with the LBM change. The LBM code also will work for derivative chests in the future, like doors does.

Please give it a test.

paramat:
I have on objections, just feel neutral.

sofar:
I'll merge this considering the approvals (assuming rubenwardy's like is an approval as well). Maybe @smalljoker wants to chime in?

paramat:
@sofar maybe you can double-check conversion before merging, since it's critical?

paramat:
Rubenwardy is +1 for concept, has not tested or reviewed.

mods/default/nodes.lua
```diff
@@ -1766,188 +1766,276 @@ minetest.register_node("default:lava_flowing", {
...
-	default.gui_slots ..
-	"list[current_name;main;0,0.3;8,4;]" ..
-	"list[current_player;main;0,4.85;8,1;]" ..
-	"list[current_player;main;0,6.08;8,3;8]" ..
-	"listring[current_name;main]" ..
-	"listring[current_player;main]" ..
-	default.get_hotbar_bg(0,4.85)
-
-local function get_locked_chest_formspec(pos)
+local function get_chest_formspec(pos)
```
VanessaE: Please make this and related functions/strings global so that I can modify them from a mod (e.g. pipeworks), without having to redefine the node.

numberZero: Ideally, if made global, that particular function should support custom chest size as well. But that seems to be problematic due to the background.

numberZero: But well, is that really in scope of this PR?

sofar: Mind if I do that in a subsequent PR? The locked formspec was never public to begin with. I have no issues against doing so, but I'd prefer to do it additionally so it can get some proper discussion.

VanessaE: That's fine by me.

sofar:
Verified once more that chests, with all content, including metadata of itemstacks, are all converted properly for both locked and normal chests.

paramat:
011ad78b42146725d93761473c7fd06698b2ca73
At last.

PR #796, opened 2016-01-21
Basic stone walls, using NDT_CONNECTED.

sofar:
These basic connected wall nodes automatically connect
to neigboring stone blocks, other wall blocks and anything
that's "cracky". The do not connect to wood (fences will do
that).

The walls are generated using a new walls.register() API.
Documentation on the API is included in game_api.txt.

This change requires minetest/minetest#3503.

### DIFF
--- a/game_api.txt
+++ b/game_api.txt
@@ -139,6 +139,17 @@ Allows creation of new fences with "fencelike" drawtype.
 	groups = {choppy = 2, oddly_breakable_by_hand = 2, flammable = 2},
 	sounds = default.node_sound_wood_defaults(),
 
+#Walls API
+---------
+The walls API allows easy addition of stone auto-connecting wall nodes.
+
+walls.register(name, desc, texture, mat, sounds)
+^ name = "walls:stone_wall". Node name.
+^ desc = "A Stone wall"
+^ texture = "default_stone.png"
+^ mat = "default:stone". Used to auto-generate crafting recipe.
+^ sounds = sounds: see [#Default sounds]
+
 Farming API
 -----------
 
--- a/mods/walls/depends.txt
+++ b/mods/walls/depends.txt
@@ -1,0 +1,1 @@
+default
--- a/mods/walls/init.lua
+++ b/mods/walls/init.lua
@@ -1,0 +1,61 @@
+
+--[[
+
+Walls mod for Minetest
+
+Copyright (C) 2015 Auke Kok <sofar@foo-projects.org>
+
+This program is free software. It comes without any warranty, to
+the extent permitted by applicable law. You can redistribute it
+and/or modify it under the terms of the Do What The Fuck You Want
+To Public License, Version 2, as published by Sam Hocevar. See
+http://sam.zoy.org/wtfpl/COPYING for more details.
+
+--]]
+
+walls = {}
+
+walls.register = function(wall_name, wall_desc, wall_texture, wall_mat, wall_sounds)
+	-- inventory node, and pole-type wall start item
+	minetest.register_node(wall_name, {
+		description = wall_desc,
+		drawtype = "nodebox",
+		node_box = {
+			type = "connected",
+			fixed = {{-1/4, -1/2, -1/4, 1/4, 1/2, 1/4}},
+			-- connect_bottom =
+			connect_front = {{-3/16, -1/2, -1/2,  3/16, 3/8, -1/4}},
+			connect_left = {{-1/2, -1/2, -3/16, -1/4, 3/8,  3/16}},
+			connect_back = {{-3/16, -1/2,  1/4,  3/16, 3/8,  1/2}},
+			connect_right = {{ 1/4, -1/2, -3/16,  1/2, 3/8,  3/16}},
+		},
+		connects_to = { "group:cracky", "group:wall", "group:stone" },
+		paramtype = "light",
+		is_ground_content = false,
+		tiles = { wall_texture, },
+		walkable = true,
+		groups = { cracky = 3, wall = 1, stone = 2 },
+		sounds = wall_sounds,
+	})
+
+	-- crafting recipe
+	minetest.register_craft({
+		output = wall_name .. " 6",
+		recipe = {
+			{ '', '', '' },
+			{ wall_mat, wall_mat, wall_mat},
+			{ wall_mat, wall_mat, wall_mat},
+		}
+	})
+
+end
+
+walls.register("walls:cobble", "Cobblestone Wall", "default_cobble.png",
+		"default:cobble", default.node_sound_stone_defaults())
+
+walls.register("walls:mossycobble", "Mossy Cobblestone Wall", "default_mossycobble.png",
+		"default:mossycobble", default.node_sound_stone_defaults())
+
+walls.register("walls:desertcobble", "Desert Cobblestone Wall", "default_desert_cobble.png",
+		"default:desert_cobble", default.node_sound_stone_defaults())
+


sofar:
This was part of the PoC testing code, but these nodes are ready to go and very much usable. As stated, this depends on the connected nodeboxes PR in minetest to be merged.

Once that patch is merged, I will submit a second commit to convert the existing fences to this new nodebox type (the conversion is seamless). However, There is still a fence API PR that is still pending as well, that should be merged prior.

kilbith:
:-1: for registering a shitload of walls (11). This is not an obese game like Dreambuilder !

Just stick with the cobble walls and mossy cobble walls.

kilbith:
Also I find the middle part ridiculously thin and inconsistent compared to the posts. Can you thickened it by 2 pixels and leveled it up by 1 pixel ? Like that : 

screenshot_20160121_112501

sfan5:
:+1:

sofar:
@kilbith agreed it was a bit over the top with 12 new wall types :)

How about I leave all the cobble walls in (normal, mossy and desert cobble - 3 total) ?

I'm open to adjusting the model - that's what this PR is for. Making the model a bit beefier is certainly a good suggestion. Anyone else?

kilbith:
> How about I leave all the cobble walls in (normal, mossy and desert cobble - 3 total) ?

Perfect.

sofar:
Also, I think we can make the model a bit more interesting, and give it the "rustic" appearance.

Here's a slightly thicker wall (6px thick as you requested), but with bevelled tops:

![bevelled tops]
(http://i.imgur.com/jdNx6WG.png)

kilbith:
Hell, not that :-1: Keep it simple, please.

Detailed nodebox models are also unoptimized (lot of hidden faces/vertices), causing troubles on low-end hardware when these are massively placed.

sofar:
You either are cheering as the greatest fan, or vehemently oppose.

How about we let some more folks chime in, first? Unoptimized node boxes seems like an inconsequential issue to me at this point.

kilbith:
> You either are cheering as the greatest fan, or vehemently oppose.

Because I exactly know what this game needs or not, my experience talks in my favor about what thing will succeed or fail. Any move distancing the blocky design and the KISS principle is unwelcome, visually and performatively.

rubenwardy:
What experiences are you talking about? Genuine question.

tenplus1:
+1 for simple walls, I can already feel the lag on the bevelled look.

sofar:
@tenplus1 I'm assuming you're serious, so, I will consider it seriously as well.

red-001:
I prefer the thinner walls, they would look even better if they where higher.

MoNTE48:
:+1:  +1 for simple walls, too

sofar:
While I'm totally not convinced that shaving off the nodeboxes is going to make a drastic framerate, I have pushed a version that's as dull as could be:

![whatever]
(http://i.imgur.com/I1KbnkS.png)

If this has people's preference, I can live with that, and I'd rather see these go in than stay out of minetest_game.

sofar:
BTW I do prefer the thicker look myself.

rubenwardy:
This adds quite nice decoration to things like cottages etc. However, fences already sort of do this, so I'm a little indecisive.

If I were making a subgame, I'd include something like this, because I like they'd look nicer than fenses for certain builds. However, I would probably add more things to tie into it to make it make more sense, like arches and window boxes etc (although they would need to keep torwards the voxel style, and not be homedecor-ish).

I'm leaning towards a :+1: at the moment.

What's the game play difference from this vs fenses? Are they the same apart from looks?

> Because I exactly know what this game needs or not, my experience talks in my favor about what thing will succeed or fail. 

This strikes me as a little absolute, arrogant and assertive. Just because you have contributed to minetest_game before, and have been active in other ways, doesn't mean that you're automatically that you're the authority on this. Yes, you do have the right to an opinion - but that's not what this sounds like, from the way you phrased it it's very certain and assertive. Particularly the "exactly" part. I guess that this isn't really your intent or what you meant. So I'd appreciate a clarification.

However, saying that, I don't think wood walls makes sense with this draw style. I think that this style should be limited to stone and mossy. Brick walls possibly. I like the bevel on top of the wall, but not on top of the post.

sofar:
The walls do not have an extended collision box, so they do not block anyone attempting to jump over.

sofar:
BTW, it is possible to make these nodeboxes extend vertically, and make them unjumpable(*).

![unjumpable]
(http://i.imgur.com/rZOECEE.png)

The drawback to this method is that the collision box is calculated on the nodeboxes, so you can't make them obstruct a jump and not block the view at the same time. Although it somewhat works out OK. 

Plus it seems something that fits in an extended wall mod.

(*)Having them taller still doesn't prevent shift-jumping.

rubenwardy:
I like how you can jump over them, it makes them different from fences.

rubenwardy:
I'd also prefer it if there was only a post every corner, not every node.

sofar:
```a post every corner```

@rubenwardy that's a long discussion thread, and I'll attempt to explain why this isn't implemented.

First, consider the logic needed to make this work (another flag, inconsistent visuals)

Second, if a block is placed on top, would you want the post to be drawn? (more options in the nodedef?)

Now, consider that a simple nodebox node with a single fixed nodebox, rotatable, could be trivially maintained and updated by a lua mod to provide that visual aspect for exactly the circumstances where you'd want that to happen.

I've thought this over quite long, and, I realize that I probably can't pleasure anyone in a PR with an approach that attempts to do just that, so, the best way forward is to leave that kind of change to a mod.

Hence, these walls are simple. They can be easily enhanced to do what you want in lua (without making the node count use 15+ nodedefs, like current existing implementations do) with just 1 extra nodedef.

paramat:
I prefer the simpler wall design without bevels, more consistent with eveerything else we have.

C1ffisme:
People are going against the bevel top wall...

In my opinion, it looks good, but if it's really causes too much lag than I suppose we shouldn't add it.

0-afflatus:
I don't like this. IMO xwalls mod does it better. A simple uniform thin wall piece is enough, if you want posts you can use normal nodes. It's better than the ugly stone fences, but ultimately I don't actually see the need for this at all. :open_hands:

sofar:
@0-afflatus this mod doesn't replace xwalls, however it uses a new game node type that massively allows to reduce node defininitions. In order for mods like xwalls to understand how to benefit from that, there needs to be examples and ready to use code, which this patch adds.

0-afflatus:
@sofar I think I'm not quite understanding your point. Mine is aesthetic and from a building POV, the shape of your walls seems overcomplicated to me. I just think building blocks should be the simplest possible shapes, otherwise they start to dictate building style.

sofar:
@0-afflatus The name "walls" is perhaps confusing. This isn't a block that usable for the wall of a house, since it leaves obvious gaps. This mod's use of the connected node drawtype is more for decorative brick countryside stone fences. Your use of this block type is entirely voluntary, and you're free to make subgames that exclude these nodes.

0-afflatus:
Sure, that's why I'm trying to avoid voting on most MTG issues. My opinion isn't strictly relevant here.

sofar:
rebased.

sofar:
Updated the craft recipe to return 6 items, not 1

C1ffisme:
@sofar That makes slightly more sense than the original recipe.

sofar:
I was testing this in survival mode, and having to dig 6 cobble stone for 1 wall segment was brutal :)

C1ffisme:
@sofar Yeah, I built things using lots of fences in Minecraft, and getting the wood was a *pain*. This is probably a good balance. And more realistic.

paramat:
:-1: Sorry, beautifully done but i feel unnecessary.

C1ffisme:
@paramat Yes, but they don't really look good. They aren't as nice looking as this pull request.

And I'd like to have an example node to read if I need to make an NDT_CONNECTED node.

sofar:
I'll likely submit a PR to replace the ```default:fence``` nodes to use this mechanism as well, so that will likely be the canonical example code available.

C1ffisme:
@sofar Probably a good idea. We should completely remove the fence drawtype for good.

sofar:
I still prefer this to be added, instead of having dozens of other mods implementing something almost the same but perpetuating the non-connected node stuff with 5+ nodedefs, sometimes more.

Ekdohibs:
:+1:

sofar:
- Updated with API changes

sofar:
Slight update to the top of the wall. When something is put on top, it grows the last pixel to make it usable as a pillar.

paramat:
How about making the pillars one node high to start with? makes more sense and is simpler.

sofar:
Kilbith suggested the same as well, so I'm more than welcome to do so. I'll change it.

paramat:
f8f7502e3253d50e4e0829fdd6e2ad7aab2e4cd3

PR #805, opened 2016-01-25
Make ocean biomes more attractive.

sofar:
Place larger clay blobs in all ocean biomes, but only at deeper
water depths (-10 through -50).

These biomes are very unattractive, and clay is relatively
sparse in the game already. These larger blobs may provide
an attractive resource for players to explore and mine in
the underwater areas. Since these will always be close to
the seafloor, they are likely more risky to dig due to
potential drowning.

In order to use the biomes option for ore registration, we
need to register biomes before ores.

### DIFF
--- a/mods/default/mapgen.lua
+++ b/mods/default/mapgen.lua
@@ -1394,8 +1394,8 @@ if mg_params.mgname == "v6" then
 	default.register_mgv6_decorations()
 	minetest.register_on_generated(default.generate_nyancats)
 elseif mg_params.mgname ~= "singlenode" then
-	default.register_ores()
 	default.register_biomes()
+	default.register_ores()
 	default.register_decorations()
 	minetest.register_on_generated(default.generate_nyancats)
 end


sofar:
@paramat just an idea here... we could do the same for placing sand blobs in dirt, which would diversify the seafloors drastically.

rubenwardy:
The title is misleading, really the commit should be called "Add large clay blobs at low sea depths".
Making the ocean more attractive is a rational, not a real change.

I like the idea, but haven't tried it out to see the consequences yet.

RHRhino:
Sry @sofar but I doubt that some more clay would make the ocean much more attractive for players. I think the only way to make it more attractive would be a new resource that can only be found there and nowhere else.

Jordach:
What about a new type of sand, and minerals - like cyanic copper, or
aquamarine glass (9 fragments -> aquamarine grit -> furnace -> aquamarine
glass). Cyanic copper could be melted into a stronger form of copper -
cannot be alloyed, but strong as steel [or mese?] due to the salty contents.

Another nice thing would be to add some sort of coral type object, in a few
colours and patterns to influence builders. (Light emitting ones would also
be nice!)

paramat:
Good idea to register biomes before ores, this will be necessary for the future.
For simplicity i would prefer to just change the lower limit for the current clay blobs. Blob ore uses expensive 3D noise so i'm not keen on larger blobs.
Why specify the biomes when y limits will keep blobs to the oceans?

sofar:
@Jordach I prefer to use the tools already present - without adding new nodes, first. With some creativity using the nodes we already have, I think we can come a long way.

sofar:
@paramat I'll take a look at reusing the existing blob definition.

sofar:
@paramat the idea was to make ocean biomes more interesting, not introduce more clay everywhere. The restriction to ocean biomes accomplishes that. Perhaps we should restrict it to mgv7 alone, so the biome selector can be used.

sofar:
Changed this to just commit the order reversal.

I think I still want to use biomes and do a bit more about the sub-oceanic surface in v7 only - adding clay and gravel sheets may be the way to go, as I think sheets would do just fine, and will be less resource intensive - even if larger.

This can probably already be merged.

paramat:
Oops, i see now that allowing it in dirt requires the ocean biomes to be specified.

sofar:
Yeah, I'm giving this one a bit more thought before adding an actual change in seafloor material.. We'll revisit.

paramat:
I have been thnking of making one of the biomes have a gravel beach for some variety.

sofar:
Yes, ideally it would fit well in a colder biome - ice+gravel would do well as a combo.

I would still make it variable between sand and gravel, and not monotonously gravel only, if possible.

paramat:
38a835132758c1df0e345b270e3dc973dc5cfda1

PR #807, opened 2016-01-25
Big simplification of slabs combination

kilbith:
Why doing things complicated when you can make them simple ?

### DIFF
--- a/mods/stairs/init.lua
+++ b/mods/stairs/init.lua
@@ -110,6 +110,14 @@ end
 -- Register slabs.
 -- Node will be called stairs:slab_<subname>
 
+local function get_node_face(pointed_thing)
+	local ay = pointed_thing.above.y
+	local uy = pointed_thing.under.y
+
+	if     ay > uy then return 1	 -- Top face pointed.
+	elseif ay < uy then return 2 end -- Bottom face pointed.
+end
+
 function stairs.register_slab(subname, recipeitem, groups, images, description, sounds)
 	minetest.register_node(":stairs:slab_" .. subname, {
 		description = description,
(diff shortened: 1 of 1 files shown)


everamzah:
Seems alright at first, but placing a slab adjacent on the side results in a combined node.  Normally, doing this would result in an extension, like a bridge.  I like it, though, it's fast.

Jordach:
James, I think building support structures is more fun - then you've got to
plan your buildings correctly, and not relying on silly "game physics".
Scaffolding is cheap, if you have the dirt.

kilbith:
> Seems alright at first, but placing a slab adjacent on the side results in a combined node.

Can you be clearer, please ? You mean if the first slab is placed on its side ?

everamzah:
Okay I will test it out.  I was just thinking about how the on_place callback is important here for other reasons: such as checking protection!

everamzah:
I will attempt to be more clear: Placing the slab on the side of another slab normally results in another slab.

With your patch:
screenshot_20160125_082507

Current behavior:
screenshot_20160125_082810

everamzah:
Jordach does, however, bring up a very interesting point about gameplay here, and I think he's onto something.

kilbith:
Ah, right. Well you can hold sneak + right-click for placing slabs alongside.

everamzah:
That's true, and should be good enough.  The problem I find is that you have to press sneak+RMB to place anything at all to the slab!  Also, protection is not being checked.

kilbith:
Protection check added.

everamzah:
Protection check works as expected.  However, I find it awkward that one cannot place another node to a slab without holding down sneak.  Is it reasonable to check that the player isn't holding a slab of the same type?

kilbith:
Fixed.

paramat:
Beautifully simple but makes building slabs to slabs more difficult, builders won't accept it. So unfortunately :-1:

kilbith:
Oh yes, pressing sneak is very difficult. Too bad missing a much faster code.

C1ffisme:
@paramat Actually, when I'm building roofs I hold my sneak key down, since falling off would mean getting sand/dirt to get back up again.

Although I can see a problem with this if you were flying, since the sneaking and descent keys are the same key.

everamzah:
Personally it is not a matter of having to depress an extra key, kilbith, it is more a matter of breaking expected behavior.  If you right click a node that is not a furnace or a chest you get an adjacent node.  I can already envision the first bug report entitled "can't place anything on slabs".

kilbith:
What are you talking about ? You can place nodes on slabs currently (without sneak), check latest modifications.

everamzah:
I didn't see your "fixed" comment, and hadn't checked before commenting.  My apologies!

I tested now and see everything Looks Good To Me!

kilbith:
OK I've solved the slab-to-slab thing by checking the top face of the pointed node.

Now the feature works exactly than before, but is much faster.

@paramat I'd like a re-evaluation / testing, please.

everamzah:
They're not snapping to the bottom of another slab, like the original method does.

kilbith:
Fixed.

everamzah:
Adding to the side of a slab attached to the ceiling places the slab below.
screenshot_20160125_233917

everamzah:
Here, too
screenshot_20160125_234111

paramat:
Okay seems potentially good, although i'm not familiar with current slab behaviour as i don't build much.

kilbith:
Currently there's just one thing remaining to solve : placing slab with facedir at 20 under a regular block, still using the same callback.

Consider this PR as WIP, but is worth it. I leave it inactive for now.

everamzah:
When this is all said and done, I think you should consider using minetest.place_node() instead of minetest.set_node(), in order to register the placement of the slab with the on_place callback.

This may remove the need for the protection check you added.

The rational is some mod checking the number of slabs placed.  I did a quick search and didn't see any old mods relying on this, offhand, but one example would be using rubenwardy's awards mod and registering an award for number of slabs placed.

kilbith:
You're missing the essential point of using `on_rightclick`.

The current method is : placing a slab above -> removing this slab -> placing a new node below.
The new method : placing a new node at pointed position.

It's more efficient.

everamzah:
Yes, but doing so without registering this placement may not be proper.  This was my initial thought in response to what is otherwise a very "snappy" enhancement!

Har har.

mods/stairs/init.lua
```diff
@@ -124,87 +132,31 @@ function stairs.register_slab(subname, recipeitem, groups, images, description,
...
-				-- If the item was taken from the fake stack, decrement original
-				if success then
-					itemstack:set_count(fakestack:get_count())
-				-- Else put old node back
+			if wield_item == "stairs:slab_" .. subname and node.param2 <= 3 and face then
+				if face == 1 then
+					minetest.set_node(pos, {name = recipeitem, param2 = node.param2})
 				else
-					minetest.set_node(slabpos, slabnode)
+					minetest.set_node(pointed_thing.above, {name = wield_item, param2 = 20})
```
ShadowNinja: What if pointed_thing.above is a solid node (can happen if it's a nodebox).
You should use a helper like item_place_node for this, there are a lot of things to get wrong.

paramat:
Since kilbith has left, anyone feel like finishing this, if indeed it is a good method?

paramat:
Will close in 1 month, please copy the code for your use if you intend to rework this.

sofar:
possibly related: #1109 

Maybe I'll brush this up, since #1109 exposes a simple issue and I think we should incorporate some elements of this PR in it (if not whole).

paramat:
Replaced by #1111

PR #855, opened 2016-02-16
~~~Hunger.~~~ Stamina

sofar:
This code adds ~~~hunger~~~ stamina to minetest_game.

This code is entirely derived from Blockmen's hunger+hud mod.  Credits
and copyright are preserved as they are compatible with minetest_game.

gameplay:

The player stat 'stamina' is added. The higher the value, the better. A
value of 5 or higher causes the player to slowly get healed over time,
at the cost of stamina. One can only be healed if the stamina value is
higher than, or equal to the health value.

Eating food causes 'stamina' to increase. The visual range of stamina
is 0-20, but one can eat up to 30 stamina. If the stamina is at max,
items are not eaten anymore.

At stamina 2 or lower, the player is damaged periodically, and will
die if the stamina value isn't increased due to eating.

If eating a food that has a negative hp_change value, such as red
mushrooms, the player takes periodical damage to the amount of the
hp_change value. During this time, the player is "poisoned". This
poison damage can not kill the player.

Added by me is code to add stamina on crafting. I intend to add
stamina based on combat as well, at least pvp combat.

code:

This mod started as BlockMen's hunger+hud mod. I trimmed the entire
hud mod away and placed the stamina bar on top of the hp hud element,
which will make the relation between stamina and hp a bit more clear,
and doesn't require moving the air hud element.

The stamina mod was reduced to fit a single file. I trimmed some
legacy code, revised all local/global usage. Renamed most internal
variables, and generally did a lot of removals of needlessly complex
code favoring simplicity.

The food registration code is entirely removed. Instead we assume food
just adds stamina or ticks damage as poison. Food eating callbacks are
still called. Subsequently, this mod will likely not properly work
with BlockMen's hud code anymore, nor is that intended to work. Food
mods just have to call item_eat().

I've moved and credited authors for the used textures. I replaced
the food sound of this mod with 3 different sounds.

The code points out several issues with core: There's no decent
place to store player stats. This mod solves it by creating a per
player inventory purely for storing an integer value. Intercepting
core.do_item_eat() also isn't the nicest way to make this work.

- [ ] - Remove ItemStack abuse


sofar:
reserved

mods/hunger/README.txt
```diff
@@ -0,0 +1,45 @@
+Minetest mod "Hunger"
+=====================
+
+(C) 2015 - Auke Kok <sofar@foo-projects.org>
+(C) 2015 - BlockMen
```
BlockMen: Other way round and 2016 for you...

mods/hunger/init.lua
```diff
@@ -0,0 +1,317 @@
...
+	end
+	local name = player:get_player_name()
+	local level = hunger_players[name].level
+
+	level = math.min(HUNGER_MAX, level)
+	level = math.max(level, 0)
+
+	inv:set_stack("hunger", 1, ItemStack({name = ":", count = level + 1}))
+	return true
+end
```
BlockMen: Well, I know that this "method" works, but I'm stongly against using this hack as "official way" to save the hunger.
Thats one of the reasons why I open the PR for the engine.

BlockMen:
I don't see any profit from ripping hunger appart from hud. If we add new statbars we should improve current "shit" aswell.

:-1: from me as is, mainly beccause of the hacky way of storing hunger value.

rubenwardy:
:-1: due to hacky way of storing hunger

mods/hunger/init.lua
```diff
@@ -0,0 +1,317 @@
...
+local hunger_timer = 0
+local health_timer = 0
+local action_timer = 0
+
+local function hunger_globaltimer(dtime)
+	hunger_timer = hunger_timer + dtime
+	health_timer = health_timer + dtime
+	action_timer = action_timer + dtime
+
+	if action_timer > HUNGER_MOVE_TICK then
```
LeMagnesium: Couldn't you use `minetest.after` instead of a globalstep callback? I understand that there are multiple tick values taken in count here but the use of timers and globalstep callbacks called every server step have a cost.
(see [this](https://github.com/MinetestForFun/server-minetestforfun/issues/179#issue-97929624) for more explanations)

C1ffisme:
> :-1: from me as is, mainly beccause of the hacky way of storing hunger value.

Erm... But didn't you write it? Why not post an issue like "[Suggestion] Add a better player value storage method"

I personally really don't care if it is a hack. Hacks can be replaced.

PilzAdam:
> Eating food causes 'hunger' to increase.

This doesn't sound right-

> The visual range of hunger is 0-20, but one can eat up to 30 hunger. If the hunger is at max,
items are not eaten anymore.

Why? Whats the benefit from hiding this?

raymoo:
Could there be a function exposed for increasing exhaustion? In case a mod wants to add exhaustion to some of its actions. Or is it intended that mods just read directly from hunger_players?

sofar:
>> Eating food causes 'hunger' to increase.
>This doesn't sound right.

I agree. I was thinking of renaming the stat to "stamina", which seems more intuitive.

>>The visual range of hunger is 0-20, but one can eat up to 30 hunger. If the hunger is at max,  items are not eaten anymore.
>Why? Whats the benefit from hiding high values?

This is what the original implementation did. The benefit is that if you eat a little bit over the max of 20, it's not immediately lost. I'm completely neutral on this, but I changed the mechanic to consume hunger when it's healing, and so the extra buffer may be nice.

sofar:
@raymoo I've thought about that as well, this function also has a poison mechanic, which would conflict with work that you're doing in your mods as well. I think we'll want a simple API to allow mods to manipulate hunger values, though.

mods/hunger/README.txt
sofar: sure, will update.

mods/hunger/init.lua
sofar: I'm 100% in agreement with you on this.

mods/hunger/init.lua
sofar: This function likely needs a full rewrite to optimize. However, I don't think I'll be able to remove the globalstep part entirely, mostly because the proper replacement using ```minetest.after``` would cause lots of timers to fire per player, and you'd end up with more timer pressure from that.

sofar:
I've renamed the whole thing to stamina. I don't think in-game people will have a hard time understanding what the idea is, so we might as well choose a name for the stat that is logical (e.g. increased is better, whereas hunger is inverse). Updated (C) comment as well.

Meanwhile I'm publishing the mod for those who want to use the codebase as mod here:
https://github.com/minetest-mods/stamina

rubenwardy:
Stamina sounds like a sprint/running thing

sofar:
At least higher values of it suggest that it's better for the player.

mods/stamina/README.txt
```diff
@@ -0,0 +1,45 @@
+Minetest mod "Stamina"
+=====================
+
+(C) 2015 - BlockMen
+(C) 2016 - Auke Kok <sofar@foo-projects.org>
+
+
+About this mod:
+~~~~~~~~~~~~~~~
+This mod adds a stamina, or "hunger" echanic to Minetest. Actions like
```
rubenwardy: > echanic

mods/stamina/README.txt
```diff
@@ -0,0 +1,45 @@
...
+If a player is low on stamina, they start taking periodical damage,
+and ultimately will die if they do not eat food.
+
+Eating food no longer heals the player. Instead, it increases the
+stamina of the player. The stamina bar shows how well fed the player
+is. More bread pieces means more stamina.
+
+
+For Modders:
+~~~~~~~~~~~~
```
rubenwardy: Subsections should be ----------- to be consistent

```
Header 1
=======

Header 2
------------
```

C1ffisme:
Well... Erm... Creative thinking on the idea!

Although if it is stamina/energy, what about making sleep recharge half a point of stamina?

sofar:
>Although if it is stamina/energy, what about making sleep recharge half
>a point of stamina?

Possibly, yes. I do want to force players to farm food, though, so it should not give much stamina back. Perhaps we can give it a point of stamina only if the stamina bar is less than half.

I've also considered making stamina loss higher at night to make beds more appealing.

0-afflatus:
I like that item_eat doesn't have to be overridden. That is important to me.
I'm not keen on adding an extra player stat, could someone explain why this is necessary? As everyone else seems to accept the necessity of this I guess I'm just failing to understand something.
I like very much that this doesn't depend on majorly modifying the HUD, that was one of the reasons I have never used Blockmen's mod.
I'd prefer that stamina loss or whatever is consistent. I don't like the idea of it being different at night.
The simpler the better I say and if it has to be brutal sobeit.

mods/stamina/README.txt
sofar: thanks, I'll pick these up

tenplus1:
A feature like this I would leave in the hands of a mod and not in the game engine since you get more control over the mod itself and what heals, hurts, satiates and poisons.

BlockMen:
@sofar I don't see how renaming it to "stamina" solves the mentioned problems. Still :-1:

sofar:
@blockmen, you don't need to repeat that, I had no illusions that that would solve the issue.

everamzah:
Internally it's called "Satiation" -- which is correct when full.

sofar:
- added exhaustion when fighting
- added exhaustion when jumping
- removed cap of 30. stamina no longer has a cap, but you can't eat more food if it's at 20, this means that effectively food isn't ever wasted.

C1ffisme:
@sofar Yeah, I was thinking something like that.

> removed cap of 30. stamina no longer has a cap, but you can't eat more food if it's at 20, this means that effectively food isn't ever wasted

This will save many wasted apples. The next step is to make an eating animation. But maybe not now.

MoNTE48:
@C1ffisme eating animation
https://forum.minetest.net/viewtopic.php?f=11&t=14006
Look very good with 32 textures!

C1ffisme:
@MoNTE48 Hmm... Not quite what I was going for. What I really meant was how eating works in Minecraft, where it takes time to eat, and you can't just heal instantly. That might make more of a challenge on PVP servers.

sofar:
Eating food particles (it's not really an animation) shouldn't be part of an `item_drop` mod. I really dislike how some people are making kitchen sink mods :(

The food particles in the screenshot on the forum look terrible, but I suppose it can be cleaned up and it's certainly appropriate to add it to this mod, where it's entirely appropriate (I already added eating sounds, as well).

MoNTE48:
@sofar, Yes, when I use the form of the third person, these particles are simply awful. \
As it has no place in item_drop. I do not like it when people combine several independent functions.

C1ffisme:
This is the eatimation I was talking about. Where you pull the food towards you, and shake it up and down while particles apear.

sofar:
@C1ffisme that's probably entirely doable in lua, actually.

C1ffisme:
@sofar Ooooh... I'm not sure about the wielditem HUD movement, though. Is that possible yet?

MoNTE48:
@sofar, really? I wait look your realisation :)

MoNTE48:
@C1ffisme Yes, there are api to wield the speed change

sofar:
@MoNTE48 I'd create an entity with the texture of the thing beaing eaten, and embed an animation that shakes it up and down in front of the player's view.

C1ffisme:
Erm, no @MoNTE48 I was talking about the movement itself, not the speed. For example, instead of the item going in the regular circle that happens when you punch, it turns towards you and shakes up and down.

@sofar Good idea...

MoNTE48:
@sofar This will be done for each eat item?
This is more works...

sofar:
@MoNTE48 solving one problem at a time :)

MoNTE48:
@sofar ok. You could put a famine after the health (as in Minecraft), instead of the above health? It look ugly.

sofar:
@MoNTE48 not without moving the breath bar.

MoNTE48:
@sofar and moving breath bar too. That is as it is made in Minecraft (and the way made in BlockMen) - the most convenient and beautiful option, I think.

sofar:
I have no intention of moving it at this point. The argument that it should be like MC is moot. I like that stamina sits on top of health, it shows the relationship between them.

MoNTE48:
But normally too many objects on the left and right quite empty. :-1:

sofar:
This PR isn't going to get merged any time soon (or possibly never), so the thumbs downs are unnecessary.

MoNTE48:
Yes, but I really like it and I would use it in my subgame.

sofar:
I'd like this thread to focus on the lack of player metadata storage, instead (which is honestly the whole point of this PR).

If you want to discuss the stamina implementation, let's do that here: https://github.com/minetest-mods/stamina/. We can take the stamina mod to a different level in that location anyway.

MoNTE48:
https://github.com/minetest-mods/stamina/issues/1

BlockMen:
> Yes, but I really like it and I would use it in my subgame.

@MoNTE48 since this is actually my hunger mod, what excactly is wrong for you with https://github.com/BlockMen/hunger ? Without any feedback how should things get improved...

MoNTE48:
BlockMen, Your hunger is working very well, but there are some times when it works with a delay and requires optimization. In addition, I would like to see the food animation if this can be done sofar

BlockMen:
> but there are some times when it works with a delay and requires optimization

@MoNTE48 could you be a bit more precise? Maybe open an issue for that at https://github.com/BlockMen/hunger

MoNTE48:
@BlockMen, https://github.com/BlockMen/hunger/issues/9

davisonio:
Should be added to core in my opinion.

BlockMen:
@davisonio Could you post your opinion here (https://github.com/minetest/minetest/pull/3388) plz? That is the main PR for hunger and the engine is visted by far more ppl, so we get a much better view on what should be done concerning hunger. thx.

sofar:
I'm closing this for now.

The original mods and my fork remain available and I am maintaining my simplified/combined hud/hunger mod for everyone to use.

I'm not done with this idea, but not in this form, just now.

PR #861, opened 2016-02-21
Let mods specify a modname in doors.register def.

oleastre:
This let's mods to use the doors api and create nodes in their own namespace.
This makes it easier to switch from the old api to the new one and solves #854 (the door node name contains the mod name).

To register nodes in your mod namespace, just add a "modname" attribute in the doors definition.

On the implementation level, I've let the door basename attribute, and added a "fullname" that's the concatenation of basename and modname. 
In fact, the modname attribute could be dropped and rollback to the old way of doing by registering doors with the modname included in the door name like "my_mod:my_door".

I know this tries to solve the same issue as #858 but I find it more clean for mods that want to switch to the new api.

### DIFF
--- a/mods/doors/init.lua
+++ b/mods/doors/init.lua
@@ -119,7 +119,7 @@ function _doors.door_toggle(pos, clicker)
 	local meta = minetest.get_meta(pos)
 	local state = meta:get_int("state")
 	local def = minetest.registered_nodes[minetest.get_node(pos).name]
-	local name = def.door.basename
+	local fullname = def.door.fullname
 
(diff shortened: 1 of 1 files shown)


sofar:
This doesn't solve the same issue as in #858, and could in addition be merged on top of it. I'll check and see if the patches can both apply without conflicts.

sofar:
In light of the comment above, would you mind if I rewrite this PR and submit a version of it myself?

sofar:
I've pushed a modified version of this patch with credit mention in #858

PR #874, opened 2016-02-25
Allow non-players to dig bones

MT-Modder:
This will fix https://github.com/minetest-technic/technic/pull/257. And probably any future problems involving Technic/Pipeworks Machines.

### DIFF
--- a/mods/bones/init.lua
+++ b/mods/bones/init.lua
@@ -43,7 +43,13 @@ minetest.register_node("bones:bones", {
 	
 	can_dig = function(pos, player)
 		local inv = minetest.get_meta(pos):get_inventory()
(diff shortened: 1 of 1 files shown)


NathanSalapat:
While I'm no expert this looks good to me. I was actually just about to start working on the exact same problem for the guys over at the Pixel Shadow.

MT-Modder:
Looks like locked chests have similar code. I'll wait for devs to see if it should be included in this PR or not.

PilzAdam:
Looks good, although the code can be shortened a bit.

MT-Modder:
@PilzAdam I see it now :smile:. I'm gonna redo with the changes to locked chests and steel doors.

PR #877, opened 2016-02-26
Add a simple, but stylish fence gate.

sofar:
This fence gate builds on NDT_CONNECTED by assuming fence
nodes will automatically connect to it's sides properly. The
fence gate will open and close just like doors, with sounds,
but it only opens one way. The gate sticks out quite a bit and
can be bumped into, so the fence may be used as some sort of
path switch.

This change requires #873, otherwise it looks bad.

- [X] - Add craft recipe
- [x] - Custom sounds
- [x] - move to doors mod


Napiophelios:
I think the open/close sounds from your animated chest would be cooler.
They are kinda creeky sounding.

sofar:
@Napiophelios I've attempted before to replace the steel door sounds, and I think the current door sounds aren't really a good fit for the fence either. I'm not sure the chest sounds would be nice here, maybe I'll look for some new sounds.

0-afflatus:
I love this idea. Nice one!

C1ffisme:
:+1: 

This is a good idea for those in minetest who want to make better looking farms or barriers.

NathanSalapat:
:+1: 
We need gates, fences are hard to use when you can't control traffic passing through.

paramat:
Beautiful work.

sofar:
@paramat I didn't necessarily wanted to make this a separate mod initially, that's just what I used while testing. Do we want this to end up in default? I'm kinda neutral on that. It may be nice to split up some things a bit.

C1ffisme:
@sofar Fence gates could go in the same mod as doors, since they are similar objects.

sofar:
I don't intend to make them work like doors, or implement the doors API, though. And fences are in default.

C1ffisme:
They don't work exactly like doors, but they are very similar. The doors mod could hold "door-like" objects, since the default mod holds "default-like" objects.

sofar:
maybe split up fences into a mod, put the gate with the fences, and alias the nodes to default:?

C1ffisme:
@sofar Hmmm... Two nodes still sounds like a bit too little for one mod, although I suppose this is the case for TNT.

sofar:
I've adjusted the collision box such that if you use two adjacent opposed fence gates, you can move entities through the gates larger than 1.0. This makes it easier to coerce animals through these gates.

rubenwardy:
I like this

0-afflatus:
Animals don't like 1 node gaps.

paramat:
Because this is similar to doors i think it's fine as a seperate mod. I think keep fences in default to avoid headache.

sofar:
@Animals don't like 1 node gaps

With 2 gates open facing eachother, the gap is 1.8+ wide.

sofar:
I've added non-default door sounds to the gate.

kilbith:
I wouldn't make a separate mod for this specific thing. It's a door, put it in `doors`.

sofar:
I may end up moving this into doors.

0-afflatus:
> It's a door, put it in doors.

That would be the most logical.

C1ffisme:
:+1: for putting it in ```doors```.

paramat:
Saw the video, sounds are good. In doors mod is okay by me.

sofar:
- moved do "doors" ("doors:fencegate_open|closed")
- renamed relevant files
- merged README parts

Nice thing is that searching for "doors" makes it show up, but also "fence" makes it show up in search.

Ekdohibs:
:+1:

C1ffisme:
@sofar That's pretty awesome, you can search for 'doors', 'fence' or 'gate' to find it. So if you need a cowboy saloon *door*, or an entrance in your *fence*, or you just need a *gate* for whatever reason, it's easily searchable.

0-afflatus:
The other thing that I think is particularly awesome about this idea is the the gate actually aligns with the fence. It annoys me intensely that doors don't align with fence-like nodes - I totally understand the limitations, but it does look crap.

paramat:
:+1:

paramat:
:+1: When those are done.

sofar:
Fixed indentation.

paramat:
:+1:

sofar:
BTW, If we can merge the connected_nodes patch, then I will change this node to only allow ```connect_sides``` to connect from the appropriate sides. This would help prevent half-fence bars extending into air from 2 directions to this node.

sofar:
with ```connect_sides``` in place - fences won't connect unless from the side.

Fixer-007:
:+1:

Thomas--S:
:+1:

sofar:
- Updated with API changes

paramat:
b998ab1b3ae8cb7f21fab30819a013b7d8e00aec

PR #879, opened 2016-02-28
Only register dye crafts from flower group if flower exists

ghost:
This makes a recipe only for purple, white, orange, red, yellow, blue -- currently, all colors are registered for dye crafts.

Fixes https://github.com/minetest/minetest_game/issues/878

### DIFF
--- a/mods/dye/init.lua
+++ b/mods/dye/init.lua
@@ -39,12 +39,19 @@ for _, row in ipairs(dyelocal.dyes) do
 		description = description,
 		groups = groups
 	})
-	minetest.register_craft({
-		type = "shapeless",
-		output = item_name.." 4",
-		recipe = {"group:flower,color_"..name},
-	})
+	local flowers = {"white", "violet", "blue", "yellow", "orange", "red"}
+
+	for i=1, #flowers do
+		if flowers[i] == name then
+			minetest.register_craft({
+				type = "shapeless",
+				output = item_name.." 4",
+				recipe = {"group:flower,color_"..name},
+			})
+		end
+	end
 end
+
 -- manually add coal->black dye
 minetest.register_craft({
 	type = "shapeless",


Napiophelios:
:+1:  should flowers? be added to depends.txt or does it really matter?

everamzah:
Dye has no dependencies as there are no references to exterior functions/tables.

kilbith:
The whole file needs a rewrite. See the tables above and how the values are combinated (wrongly).

mods/dye/init.lua
```diff
@@ -39,12 +39,19 @@ for _, row in ipairs(dyelocal.dyes) do
...
 		groups = groups
 	})
-	minetest.register_craft({
-		type = "shapeless",
-		output = item_name.." 4",
-		recipe = {"group:flower,color_"..name},
-	})
+	local flowers = {"white", "violet", "blue", "yellow", "orange", "red"}
+
+	for i=1,#flowers do
```
rubenwardy: space needed

PilzAdam:
:-1: The dye mod is independent of our flower mod. Just because our flower mod doesn't have brown flowers doesn't mean that you can't craft brown dye from brown flowers.

C1ffisme:
@PilzAdam Easily fixed. Add this table to flowers, then add "flowers?" to the dye mod. If no flowers mod exists, then no crafting registry is preformed, if it does exist, it uses the flowers table. Then dyes can be crafted from whatever you register in any other mod (carrots -> orange dye, etc.).

PilzAdam:
@C1ffisme there is nothing to "fix" since there is no problem. The dye mod should register the desired recipes for all possible colors of flowers.

C1ffisme:
> The dye mod should register the desired recipes for all possible colors of flowers.

Eh... This doesn't sound like very good programming. Mods should define their own recipes, not the dye mod define recipes expecting other mods. I would rather see the recipes for dye in the flowers mod.

kilbith:
> The dye mod should register the desired recipes for all possible colors of flowers.

The `dye` mod should not register useless recipes with groups that no items use.

A small verification could be added to this mod : 

``` Lua
for _, def in pairs(minetest.registered_items) do
     if not def.groups[<color_group>] then break end
     minetest.register_craft(...)
end
```

paramat:
:-1: PilzAdam is correct.

sofar:
@kilbith's check can only happen if the ```dye``` mod is loaded after all the possible other mods that add flowers. That would mean that the ```dye``` mod's ```depends.txt``` lists them all.

I don't think that's reasonable. It's not ideal, but I prefer the current situation.

kilbith:
@sofar Wrong. `minetest.registered_*` intervene after all mods have been loaded. I checked.

0-afflatus:
:-1:

sofar:
@kilbith yes, using a ```minetest.after()``` would likely work, but removing recipes already exposed to mods and then removing them is quite unfriendly. Mods may have depended on ```dye``` and looked at the recipes that it created, for instance.

PR #885, opened 2016-02-29
Doors: Prevent placement in protected areas.

sofar:
https://forum.minetest.net/viewtopic.php?f=42&t=1523&start=500#p208773

Currently, doors can be placed inside others' protected areas. We
need to check protection for both bottom and top halves.


sfan5:
00a0d9a5d4bda9a3a33d1992218c3e8094559730

PR #906, opened 2016-03-11
Add Flint and Flint-and-Steel (producing fire)

kilbith:
Reasons : 
- Fire burns, wielding it in hand does not make sense and looks odd.
- Giving the possibility to place fire in survival mode.
- 2 fire items in the inventory looks redundant.

Featuring : 
- Flint is obtained by digging gravel with a rarity of 16.
- Flint-and-Steel is craftable with a flint and a steel ingot.
- Flint-and-Steel can be used 64 times.

Textures by Gambit.

capture d ecran de 2016-03-11 18-36-45

capture d ecran de 2016-03-11 18-36-39

Cc: @PilzAdam @paramat @sofar


RHRhino:
Wow the textures look great! :heart_eyes:

C1ffisme:
Nice! :+1:

kilbith:
Please don't post this kind of comments unless if you're a maintainer.

Just add your reaction in the 1st comment (smiley on the top-right).

sofar:
I'd personally really had preferred if this wasn't a copy of MC's Flint and Steel, but something a bit more realistic and original as "Flint and Tinder", combining flint and sticks.

Yes, I know MT doesn't aim to be realistic, but requiring steel ingots to make a fire while a furnace requires only a log of wood and 8 cobblestone isn't realistic either. So you can make a fire in a furnace with a solid piece of wood, but not by rubbing sticks together? :)

kilbith:
_"The copy of MC"_ has never been an argument for me. Lots of stuff in Minetest Game are directly inspired from MC (deny it or not). Look at the moon, not the finger pointing it.

Producing fire from clashing flint and steel is realistic already. Also the Gambit's artwork is really good -- making a new texture for the tinder would be of inferior quality and I'm not going to make that.

sofar:
A use sound might be nice.

kilbith:
> A use sound might be nice.

I'm open to suggestions.

paramat:
Although it may be a little illogical i like it so :+1: 
Sound would be nice but not essential for now.

PilzAdam:
Doesn't this make it too easy to get fire? I always liked that you need to get lava first, which is very dangerous and not immediately available.

kilbith:
From what I see on servers, players **need** to place fire into fireplaces or as other decorations. You can't place it with lava or painfully.

Ekdohibs:
:+1:

PilzAdam:
Maybe just craft fire from flint and steel? Otherwise you can get relatively much fire from little resources.

kilbith:
No thanks, this is not craftable and that doesn't sound right as explained above.

> Fire burns, wielding it in hand does not make sense and looks odd.

... and I'd prefer an action onto the world to make fire. That's more fun.

C1ffisme:
@PilzAdam I'm not sure why you are against having a quick way to make fire. It's not as if it does anything game-breaking at the moment.

PilzAdam:
@C1ffisme having everything easily available is boring gameplay, isn't it? Also, I'm not objecting this, but I want people to think first about gameplay implications and possible other solutions to the problem before adding whatever they thought of first.

C1ffisme:
@PilzAdam Yes, but fire is not a large milestone, unlike Minecraft where it is required to make the Nether Portal. (You could use lava, but that's not practical.)

sofar:
Maybe reduce uses dramatically? e.g. 5 - 10 uses only?

C1ffisme:
@sofar That would mean I can get rid of excess flint easily.

(We need to consider that too, flint doesn't have a big use other than this flint and steel.)

Although, we do need to consider the use of a steel ingot. That means there is 10 uses for every steel ingot. Not a lot for people who don't use it every day, but if you need to make a lot of torch things...
screenshot_20160312_144750
(Imagine making a road 500 blocks long using these as light sources.)

sofar:
This patch puts down a basic_flame, that will automatically die after a while, so it's not usable to make permanent lighting.

C1ffisme:
@sofar Hmm... I'm not sure why someone would want non-permanent fire for anything except maybe getting rid of excess wood (very inefficiently).

Maybe make a flame placed on coal blocks be a perma-flame?

sofar:
I wouldn't make it perma, but I would make it 100+ game days long. Could be nodetimer based, as well.

paramat:
Delaying merge because discussion ongoing.

C1ffisme:
@sofar Hmm... If it's that long, than why not keep it permanent anyway?

sfan5:
>Giving the possibility to place fire in survival mode.

Can't believe we were missing this. :+1:

paramat:
28884cc784626544871bd28032130899393c74e9

sofar:
@C1ffisme could be a fun RP element - turning lights back on in places you haven't visited in a while.

C1ffisme:
@sofar Hmmm... I see now.

PR #912, opened 2016-03-12
added option to disable bones

tenplus1:
Added the option to disable bone drops so that players can use alternative mods like DropOnDie instead.
[updated version of #896 , also see #203  ]


PilzAdam:
If the use-case is that mods replace this functionality then there should be a way for mods to override the functions of the bones mod, not a setting for the player.

tenplus1:
You mean add a global variable that can be set inside such mods that Bones will check before dropping inventory inside a bone node ?

PR #914, opened 2016-03-13
Option to disable bone drops

tenplus1:
I've added the option to disable bone drops by adding 'bones.enable = true' which can easily be set to false inside other mods like DropOnDie.  Also some code cleanup.


tenplus1:
Changed my DropOnDie mod to use new flag: https://github.com/tenplus1/dropondie/blob/master/init.lua

PR #918, opened 2016-03-13
Allow for shift+clicking inventory into trash

ghost:


Fixer-007:
Nice, I was thinking about the same but never opened the issue. :+1:

C1ffisme:
Pretty good to have this, makes building be quicker.

kilbith:
Approved (as the author of the inventory).

everamzah:
Do you want enter key inputting search queries, too?
https://www.youtube.com/watch?v=gX-fNU_TNco
https://github.com/everamzah/creative

DonBatman:
Please add enter for search queries

kilbith:
> Do you want enter key inputting search queries, too?

Your code is utterly hacky, so no.

It's not clean to latently call a formspec inside `fields.quit`. Also, it redirects to a regular player's formspec, not a detached inventory. Big :-1:

everamzah:
My thoughts exactly.  This PR simply adds the listring[], which is super handy.

kilbith:
> Except for the redirection part which I don't understand, but that's another topic.

`minetest.show_formspec` is server-side and subject to lag. A detached inventory is not tied to anything (player or node) and thus not subject to lag. The logic you made is leaving the detached inventory and redirect to a player's formspec, which is dirty. The engine should be modified instead of hacking around. But yes, another topic...

kilbith:
Integrated in https://github.com/minetest/minetest_game/pull/945, as it's very minor.

Feel free to close your PR.

PR #927, opened 2016-03-14
added missing alias for default:ladder

verymilan:
there was an alias missing: `default:ladder_wood -> default:ladder`


kilbith:
No, the wooden ladder has changed of name since https://github.com/minetest/minetest_game/commit/7f317871d48bd261faa16bbe2cfc87e551d2ae6e

tchncs:
Oh, thanks for the info - since the latest fence changes are incompatible with the current client releasem, i created a fusion of the current state with my changes and without the fence changes...looks like i mixed up something :)

PR #946, opened 2016-03-17
Xpanes: convert to connected nodeboxes.

sofar:
I've rewritten this to use connected nodeboxes, but with a caveat. In
order to make flat nodes look better, I'm keeping one non-connected
pane that is flat around to convert flat sections to the flat nodes
instead of connected, as these look better and are easier to work
with. Once more sides are needed we convert the panes on the fly to
connected nodes and recalculate the shape.

We don't paint any of the half-panes that the previous generation of
xpanes did. There's no need and it's harder to work with. Updating the
nodes also seems more natural and placement and removal works straight
forward.

This is _missing_ any compat/conversion code, so it's not yet complete.
I expect to use an LBM to perform the conversion, which is the right
tool to do so.

Check out https://youtu.be/13M6Hz8oQ1I for how it looks.
- [x] - conversion code.

### DIFF
--- a/mods/xpanes/init.lua
+++ b/mods/xpanes/init.lua
@@ -1,156 +1,146 @@
-xpanes = {}
 
-local function rshift(x, by)
-	return math.floor(x / 2 ^ by)
+local function is_pane(pos)
+	return minetest.get_item_group(minetest.get_node(pos).name, "pane") > 0
 end
 
-local directions = {
-	{x = 1, y = 0, z = 0},
-	{x = 0, y = 0, z = 1},
-	{x = -1, y = 0, z = 0},
-	{x = 0, y = 0, z = -1},
-}
+local function connects_dir(pos, name, dir)
+	local aside = vector.add(pos, minetest.facedir_to_dir(dir))
+	if is_pane(aside) then
+		return true
+	end
 
-local function update_pane(pos, name)
-	if not minetest.get_node(pos).name:find("^xpanes:"..name) then
+	local connects_to = minetest.registered_nodes[name].connects_to
+	if not connects_to then
+		return false
+	end
+	local list = minetest.find_nodes_in_area(aside, aside, connects_to)
+
+	if #list > 0 then
+		return true
+	end
+
+	return false
+end
+
+local function swap(pos, node, name, param2)
+	if node.name == name and node.param2 == param2 then
 		return
 	end
-	local sum = 0
-	for i, dir in pairs(directions) do
-		local node = minetest.get_node(vector.add(pos, dir))
-		local def = minetest.registered_nodes[node.name]
-		local pane_num = def and def.groups.pane or 0
-		if pane_num > 0 or not def or (def.walkable ~= false and
-				def.drawtype ~= "nodebox") then
-			sum = sum + 2 ^ (i - 1)
+
+	minetest.set_node(pos, {name = name, param2 = param2})
+end
+
+local function update_pane(pos)
+	if not is_pane(pos) then
+		return
+	end
+	local node = minetest.get_node(pos)
+	local name = node.name
+	if name:sub(-5) == "_flat" then
+		name = name:sub(1, -6)
+	end
+
+	local any = node.param2
+	local c = {}
+	local count = 0
+	for dir = 0, 3 do
+		c[dir] = connects_dir(pos, name, dir)
+		if c[dir] then
+			any = dir
+			count = count + 1
 		end
 	end
-	if sum == 0 then
-		sum = 15
-	end
-	minetest.set_node(pos, {name = "xpanes:"..name.."_"..sum})
-end
 
-local function update_nearby(pos, node)
-	node = node or minetest.get_node(pos)
-	local name = node.name
-	if not name or node.name:sub(1, 7) ~= "xpanes:" then
-		return
-	end
-	local underscore_pos = string.find(name, "_[^_]*$") or 0
-	local len = name:len()
-	local num = tonumber(name:sub(underscore_pos+1, len))
-	if not num or num < 1 or num > 15 then
-		name = name:sub(8)
+	if count == 0 then
+		swap(pos, node, name .. "_flat", any)
+	elseif count == 1 then
+		swap(pos, node, name .. "_flat", (any + 1) % 4)
+	elseif count == 2 then
+		if (c[0] and c[2]) or (c[1] and c[3]) then
+			swap(pos, node, name .. "_flat", (any + 1) % 4)
+		else
+			swap(pos, node, name, 0)
+		end
 	else
-		name = name:sub(8, underscore_pos - 1)
-	end
-	for i, dir in pairs(directions) do
-		update_pane(vector.add(pos, dir), name)
+		swap(pos, node, name, 0)
 	end
 end
 
-minetest.register_on_placenode(update_nearby)
-minetest.register_on_dignode(update_nearby)
+minetest.register_on_placenode(function(pos, node)
+	if minetest.get_item_group(node, "pane") then
+		update_pane(pos)
+	end
+	for i = 0, 3 do
+		local dir = minetest.facedir_to_dir(i)
+		update_pane(vector.add(pos, dir))
+	end
+end)
 
-local half_boxes = {
-	{0,     -0.5, -1/32, 0.5,  0.5, 1/32},
-	{-1/32, -0.5, 0,     1/32, 0.5, 0.5},
-	{-0.5,  -0.5, -1/32, 0,    0.5, 1/32},
-	{-1/32, -0.5, -0.5,  1/32, 0.5, 0}
-}
+minetest.register_on_dignode(function(pos)
+	for i = 0, 3 do
+		local dir = minetest.facedir_to_dir(i)
+		update_pane(vector.add(pos, dir))
+	end
+end)
 
-local full_boxes = {
-	{-0.5,  -0.5, -1/32, 0.5,  0.5, 1/32},
-	{-1/32, -0.5, -0.5,  1/32, 0.5, 0.5}
-}
-
-local sb_half_boxes = {
-	{0,     -0.5, -0.06, 0.5,  0.5, 0.06},
-	{-0.06, -0.5, 0,     0.06, 0.5, 0.5},
-	{-0.5,  -0.5, -0.06, 0,    0.5, 0.06},
-	{-0.06, -0.5, -0.5,  0.06, 0.5, 0}
-}
-
-local sb_full_boxes = {
-	{-0.5,  -0.5, -0.06, 0.5,  0.5, 0.06},
-	{-0.06, -0.5, -0.5,  0.06, 0.5, 0.5}
-}
-
-local pane_def_fields = {
-	drawtype = "airlike",
-	paramtype = "light",
-	is_ground_content = false,
-	sunlight_propagates = true,
-	walkable = false,
-	pointable = false,
-	diggable = false,
-	buildable_to = true,
-	air_equivalent = true,
-}
-
+xpanes = {}
 function xpanes.register_pane(name, def)
 	for i = 1, 15 do
-		local need = {}
-		local cnt = 0
-		for j = 1, 4 do
-			if rshift(i, j - 1) % 2 == 1 then
-				need[j] = true
-				cnt = cnt + 1
-			end
-		end
-		local take = {}
-		local take2 = {}
-		if need[1] == true and need[3] == true then
-			need[1] = nil
-			need[3] = nil
-			table.insert(take, full_boxes[1])
-			table.insert(take2, sb_full_boxes[1])
-		end
-		if need[2] == true and need[4] == true then
-			need[2] = nil
-			need[4] = nil
-			table.insert(take, full_boxes[2])
-			table.insert(take2, sb_full_boxes[2])
-		end
-		for k in pairs(need) do
-			table.insert(take, half_boxes[k])
-			table.insert(take2, sb_half_boxes[k])
-		end
-		local texture = def.textures[1]
-		if cnt == 1 then
-			texture = def.textures[1].."^"..def.textures[2]
-		end
-		minetest.register_node(":xpanes:"..name.."_"..i, {
-			drawtype = "nodebox",
-			tiles = {def.textures[3], def.textures[3], texture},
-			paramtype = "light",
-			groups = def.groups,
-			drop = "xpanes:"..name,
-			sounds = def.sounds,
-			node_box = {
-				type = "fixed",
-				fixed = take
-			},
-			selection_box = {
-				type = "fixed",
-				fixed = take2
-			}
-		})
+		minetest.register_alias("xpanes:" .. name .. "_" .. i, "xpanes:" .. name .. "_flat")
 	end
 
-	for k, v in pairs(pane_def_fields) do
-		def[k] = def[k] or v
-	end
+	local flatgroups = table.copy(def.groups)
+	flatgroups.pane = 1
+	minetest.register_node(":xpanes:" .. name .. "_flat", {
+		description = def.description,
+		drawtype = "nodebox",
+		paramtype = "light",
+		is_ground_content = false,
+		sunlight_propagates = true,
+		inventory_image = def.inventory_image,
+		wield_image = def.wield_image,
+		paramtype2 = "facedir",
+		tiles = {def.textures[3], def.textures[3], def.textures[1]},
+		groups = flatgroups,
+		drop = "xpanes:" .. name .. "_flat",
+		sounds = def.sounds,
+		node_box = {
+			type = "fixed",
+			fixed = {{-1/2, -1/2, -1/32, 1/2, 1/2, 1/32}},
+		},
+		selection_box = {
+			type = "fixed",
+			fixed = {{-1/2, -1/2, -1/32, 1/2, 1/2, 1/32}},
+		},
+		connect_sides = { "left", "right" },
+	})
 
-	def.on_construct = function(pos)
-		update_pane(pos, name)
-	end
-
-	minetest.register_node(":xpanes:"..name, def)
+	local groups = table.copy(def.groups)
+	groups.pane = 1
+	groups.not_in_creative_inventory = 1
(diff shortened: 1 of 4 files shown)


Rui914:
```
local nodenames = {}
for i = 1, 15 do
	nodenames[#nodenames + 1] = "xpanes:" .. name .. "_" .. i
end

minetest.register_lbm({
	name = "xpanes:convert_xpanes_to_connected_nodeboxes_" .. name,
	nodenames = nodenames ,
	action = function(pos)
		minetest.set_node(pos, {name = "xpanes:" .. name"})
		for i = 0, 3 do
			local dir = minetest.facedir_to_dir(i)
			update_pane(vector.add(pos, dir))
		end
	end
})
```

mods/xpanes/init.lua
```diff
@@ -1,156 +1,148 @@
...
+			type = "fixed",
+			fixed = {{-1/2, -1/2, -1/32, 1/2, 1/2, 1/32}},
+		},
+		connect_sides = { "left", "right" },
+	})
+
+	local groups = table.copy(def.groups)
+	groups.pane = 1
+	groups.not_in_creative_inventory = 1
+	minetest.register_node("xpanes:" .. name, {
```
Rui914: "xpanes:" > ":xpanes:"

kilbith:
Good job. 2 remarks : 
- The current crossed-panes (the node you place at first without connection) is oftenly used for various purposes like table legs or thing for suspending blocks. Your initial node is just a simple pane.
- At [1:55](https://youtu.be/13M6Hz8oQ1I?t=115) of your video, the selection box is small and quite hard to point when the pane is crossed.

But that shouldn't block this PR as the benefit is very interesting (15 nodes less for each pane).

mods/xpanes/init.lua
sofar: thanks, I'll get that.

sofar:
@Rui914 thanks, that indeed looks right, although I wonder if we can make it a single lbm that is used for all derivatives.

sofar:
@kilbith the selection box being small is something I think I can fix in core.

Fixer-007:
Nice! I will look at them.

Rui914:
```lua
local nodenames = {}

function xpanes.register_pane(name, def)
	for i = 1, 15 do
		nodenames[#nodenames + 1] = "xpanes:" .. name .. "_" .. i
	end
	...
end

minetest.after(0, function()
	minetest.register_lbm({
		name = "xpanes:replace_legacy_panes",
		nodenames = nodenames,
		action = function(pos, node)
			local n = node.name:match("xpanes:.+_(%d+)")
			node.name = node.name:sub(1, -(#n + 1))
			minetest.set_node(pos, node)
		end
	})
end)
```
Too hackish and operation is unconfirmed.

sofar:
I have a better idea, give me a sec.

sofar:
@rui914's code was close, but no cigar :)

Problem is that lbm's can't run in ```minetest.after(0``` context, which is correct.

However, we can easily use aliases to tell core that all the old nodes are actually flat panes and belong to ```group:pane```, and then the lbm becomes trivial, we just call ```update_pane()``` on all panes and around all panes.

Tested conversion from master to xpanes several times, good results.

sofar:
Oh and this conversion will also work excellent for mods using the ```register_pane()``` API.

kilbith:
@paramat Would it be wise to mark it as milestone ? As it's a feature and the freeze happens tomorrow...

sofar:
I don't consider this WIP anymore. Especially with the selection boxes fixed, this is a nice reduction of nodedef count, and the code isn't any more complex.

sofar:
I'm taking a closer look at the texture usage in this patch. It could be improved.

sofar:
Fixed up the texture usage and docs.

Ferk:
Did this not make it for the 0.4.14?

mods/xpanes/init.lua
```diff
@@ -1,156 +1,147 @@
...
+		sounds = def.sounds,
+		node_box = {
+			type = "connected",
+			fixed = {{-1/31, -1/2, -1/31, 1/31, 1/2, 1/31}},
+			connect_front = {{-1/31, -1/2, -1/2, 1/31, 1/2, -1/31}},
+			connect_left = {{-1/2, -1/2, -1/31, -1/31, 1/2, 1/31, 1/31}},
+			connect_back = {{-1/31, -1/2, 1/31, 1/31, 1/2, 1/2}},
+			connect_right = {{1/31, -1/2, -1/31, 1/2, 1/2, 1/31}},
+		},
+		connects_to = {"group:pane", "group:stone", "group:glass", "group:wood", "group:tree"},
```
ShadowNinja: This should actually connect to anything walkable, and I think that type of connection is pretty common, so it should have core support.

ShadowNinja:
I don't like how you've got a hybrid now where server-side updates are used sometimes and client-side ones other times.  A `connect_none` or similar option should probably be added for this type of use case.

sofar:
We don't implement facedir rotation for connected nodeboxes, so I felt this would become a kludge quickly. Yes, it's one extra node, but the benefit is that we never have half panes, which I think is visually much more appealing.

ShadowNinja:
Whether or not half-panes are desirable is debatable...

sofar:
The flat pane itself makes it "connect" to stuff that's not in the explicit connect_to list. That means that I can place it in between two furnaces and it still looks connected:



You can't do this with the half-panes.

ShadowNinja:
@sofar: But then it breaks if you try to use more than one pane between the nodes.  That's unexpected behavior.

sofar:
@shadowninja, that's not what happens, for example:

Ferk:
@sofar I think he means in the X axis, not Y.

imho, the main issue is that we lose the + shaped node that some people used for tables and decorations, like kilbith said. So some maps or schematics might be a bit broken with this.

The lack of facedir rotation for a "+" node wouldn't be a problem. ShadowNinja's idea of a connect_none property looks interesting. It would simplify it a lot and you'll still get it to connect to stuff whenever it's actually supposed to connect_to.

mods/xpanes/init.lua
Ferk: Imho, it should be possible to override it in the definition. Using something like this:

     connects_to = def.connects_to or {"group:pane", "group:stone", "group:glass", "group:wood", "group:tree"}

sofar:
> I think he means in the X axis, not Y.

? it works the same in all axes, X, Z. I just stacked them to make the screenshot smaller, but Y is irrelevant.

Ferk:
When spreading the panes in parallel there's necessarily gotta be a difference between connected_to nodes or not. Thought on second read, I doubt that's what ShadowNinja meant... my bad

lightonflux:
Will this be merged before the freeze?

sofar:
No rebase needed, but it needs 1-2 devs to actually use this code a bit.

The behavior is very different, but I think it's intuitive for players and understandable, and will lead to easier building than the old panes.

@nore, @paramat, others, please try this in-game.

norey:
@sofar why are you tagging me? How am I related to this project?

sofar:
@norey apologies, someone else is using several handles of which one is one letter off from yours, so github autocompleted it to yours. My bad.

norey:
@sofar don't worry about it

sofar:
Needs review - if this is OK then I can make more time addressing issues if they are found, but there are people who want to add more panes who are blocked/have conflicts with this PR.

Fixer-007:
What I will miss a lot are glass half-panes :/ that can create nice effect of glass entrance.

paramat:
Man, i need to review ths.

paramat:
@rubenwardy if possible can you look at this? Seems good work and deserves attention. I'll try to review also.

paramat:
Watched the video and the behaviour looked good.
I won't miss the half-panes, the loss of the cross shape might annoy some, but the state of xpanes is not good and i feel a significant rework is welcome.
Current xpanes update 'on construct' which i guess is not called for placement by voxelmanip or schematic? I guess xpanes will not rotate if a schematic is rotated?

mods/xpanes/init.lua
```diff
@@ -1,156 +1,147 @@
...
-			sum = sum + 2 ^ (i - 1)
-		end
+
+	local connects_to = minetest.registered_nodes[name].connects_to
+	if not connects_to then
+		return false
 	end
-	if sum == 0 then
-		sum = 15
+	local asidenode = minetest.get_node(aside)
```
paramat: Seems unused in this function?

mods/xpanes/init.lua
paramat: I guess since the connects_to list has to be a list of specified nodes it has to be kept reasonably short and it's not possible to check for walkable.
A settable 'connects_to' could be good though?

paramat:
Reviewed and code seems ok. Also added a line comment and replied to a previous line cmment.
Overall i like this.

mods/xpanes/init.lua
```diff
@@ -1,156 +1,147 @@
...
+		is_ground_content = false,
+		sunlight_propagates = true,
+		description = def.description,
+		tiles = {def.textures[3], def.textures[3], def.textures[1]},
+		groups = groups,
+		drop = "xpanes:" .. name .. "_flat",
+		sounds = def.sounds,
+		node_box = {
+			type = "connected",
+			fixed = {{-1/31, -1/2, -1/31, 1/31, 1/2, 1/31}},
```
paramat: Just wondering why 1/31? To avoid z-fighting with 1/32?

mods/xpanes/init.lua
sofar: Oh IC, yes, this was needed before I optimized this code to use `find_nodes_in_area()`. I'll remove it.

mods/xpanes/init.lua
sofar: The pane ends up 1/16 (1 px) thick this way, and centered in the node, which has the best visual.

paramat: I must be misunderstanding something, the flat pane and the current xpanes use 1/32.
2 * 1/32 = 1/16.

paramat:
Otherwise +1.

mods/xpanes/init.lua
paramat: 1/2 - 1/32 != 1/31 ?

mods/xpanes/init.lua
```diff
@@ -1,156 +1,146 @@
...
+			connect_front = {{-1/31, -1/2, -1/2, 1/31, 1/2, -1/31}},
+			connect_left = {{-1/2, -1/2, -1/31, -1/31, 1/2, 1/31, 1/31}},
```
paramat: Has 7 entries.

paramat:
^ Just saw an error.

paramat:
The build fail issue is minor.

mods/xpanes/init.lua
t4im: Also these double nested tables `{{}}` are unnecessary and use additional persistent ram. Only a little of course, but they'll just end up copied everywhere.

paramat:
@rubenwardy since this s a major change it would be good to hear your opinion too.

mods/xpanes/init.lua
sofar: Yes, the current xpanes pane is smaller than 1px. I changed this on purpose to 1px thickness, which feels more coherent.

mods/xpanes/init.lua
sofar: oh, this needs fixing.

sofar: turns out the first 6 elements were all OK, so this wouldn't have caused any problems. Removed it otherwise, of course.

sofar:
Removed that erroneous extra array element.

mods/xpanes/init.lua
```diff
@@ -1,156 +1,146 @@
...
-	minetest.set_node(pos, {name = "xpanes:"..name.."_"..sum})
+
+	return false
 end
 
-local function update_nearby(pos, node)
-	node = node or minetest.get_node(pos)
-	local name = node.name
-	if not name or node.name:sub(1, 7) ~= "xpanes:" then
+local function swap(pos, node, name, param2)
```
rubenwardy: this seems confusingly names to me, it's more of a set_node which avoids excess work

paramat:
I think we must be misunderstanding each other.

The current xpanes correctly use 1/32 and 2 * 1/32 = 1/16 = 1 pixel thick:
```
local half_boxes = {
-	{0,     -0.5, -1/32, 0.5,  0.5, 1/32},
-	{-1/32, -0.5, 0,     1/32, 0.5, 0.5},
-	{-0.5,  -0.5, -1/32, 0,    0.5, 1/32},
-	{-1/32, -0.5, -0.5,  1/32, 0.5, 0}
-}
-
-local full_boxes = {
-	{-0.5,  -0.5, -1/32, 0.5,  0.5, 1/32},
-	{-1/32, -0.5, -0.5,  1/32, 0.5, 0.5}
-}
```
Youe flat pane correctly uses 1/32:
```
+		node_box = {
+			type = "fixed",
+			fixed = {{-1/2, -1/2, -1/32, 1/2, 1/2, 1/32}},
+		},
```
However your connected nodebox variant uses 1/31 which is incorrect:
```
		node_box = {
			type = "connected",
			fixed = {{-1/31, -1/2, -1/31, 1/31, 1/2, 1/31}},
			connect_front = {{-1/31, -1/2, -1/2, 1/31, 1/2, -1/31}},
			connect_left = {{-1/2, -1/2, -1/31, -1/31, 1/2, 1/31}},
			connect_back = {{-1/31, -1/2, 1/31, 1/31, 1/2, 1/2}},
			connect_right = {{1/31, -1/2, -1/31, 1/2, 1/2, 1/31}},
},
```
2 * 1/31 != 1/16 these are slightly thinner panes.

So i can't approve this PR yet.

sofar:
oh HAHAHAHA, 31, not 32. Yeah, I totally read *over* that :D

paramat:
Cool will +1 once corrected.

sofar:
<lego batman> *first try*</lego batman>

paramat:
:+1:

mods/xpanes/init.lua
```diff
@@ -1,156 +1,146 @@
...
+			fixed = {{-1/2, -1/2, -1/32, 1/2, 1/2, 1/32}},
+		},
+		selection_box = {
+			type = "fixed",
+			fixed = {{-1/2, -1/2, -1/32, 1/2, 1/2, 1/32}},
+		},
+		connect_sides = { "left", "right" },
+	})
+
+	local groups = table.copy(def.groups)
```
t4im: `groups` is already defined, thus the linter warning.
You could reassign, rename or use two [`do-end` blocks](http://www.lua.org/pil/4.2.html) to solve this.

sofar: False positive: the linter is designed to watch out for "potentially" bad pieces of code, but in this case the reassignment is on purpose.

Fixing false positive static analysis warnings just to fix a warning is somewhat bad practice: the tools are designed to give false positives, forcing people to fix false positives is bad form, and the output of false positives should always be carefully inspected.

So, I'd like to leave it as is.

Ferk: The problem is not the reassignment, but the re-declaration. Just remove the "local".

    groups = table.copy(def.groups)

sofar: My point is that I prefer keeping the `local` here.

Re-using variable names and declaring them again with `local` makes it explicit that I'm creating a new, unused version.

t4im: The linter is pedantic as cobble here, but not actually throwing you a false positive.

You are defining an already defined variable in the same chunk without making it syntactically explicit, that you wanted the previous scope to end.
It's such syntactic discrepancies that the linter is meant to show you, whether they lead to a bug or not.

> Re-using variable names and declaring them again with local makes it explicit that I'm creating a new, unused version.

Lua's `do-end` syntax allows you that. It makes the variable scopes explicit instead of hiding them.

```lua
do
   local var = "why is my inventory full of cobble?"
   print(var)
end -- var isn't defined after this point anymore
```

Maybe you don't want the linter to be that strict to everyone, which is your prerogative, and an understandable decision, but even then leaving it as-is isn't an option either. You'll have to reconfigure the linter to not warn about this particular issue anymore.

Or this will just let every PR fail after this is merged, which is the only reason i'm even pointing this out. It could kill the linter off the project, before it actually warned about something more serious.

sofar: Oh FCOL already.

sofar:
Updated to avoid static analysis warning.

mods/xpanes/init.lua
t4im: > Oh FCOL already.

Did not intend to frustrate you with it.
Rather to prevent the following PR-failings.

```lua
ignore = {"411"} -- `411  │ Redefining a local variable. `
```
Is still an option, as pointed out.

sofar: I personally thought the linter was a great addition, I'm more frustrated that we're spending more time discussing it than coding ;).

t4im: Heh, fair point.

paramat:
Nice solution.

mods/xpanes/init.lua
paramat: Seems reasonable to me.

paramat:
This is not essential but you could add yourself to README code credits and license.txt code license.
Rubenwardy seems ok with this PR so it can probably be merged soon.

paramat:
It would be good to add yourself to README.txt code credits and to license.txt code license, but this could be done in a later commit if you prefer.
Rubenwardy seems fairly happy with this so it could be merged soon.

paramat:
Would you like me to merge this and add credits myself, if you're short on time?

sofar:
@paramat I didn't really care two days back, but I just looked at `git blame init.lua` and noticed that I've basically redid the entire thing :)

I've put myself in both files (`README.txt` and `license.txt`) underneath the old authors, so, thanks for asking ;)

paramat:
Thanks, after the massive job i did on updating credits and licensing we really need to keep on top of that stuff now.

paramat:
54235f32766167bbd55cae5715d6e01754715ce6

PR #967, opened 2016-03-22
Obsidian Glass Panes

ghost:
This adds obsidian glass panes to minetest_game.


kilbith:
Please GTFO and learn git and basic Lua programming. :-1:

MT-Modder:
Yeah...too many unrelated/unneeded changes relative to just adding the new pane. Try to keep pull requests to 1 commit, unless it's such a big change it's worth splitting up, which is rare. Either learn to squish your commits or open a new PR, although the latter isn't really recommended if you know how to squish.

I like the idea though.
Best of luck to you.

C1ffisme:
From a gameplay wise perspective, do we really want obsidian glass panes? I use glass panes for almost everything, and this might make the normal ones almost useless.

MT-Modder:
I think it's mainly personal preference but if they were made inflammable then that would be a plus.

C1ffisme:
> but if they were made inflammable

Are the normal ones flammable? I was not aware of this.

MT-Modder:
Actually I'm not sure, probably not. It could have been a mod I used to use: "fire_plus".

I think they should be though.

DonBatman:
I would rather see glass drop as items.

sofar:
The proper term is to ```squash``` commits.

In this case, however, I would recommend retracting this commit, and redoing it from scratch. The addition of a new xpane type should be a single block of changes 10 lines long or so, not touch code all over the place.

MT-Modder:
@sofar Thanks. You can tell I've never learned to used it. XD

paramat:
@BBmine We can squash commits on merge, but please add proper commit messages.
You might want to wait to see if we merge #946 first, which because of it's quality is higher priority.

BBmine:
@paramat I edited the init.lua and added a proper commit message.

paramat:
#946 is very likely to go ahead. Closing this because the PR will need a complete rework, and due to code issues.

BBmine:
I have learned a lot about git since I started this pull request. I think I'll rework it over Christmas.

PR #969, opened 2016-03-22
Remove `default:dirt_with_grass_footsteps` and add alias.

red-001:
This node look almost identical to dirt with grass and didn't seem to have any use.


paramat:
I think it may be used on some servers.

Calinou:
> I think it may be used on some servers.

xkcd: Workflow

paramat:
If i remember correctly, when i last worked on the grass texture i suggested removing grass with footsteps, and was told it's in use on servers. need more input on this @VanessaE 
Personally i'm neutral about it.

VanessaE:
Some servers still use grass/footsteps (most notably, those which run skyblock).  Keep it.

rubenwardy:
Then skylands can have it in their game. There is no reason to keep it here.

Dragonop:
Is it breaking something? No? Then why would you remove it...

red-001:
@Dragonop It's useless and trivial to add using a mod.

paramat:
:-1: It's in use, aliasing it means those who use it lose their footsteps, or, they would have to go around beforehand replacing all footsteps with a mod version.

red-001:
Could someone name a mod that uses it then?

red-001:
Shouldn't code to make footsteps be added to minetest game then?

paramat:
Maybe it coulld be an option in future, with engine changes to support lightweight changing of nodes to footstep versions. It can be done in Lua but is a little heavy (see my trail mod).
Anyway, note we have default:cloud which is not used in MTGame but we keep it anyway.

paramat:
Nore, sfan5 and i agree to close.

PR #983, opened 2016-03-28
Disabling tnt only disables tnt:tnt node and craft

tenplus1:
When enable_tnt = false then tnt:tnt node and it's craft are both disabled while tnt:tnt_burning and tnt:boom are available for other mods that require them but not in creative inventory.

### DIFF
--- a/mods/tnt/init.lua
+++ b/mods/tnt/init.lua
@@ -4,7 +4,7 @@ local singleplayer = minetest.is_singleplayer()
 local setting = minetest.setting_getbool("enable_tnt")
 if (not singleplayer and setting ~= true) or
 		(singleplayer and setting == false) then
-	return
+	setting = nil -- disabled tnt:tnt and craft recipe further on
 end
 
 -- loss probabilities array (one in X will be lost)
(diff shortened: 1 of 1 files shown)


mods/tnt/init.lua
```diff
@@ -256,6 +268,7 @@ minetest.register_node("tnt:tnt_burning", {
 		"tnt_bottom.png", "tnt_side.png"},
 	light_source = 5,
 	drop = "",
+	groups = {not_in_crafting_guide = 1},
```
PilzAdam: This is not a "standard" group from minetest_game; don't add this.

sofar:
I'm not agreeing here. If there's a need to remove ```tnt:tnt``` as item from the game, but keep all the API functions present, then that isn't "disabling" the mod at all, but something differently, and therefore currently not supported.

This patch changes what ```enable_tnt``` does in a way that I don't think should be the case.

tenplus1:
Pilzadam: Please explain the not standard group ?!?!

Sofar: If mods are written that need to spawn in tnt:tnt_burning for explosions then they cannot if enable_tnt is set to false since it registers nothing in the tnt mod.  This change STILL disables tnt but in a different way while allowing mods to place exploding tnt blocks.

PilzAdam:
@tenplus1 `not_in_crafting_guide` isn't used anywhere else in minetest_game. Adding this to only some random nodes is inconsistent.

DonBatman:
not_in_crafting_guide should be not_in_creative_inventory

PilzAdam:
@DonBatman this is not needed; only items that have a description are added to the creative inventory. So only items with descriptions need the `not_in_creative_inventory` group.

tenplus1:
Ouch, I typed the wrong group...  I'm going to close this pull as Sofar's TNT changes are more important just now.

PR #984, opened 2016-03-29
Add key tool for doors to switch between normal, owned and protected

tenplus1:
Rather than creating a normal door, and then a locked door with the same look etc. adding this key tool will allow mod makers to add only one door and switch between normal, owned and protected use only, with the key having 50 uses before breaking.  Backwards compatible with current doors already on map.


kilbith:
- Some style issues.
- Copypastas in your code (the metadata setters particularly), could be abstracted to a function.
- PixelBOX's key texture is better : default_key

Add.: this feature was planned by @sofar.

sofar:
I'm really not a fan of this PR.

- It introduces a fake player named "prot"
- The concept of "a protected" door isn't something in known minetest_game. There only are "owned" doors and normal doors
- The use of a "key" item to modify the door doesn't make sense. keys lock or unlock doors, this does neither

I'm open to figuring out if we can modify the doors API to better integrate with protection mods.

As for my own "keys" PR, I'll have to respin it one of these days for post-0.4.14. The reason I personally still favor that approach is that it creates an interface that can be re-used by mods and works for locked trapdoors, chests and doors alike. It wouldn't actually conflict with this PR, other than creating a set of key items as well.

tenplus1:
A fake player isnt created at all, it uses current player names and 'prot' is only the string that holds the name.

mods/doors/init.lua
```diff
@@ -394,7 +399,6 @@ function doors.register(name, def)
 		paramtype = "light",
 		paramtype2 = "facedir",
 		sunlight_propagates = true,
-		use_texture_alpha = true,
```
sofar: Unrelated, but important change. This should go in a separate PR :)

sofar:
Please split out the transparency fix, and explain why it's needed. That can go into 0.4.14, likely, still.

mods/doors/init.lua
tenplus1: have already removed those lines in the 2nd pull.

C1ffisme:
What about instead of a key, a lock? This lock would only have one use. Once you click on a door with it, the door becomes owned. If the owner holds shift and right clicks on it, it becomes a protected door. If the owner shift-clicks again, it becomes a normal door and a lock is given to the owner.

(Anyway, it makes a bit more sense then a key)

Thomas--S:
:+1: for adding a protected door
:+1: for the lock idea

mods/doors/init.lua
sofar: yes, but, in a separate pull request please.

sofar:
This will be obsolete by the planned ```is_protected()``` overhaul that est31 and me would like to work on post-0.4.14. The key item itself can then be a mod thing and not require any changes to the door API.

PilzAdam:
> The concept of "a protected" door isn't something in known minetest_game. There only are "owned" doors and normal doors

>[...]

>I'm open to figuring out if we can modify the doors API to better integrate with protection mods.

^ This. :-1: to this PR.

DonBatman:
:+1: 
This would be great to have

sofar:
Related: https://github.com/minetest/minetest/issues/3833

0-afflatus:
What is required is shared lockable items, particularly chests and doors - I don't know if this is what tenplus1 means by "protected". Currently I'm looking to convert Sokomine's locks mod, but it could really use an implementation closer to core. This is probably OT for this PR, but if we're going to introduce keys, that is what they should allow.

tenplus1:
PilzAdam: This feature will introduce the ability to protect doors now that more players and servers are using protection mods against griefers and the like.  If no protection mod is active then a door with the protection added opens and closes like a normal door giving backwards compatibility in the world.

C1ffisme:
@tenplus1 Rather than just having a 'protected' door available for singleplayer newbies to ponder on, why not disable the ability to place a protected door unless someone uses a function in their mod?

The function could be something like `doors.enable_protected()`.

Thomas--S:
I really like the idea of @C1ffisme. This would also solve the backwards compatibility problem @sofar mentioned in IRC.

tenplus1:
As the pull stands it has full backwards compatibility with the ability to add a protected door for when a protection mod is added.  If I add a new function to check for protection mods then it would inflate the code and still have lines to check for protection already, annoying devs who don't want protection in there at all.  I am simply trying to add a well needed feature to the doors mod with the ability to add protected doors if and when required.

PR #999, opened 2016-04-01
Dirt should turn to dry grass in some occasions.

sofar:
If dirt is near dirt_with_dry_grass, or sand, we don't change it
to green grass but to dirt_with_dry_grass.

### DIFF
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -384,7 +384,11 @@ minetest.register_abm({
 --
 
 minetest.register_abm({
-	nodenames = {"default:dirt_with_grass", "default:dirt_with_dry_grass"},
+	nodenames = {
+		"default:dirt_with_grass",
+		"default:dirt_with_dry_grass",
+		"default:dirt_with_snow",
+	},
 	interval = 8,
 	chance = 50,
 	catch_up = false,
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -353,31 +353,68 @@ minetest.register_abm({
 
 
 --
--- Grass growing on well-lit dirt
+-- Convert dirt to something that fits the environment
 --
 
 minetest.register_abm({
 	nodenames = {"default:dirt"},
-	neighbors = {"air"},
+	neighbors = {
+		"default:dirt_with_grass",
+		"default:dirt_with_dry_grass",
+		"default:dirt_with_snow",
+		"default:grass_1",
+		"default:grass_2",
+		"default:grass_3",
+		"default:grass_4",
+		"default:grass_5",
+		"default:dry_grass_1",
+		"default:dry_grass_2",
+		"default:dry_grass_3",
+		"default:dry_grass_4",
+		"default:dry_grass_5",
+		"default:snow",
+	},
 	interval = 6,
 	chance = 67,
 	catch_up = false,
 	action = function(pos, node)
+		-- Most likely case, half the time it's too dark for this.
 		local above = {x = pos.x, y = pos.y + 1, z = pos.z}
-		local name = minetest.get_node(above).name
-		local nodedef = minetest.registered_nodes[name]
-		if nodedef and (nodedef.sunlight_propagates or nodedef.paramtype == "light") and
-				nodedef.liquidtype == "none" and
-				(minetest.get_node_light(above) or 0) >= 13 then
-			if name == "default:snow" or name == "default:snowblock" then
-				minetest.set_node(pos, {name = "default:dirt_with_snow"})
-			else
-				minetest.set_node(pos, {name = "default:dirt_with_grass"})
+		if (minetest.get_node_light(above) or 0) < 13 then
+			return
+		end
+
+		-- Look for likely neighbors.
+		local p2 = minetest.find_node_near(pos, 1, {"default:dirt_with_grass",
+				"default:dirt_with_dry_grass", "default:dirt_with_snow"})
+		if p2 then
+			-- But the node needs to be under air in this case.
+			local n2 = minetest.get_node(above)
+			if n2 and n2.name == "air" then
+				local n3 = minetest.get_node(p2)
+				minetest.set_node(pos, {name = n3.name})
+				return
 			end
+		end
+
+		-- Anything on top?
+		local n2 = minetest.get_node(above)
+		if not n2 then
+			return
+		end
+
+		local name = n2.name
+		-- Snow check is cheapest, so comes first.
+		if name == "default:snow" then
+			minetest.set_node(pos, {name = "default:dirt_with_snow"})
+		-- Most likely case first.
+		elseif name:sub(1, 13) == "default:grass" then
+			minetest.set_node(pos, {name = "default:dirt_with_grass"})
+		elseif name:sub(1, 17) == "default:dry_grass" then
+			minetest.set_node(pos, {name = "default:dirt_with_dry_grass"})
 		end
 	end
 })
-
 
 --
 -- Grass and dry grass removed in darkness


sofar:
Fixes #998

rubenwardy:
I'd prefer:

```lua
			elseif minetest.find_node_near(pos, 1, {
					"default:dirt_with_dry_grass",
					"group:sand"}) then
				minetest.set_node(pos, {name = "default:dirt_with_dry_grass"})
```

but doesn't matter that much

paramat:
Not sure about sand though, as we expect green grass next to beaches in many biomes, this could make dry grass spread from a beach across a gren biome, simpler too.

mods/default/functions.lua
```diff
@@ -371,6 +371,11 @@ minetest.register_abm({
 				(minetest.get_node_light(above) or 0) >= 13 then
 			if name == "default:snow" or name == "default:snowblock" then
 				minetest.set_node(pos, {name = "default:dirt_with_snow"})
+			elseif minetest.find_node_near(pos, 1, {
+				"default:dirt_with_dry_grass",
+				"group:sand"
```
paramat: 2 lines above: 2 more indents

mods/default/functions.lua
```diff
@@ -371,6 +371,11 @@ minetest.register_abm({
...
+				}) then
```
paramat: 1 more indent

sofar:
Fixed indentation and removed ```group:sand```.

paramat:
:+1:

sofar:
Changed it even further: grass covering will only grow near existing cover, never in the middle. This makes the dirt cover always match the surroundings.

paramat:
Nice, will be fun to test with an accelerated ABM.

C1ffisme:
I know I've made Minecraft/Minetest comparisons about a million times, but:

In Minecraft, rather than grass growing on dirt anywhere, it requires grass to be near it. This lets you be able to grow say, Mycelium on the dirt rather than grass.

yyt16384:
@C1ffisme In Minetest you can't obtain grass nodes, so it will be really hard to place grass if it works in this way.

C1ffisme:
@yyt16384 True... Maybe we should allow the combining of grass plants and dirt to make grass blocks? (And dry grass + dirt = dry grass?)

Could make landscaping take less time.

sofar:
>@yyt16384 True... Maybe we should allow the combining of grass plants
>and dirt to make grass blocks? (And dry grass + dirt = dry grass?)
>
>Could make landscaping take less time.

Go submit the two recipes for this... I'd consider them. They sound reasonable.

C1ffisme:
What kind of recipe should I use?

- 1 Grass + 1 Dirt = 1 Grassy Dirt
- 4 Grass + 1 Dirt = 1 Grassy Dirt
- 9 Grass + 1 Dirt = 1 Grassy Dirt

sofar:
just 1 grass and 1 dirt. No need to make it expensive.

paramat:
:+1: For this PR.

C1ffisme:
@sofar Here is my pull request:  #1007

I might want to make the recipes shapeless, though.

paramat:
sofar description needs updating.
Please also can you add removal of dirt_with_snow in darkness by adding it here https://github.com/minetest/minetest_game/blob/master/mods/default/functions.lua#L387

sofar:
@paramat I don't see why you would add that for a snow cover. Grass will die without light, but snow is already dead, and in dark corners, where the sun can not reach it, snow tends to stick around longer.

paramat:
'Dirt-with-snow' is actually dirt with grass and snow, adding snow to dirt-with-grass turns it into dirt-with-snow.
That ABM acts only in complete darkness, for example under solid nodes, dirt-with-snow would never be placed in the darkest caves anyway.

paramat:
Inside this grass-spreadng ABM there could be checks for long grass or long dry grass above the dirt node, for speed these extra checks could occur only if the spreading check fails to find anything to spread.

sofar:
I've added a second commit to change dirt to dirt_with_[dry_)grass if we have dirt that's under a (dry) grass plant. It uses an ABM with the neighbor feature, so it should be relatively lightweight.

sofar:
will take care of the dirt_with_snowy_grass thing a bit later and fold that in as well.

sofar:
Added removal of dirt_with_snow under e.g. other dirt.

C1ffisme:
I recently had a different idea:

This assumes that grass is actually wheat, since it drops wheat seeds. (It doesn't make a lot of sense for it to just be holding wheat seeds that blew from wind, since wind doesn't exist in minetest, nor is there wheat plants anywhere else in the world.)

So my idea is, you could plant wheat seeds on grass, and it would eventually grow into a dirt with grass block.

Expensive? Maybe a bit... I bet tweaking drop numbers could help with this...

What about dry grass? Well, if grass is grown in these conditions:

- No dirt_with_grass blocks within X blocks
- No water within X blocks
- Desert Sand or dirt_with_dry_grass within X blocks

Then it will change into dirt_with_dry_grass.

Thoughts?

sofar:
@C1ffisme 

> Thoughts?

yes, your post is confusing, and I think not pertinent to this PR.

C1ffisme:
> yes, your post is confusing,
 
Oops. I said grass instead of dirt. The basic idea is that you can plant wheat seeds on dirt to get grass, since grass drops wheat seeds, therefore grass is (possibly) wheat.

> and I think not pertinent to this PR.

This is an alternative to the grass + dirt = grassy dirt recipe.

sofar:
@C1ffisme Ah, ok, that would be relevant.

However, I think that would be highly confusing for players.

I tend to think of it as "finding wild wheat seeds in the grass" when digging grass plants. Grass isn't wheat, there's just a stray wheat plant in it sometimes.

C1ffisme:
@sofar It's still a bit strange. It's sort of like the "What came first, the Chicken or the Egg?" question. Did seeds just appear, or blow on the wind from wheat?

I think maybe a "wheat field" biome could settle this question if not my previous "grass = badly fertilized wheat" solution. The biome would have to be pretty rare, though. Since we wouldn't want players to get food too quickly. (At least, until hunger/stamina gets added.)

paramat:
> you can plant wheat seeds on dirt to get grass, since grass drops wheat seeds

Nope.

ABM looks good, perhaps we should have groups for grasses and dry grasses? I'm wondering if that would make the ABM more lightweight.
Apart from considering that :+1:

sofar:
Using groups is likely more intensive, as it causes a scan over all ContentFeatures to discover nodedefs with that group set.

paramat:
Ok :+1:

paramat:
Just thought, we could combine these 2 ABMs into 1 by having;
nodenames = dirt
neighbors = dirtwithgrass, dirtwithdrygrass, dirtwithsnow, grass 1-5, dry grass 1-5
instead of neighbors = air, as there's no point runnig the ABM on dirt under air if there are no nearby grass nodes.
The ABM could perhaps search for dirtwithgrass, dirtwithdrygrass, dirtwithsnow first as that's most likely, then if nothing is found then check for long grasses above.
Combining ABM searches helps with performance.

sofar:
yeah, that souns like a good idea actually. I'll take a look at it.

sofar:
Reordered these patches, combined all the relevant ones into two separate patches, rebased. 

This contains all the dirt-to-something abm's into a single ABM, as requested.

Tested on all likely scenarios.

sofar:
Added one more check for "air"

paramat:
Looks nice, will test.

sofar:
This isn't entirely working right yet. Dirt under dry_grass isn't getting converted, possibly due to light level.

sofar:
Ack, found the issue. The code should not have exited the function if (node not under air), but instead try (node under grass*) tests as well. Fixed, tested, and works now.

mods/default/functions.lua
```diff
@@ -353,32 +353,71 @@ minetest.register_abm({
...
+
+		-- look for likely neighbors
+		local p2 = minetest.find_node_near(pos, 1, {"default:dirt_with_grass",
+				"default:dirt_with_dry_grass", "default:dirt_with_snow"})
+		if p2 then
+			-- but the node needs to be under air in this case
+			local n2 = minetest.get_node(above)
+			if n2 and n2.name == "air" then
+				local n3 = minetest.get_node(p2)
+				if n3 then
```
paramat: Is this check necessary? since 'if p2 then' ensures one of those 3 nodes has been found.

sofar: yes, otherwise dirt under dirt_with_grass may turn to dirt_with_grass.

paramat: I mean 'if n3 then'

mods/default/functions.lua
```diff
@@ -353,32 +353,71 @@ minetest.register_abm({
 
 
 --
--- Grass growing on well-lit dirt
+-- convert dirt to something that fits the environment
```
paramat: Capital letter :)

mods/default/functions.lua
sofar: Ahh yes, we can skip that one.

mods/default/functions.lua
sofar: Sure.

sofar:
Pushed with the 2 requested changes.

paramat:
This is fun to test with rapid ABM action.
Works fine, just one thing i feel we should consider, with the light threshold at 13 we get large areas of dirt in shaded areas, i feel 11 is more reasonable:

screenshot_20160413_013740

We have lots of dense dark forests now and too much dirt is ugly.
What threshold seems good to others?

paramat:
I think 13 was chosen for when Minetest had jungles disabled and everything else was scattered appletrees, it needs to be retuned. 13 ensures that an isolated 5x5 tree will have grass growing up to the trunk.

sofar:
I don't see why we can't have dirt in darker areas resist grass due to player action - it adds a bit of variation that may be wanted. I would leave it at 13 for now.

paramat:
Yeah we can consider that later, so :+1:

paramat:
+1 from nore http://irc.minetest.ru/minetest-dev/2016-04-15#i_4584262

mods/default/functions.lua
```diff
@@ -353,38 +353,79 @@ minetest.register_abm({
...
+		end
+
+		local name = n2.name
+		-- Snow check is cheapest, so comes first.
+		if name == "default:snow" then
+			minetest.set_node(pos, {name = "default:dirt_with_snow"})
+		-- Most likely case first.
+		elseif name:find("default:grass") then
+			minetest.set_node(pos, {name = "default:dirt_with_grass"})
+		elseif name:find("default:dry_grass") then
```
kilbith: Why not using `sub` ?

paramat:
^ @sofar 

+1 from sfan5 too http://irc.minetest.ru/minetest-dev/2016-04-15#i_4584286

sofar:
Sure, name:sub will be slightly faster, I'll respin this.

sofar:
Changed to use name:sub(). Tested, repushed.

paramat:
Will merge in a few hours.

paramat:
64fe69f3823a3785c809e2e0a86d2e4e53134627
7d93272caa4503979abb9b98a05a9b801870a72b

PR #1002, opened 2016-04-01
Add sandstone wall to Walls mod

vlapsley:
Adding sandstone as a building material for the Walls mod.

### DIFF
--- a/mods/walls/init.lua
+++ b/mods/walls/init.lua
@@ -59,3 +59,5 @@ walls.register("walls:mossycobble", "Mossy Cobblestone Wall", "default_mossycobb
 walls.register("walls:desertcobble", "Desert Cobblestone Wall", "default_desert_cobble.png",
 		"default:desert_cobble", default.node_sound_stone_defaults())
 
+walls.register("walls:sandstone", "Sandstone Wall", "default_sandstone.png",
+		"default:sandstone", default.node_sound_stone_defaults())


kilbith:
:-1: See : https://github.com/minetest/minetest_game/issues/934#issuecomment-197345963

paramat:
Screenshot?

C1ffisme:
Even if we don't like it, we might as well see it.

vlapsley:
screenshot_20160402_075722

Inspiration from here:
sandstone_wall

kilbith:
It confirms what I said : that kind of walls is not single-piece. It's cobbly.

PilzAdam:
I agree with @kilbith, :-1:.

Fixer-007:
Sandstone is not cobbly by design, sandstone is made of compacted sand particles:
http://www.wallvision.com.au/wp-content/uploads/2013/11/bigstock-Stone-tile-wall-pattern-backgr-12363788.jpg
They break it in parts or saw them and then build walls with cement or even without.

C1ffisme:
What about sandstone brick as a wall? After seeing those pictures, those look more like bricks than smoothly carved stone.

vlapsley:
Using sandstone brick as the source material allows me to create something like this:

screenshot_20160402_123551

DonBatman:
I already have a mod for this plus other stones and bricks. The api makes it easy. Maybe keep walls mod with only a few and have mods so people can choose the ones they want.

C1ffisme:
@vlapsley Hmm... Looks promising, the texture might need tweaking, though.

paramat:
Sandstone brick doesn't look good as wall. Because sandstone is often sawn into large chunks i think it's reasonable to have huge blocks as in the first screenshot, that looks good to me. It would be consistent to have walls for each major stone type present in biomes.

sofar:
This is something for mods, IMHO.

sofar:
@paramat mossy cobble blocks can be obtained in survival mode, so why not allow people to craft something with it? It should be kept.

paramat:
Yes let's keep mossycobble wall.

sofar:
There's already a mod out there that offers this wall and more:

https://github.com/minetest-mods/mywalls

paramat:
Well, that's irrelevant. We have fences for every wood type, and walls for every mapgen stone type except sandstone, so seems a strange omission.

sofar:
We don't have walls for default:stone, nor default:desert_stone or default:obsidian, which are all stone.

I initially submitted this mod with walls for all materials, but people were not supportive of that, so I changed it to cover all the "cobble" materials. Now we're redoing the entire discussion again, but it's moot, since it was *accepted* and *merged* already with a proper discussion, and *only recently*.

Let this sit for a release or two, first, *then* come and discuss it again.

Ref: original PR thread - #796

paramat:
Okay, i don't feel strongly about it. By mapgen stone type i mean those types occuring in large amounts.

Gael-de-Sailly:
I think that the most important material for walls is not stone or cobble, it's stone bricks.

paramat:
The brick textures look wrong as walls.
I've removed my +1 for this as am neutral.

paramat:
Just realised this has 3 -1s, PIlzAdam, sofar and myself, closing.

PR #1006, opened 2016-04-02
Initial bushes: apple, acacia.

sofar:
Add 2 "bushes" schematics and plant them through mapgen.

These bushes are not growable through saplings. Only mapgen places
them.

The apple bush is planted in jungles and forms a large underbrush
that makes traversing the jungle less appealing, but not by much.
I've chosen to keep the cover relatively sparse, so that plenty
of jungle grass can still be found.

The acacia bush is a simple 3x3 bush, and planted sparsely like
logs. It's sparse enough to make a nice addition.

The schematics are just an initial attempt and probably need more
tuning.


sofar:
!()[http://i.imgur.com/cYhJYpk.png]
!()[http://i.imgur.com/GWeKWek.png]
!()[http://i.imgur.com/epVKATw.png]

C1ffisme:
The jungle bushes are a good idea, since they make the jungle a bit more difficult to traverse, although the pine bushes are a bit strange in that field.

What do bushes look like in the wilderness?

sofar:
I'm not happy yet with the pine bushes. I think I may just end up making 1 or 2 smaller pine tree like schematics instead, and mix those.

paramat:
Tha acacia bush looks good, i think these need to be low and wide and against the ground: 3x3x2 high plus randomness. Acacia bushes exist so +1 for that.
For pines instead i think it's better we have a second smaller conical pinetree: 3x3 horizontally, but this can wait or could be another PR.
Jungle is already very dense i'm not sure about adding more stuff, it doesn't need it.
What is needed is bushes for grasslands to provide wood. We can't use appletree leaves that drop appletree saplngs because that suggests appletrees grow there, I think a new 'bush leaves' node plus the existing 'tree' node (which is undefined species).

DonBatman:
Grass lands could have sage brush. It could drop sticks.

sofar:
!()[http://i.imgur.com/9J1y6qM.jpg]

I tuned the pine bush to be basically a little pine tree. The noise params for it were modified so that it makes small clusters. The little pine tree comes in size 3, 4 and 5 high and looks reasonable. I lowered the trunk size x1 height as well.

paramat:
I don't think these small pines should be in grassland though, because that suggests the full size tree grows there, these drop pine saplings so are essentially young full-size pines.

sofar:
@paramat the acacia bush is 3x2x3 already (sand root), it could be wider at the base if you ask me.

I think the little pine trees in stone grassland are reasonable since it won't add apple trees, but conifers are more than reasonable in almost any biome.

As for jungles, they should be diverse as that's what jungles are, and I really like the appeal of the apple bushes in them, they look really nice. I don't think it's bad to have apple tree saplings in the jungle per se.

Adding new node types is something I'd rather not do. I don't think it's needed at this point, I just want to better decorate the map with existing materials.

sofar:
I don't see why small pine trees are not appropriate - you could grow any sapling on grassland, and biome wise grassland would suit most of the trees we have just fine. Either apple bushes or small pine trees look entirely in the right place to me.

Just to show what kind of scarcity I'm going for:

sofar:
Tweaked sidelen = 16 for little pine trees.

sofar:
Added a separate decoration declaration for apple bushes in sandstone grassland. They're still fairly rare and clumped.

Fixer-007:
Nah, I was expecting bushes like in vanessas mods %) That ones kinda hacky.

paramat:
Good point, i guess a sapling will grow in grassland and have it's height limited by soil depth.
I'll test the PR.
Apparently rainforest forest floor is dark with little growth, just small plants and shrubs, so i think just the junglegrass is fine. I probably need to increase tree density to make it darker.

paramat:
I'm not keen on small pines in grassland, i think bushes are better. Those appletree bushes are too big, they're basically low trees, 3x3x2 is fine like the acacia bush.

sofar:
What you want sounds like the "bushes" in plantlifemodpack, and they don't solve the wood problem, because when you make them that small, there are no trunks in them, and therefore no wood. And that's the whole motivation for doing this.

I really don't understand why you don't want these bushes to drop saplings. That's the whole point - with a few bushes a player could start a small grove and grow some trees. Are you trying to prevent players from planting trees in grassland? That's just nonsensical! It's an open world! Trees can be grown almost anywhere!

I am opposed to adding new leaves just to do something entirely arbitrary and artificial like making leaves that don't drop saplings. And then make them use the same textures? That's just making things confusing for players.

Nah, if you don't like the bush models, we can tune them. But let's try not to create incomprehensible artifacts in game where one type of leaves drop saplings and the other doesn't. I made them all rare enough that they are going to be a nice and rare find for players, These will help players spawning in grasslands, but not much more, and the shapes are reasonable.

kilbith:
> I probably need to increase tree density to make it darker.

IIRC, you already increased it in the past, then decreased, and now increase it again ?

paramat:
Large bushes that are 3x3x2 are fine, like the your acacia bush, but no bigger, and i agree thay should have a full trunk node at the centre. I was thnking prob = 255 for the 5 in contact with the trnuk, and 191 or 223 for the rest.

Actually i'm not too bothered about whether we use existing leaf nodes for these, let's see what other devs think. After all a player can grab a sapling from a forest and grow it in grassland, and i agree should be able to.

sofar:
ok, that sounds reasonable, I'll start retuning this PR in a bit.

paramat:
Added to milestone as this will make players happier in the large grassland biomes.

sofar:
alright, I've done:

- tuned acacia and apple tree bushes as follows:
  - 3x3
  - top is never 100% leaves, always 2 corners missing, sometimes more
  - if planted on slope, will "drape" over it 1 node down
  - lowered leaf probability to make more irregular shaped
- took out my little happy pine trees (Bob Ross shed a big tear over this one).
- noise spreads set to 200

I *did* leave the apple bushes in the jungle (not in the swamp, that I removed) because I think it's a concession @paramat can make since I made a concession with the little pine trees :)

sofar:
Updated commit message and title.

sofar:
This needs noise param tuning for the apple trees in grassland, they're really way too rare now - I can't find them almost.

paramat:
:-1: If you research rainforest floor it only has small shrubs and plants, not large bushes, appletree bushes look out of place here. Rainforest is already fairly difficult to travel through, i am making it a little denser and darker and it is at the limit of density.

sofar:
I'd like this to be a bit more of an open discussion with some alternative opinions, instead of just the two of us. @Ekdohibs @sfan5 @PilzAdam ?

sofar:
I don't think a 3x3x2 schematic (especially as sparsely as it is right now) is considered a large bush. Did you try out the patch?

PilzAdam:
I generally like the idea of bushes, since they give the player a minimal amount of wood even if no trees are around (which could be very useful for mapgen v7). I also really like the dense jungles. But I have not a specific idea of how the bushes should actually look.

sfan5:
:+1: for bushes in rainforests and savanna

not sure about the others

Gael-de-Sailly:
I like very well the idea.
But I think that:
* If these plants are *young trees*, they should grow in the same biome as the trees
* If they are *bushes*, they should have their own nodes
Also, I think a slimmer trunk would be nice.

paramat:
3x3x2m is a large volume, when i researched rainforest floor it was clear it is a dark place with only a few small plants. If we want denser rainforest we can increase junglegrass density. Your screenshot of rainforest is an unusually sparse area, and it will be denser very soon due to another commit.
Apple bush seems out of place here and we could do without the extra nodes to render, FPS is already hit hard by rainforest.

sofar:
Not all rainforests are equal. Just read the wikipedia article. It says that if a hole in the canopy appears, the underbrush quickly fills the floor up to a "jungle" with many plants fighting for the sunlight, making the area very hard to move through.

paramat:
Okay, but there's no way of detecting canopy holes. I think if we add more plants to rainforest then something like ferns is more suitable.

sofar:
I've marked this as WIP. I think it's going to take a while to get most of the people to agree to something like this.

The core motivation is still to provide a few bits of wood (trunks) in grasslands.

So, obviously, savannah bushes and bushes in the jungle are secondary. I think the acacia-based savannah bushes is getting positive feedback. The jungle bushes I've added are getting very mixed results, so maybe I'll drop them for now (and yes, I'd love to add some wild plant variation like ferns but I'd like to do that after the plant mesh changes go in so we can make kick-ass fern plants with "6 leaves")

The remaining core problem with needing wood is still only solved by bushes in grassland. I don't want to make awkward looking bushes that are 1 high because it'll make the trunk show at the top, but maybe I can rotate it so that the side shows. It'll still be somewhat weird. I also don't want to make the trunk actually go into the ground like tree trunk roots.

I'm open to suggestions. I don't feel like introducing new nodes or new game concepts (read: some other way of obtaining wood), and I don't feel like we should avoid trying to artificially limit saplings that drop from used leaf nodes either as both would be counter intuitive and unneeded. If a person wants to grow a sapling in a grassland then that's their prerogative.

The dilemma whether players should be able to "grow" bushes themselves is also somewhat moot as they can be placed by the player easily, since they're so small. We are not preventing player from doing anything that's not possible right now. There may be even some game play value in encouraging players to mimic mapgen by creating their own bush shapes alongside their crafts.

Anyway, I'll likely wait a bit before updating this with some of the discussed tweaks.

paramat:
I might rework this to add just the acacia bushes in savanna (for now), as part of my biomes update.

MarkuBu:
Had a similar idea a few weeks ago. I use default leaves and wood for two sizes. In the background are basic models, along the river are small bushes and trees


screenshot_20160812_192425

paramat:
I feel that bushes are going to be so useful in many biomes that we do need a dedicated 'bush' node, that can perhaps drop sticks, but not saplings. So sorry but i disapprove of this PR, even though the acacia bush is almost perfect.

sofar:
I agree that having sticks drop would be nice(r). I just can't visualize how something like that would look (otherwise I would have made a PR already).

Fixer-007:
I somewhat dislike its look... trunk to be sure... it seems like a big small tree

paramat:
I'm thinking a node like a leaf node but generic.

sofar:
Yeah, I'm not a big fan of that. The multi-node bushes look good I think because their size is irregular and they fit well in the blocky world. If you'd make something like the dry shrub like the `shrubs` mod, it just wouldn't fit in I think. The bushes would have an entirely different look.

I'd prefer to have sticks drop over a full log myself too, but the aestetics of the current acacia bush is far better.

maybe we should place the log but with a different param2, and drop sticks in a custom dig handler?

paramat:
Sorry, that was my description of the 'bush' node itself, i meant using multiples of these in the same structure as your acacia bush. So essentially your acacia bush with dedicated generic leafy nodes.

paramat:
See #1327 
If it's okay i propose taking over on bushes as i am working on improving the biome system. I will use your acacia bush design with possible modifications discussed in #1327 .

sofar:
Yeah, let's just close this PR and see where #1327 goes. I'm optimistic.

PR #1011, opened 2016-04-03
Tweaked flower and mushroom spreading

tenplus1:
Edited flower spread abm to remove group:flora_block which is never referenced anywhere else in-game or mods, tidied spread routine and increased mushroom spread rate so they don't take over whole areas under shadow.

### DIFF
--- a/mods/flowers/init.lua
+++ b/mods/flowers/init.lua
@@ -83,6 +83,7 @@ minetest.register_abm({
 		pos.y = pos.y + 1
 		if under.name == "default:desert_sand" then
 			minetest.set_node(pos, {name = "default:dry_shrub"})
+			return
 		elseif under.name ~= "default:dirt_with_grass" then
 			return
 		end
@@ -92,18 +93,14 @@ minetest.register_abm({
 			return
 		end
 
-		local pos0 = {x = pos.x - 4, y = pos.y - 4, z = pos.z - 4}
-		local pos1 = {x = pos.x + 4, y = pos.y + 4, z = pos.z + 4}
-		if #minetest.find_nodes_in_area(pos0, pos1, "group:flora_block") > 0 then
+		local pos0 = vector.subtract(pos, 4)
+		local pos1 = vector.add(pos, 4)
+		if #minetest.find_nodes_in_area(pos0, pos1, "group:flora") > 3 or
+				#minetest.find_nodes_in_area(pos0, pos1, "group:flora_block") > 0 then
 			return
 		end
 
-		local flowers = minetest.find_nodes_in_area(pos0, pos1, "group:flora")
-		if #flowers > 3 then
-			return
-		end
-
-		local seedling = minetest.find_nodes_in_area(pos0, pos1, "default:dirt_with_grass")
+		local seedling = minetest.find_nodes_in_area_under_air(pos0, pos1, "default:dirt_with_grass")
 		if #seedling > 0 then
 			seedling = seedling[math.random(#seedling)]
 			seedling.y = seedling.y + 1
@@ -111,9 +108,7 @@ minetest.register_abm({
 			if not light or light < 13 then
 				return
 			end
-			if minetest.get_node(seedling).name == "air" then
-				minetest.set_node(seedling, {name = node.name})
-			end
+			minetest.set_node(seedling, {name = node.name})
 		end
 	end,
 })
@@ -169,17 +164,15 @@ minetest.register_abm({
 	action = function(pos, node)
 		if minetest.get_node_light(pos, nil) == 15 then
 			minetest.remove_node(pos)
-		end
-		local random = {
-			x = pos.x + math.random(-2,2),
-			y = pos.y + math.random(-1,1),
-			z = pos.z + math.random(-2,2)
-		}
-		local random_node = minetest.get_node_or_nil(random)
-		if not random_node then
 			return
 		end
-		if random_node.name ~= "air" then
+		local random = {
+			x = pos.x + math.random(-2, 2),
+			y = pos.y + math.random(-1, 1),
+			z = pos.z + math.random(-2, 2)
+		}
+		local random_node = minetest.get_node_or_nil(random)
+		if not random_node or random_node.name ~= "air" then
 			return
 		end
 		local node_under = minetest.get_node_or_nil({x = random.x,
@@ -187,9 +180,11 @@ minetest.register_abm({
 		if not node_under then
 			return
 		end
-		if minetest.get_item_group(node_under.name, "soil") ~= 0 and
-				minetest.get_node_light(pos, nil) <= 9 and
-				minetest.get_node_light(random, nil) <= 9 then
+
+		if (minetest.get_item_group(node_under.name, "soil") ~= 0 or
+				minetest.get_item_group(node_under.name, "tree") ~= 0) and
+				minetest.get_node_light(pos, 0.5) <= 3 and
+				minetest.get_node_light(random, 0.5) <= 3 then
 			minetest.set_node(random, {name = node.name})
 		end
 	end


DonBatman:
:+1:

PilzAdam:
- Don't do multiple different things in one commit / PR
- IIRC the flora_block group was there for some mod compatibly, maybe ask @VanessaE

tenplus1:
PilzAdam: I checked all mods for group:flora_block and nothing was in use, so was a useless piece of code, and the reason for multiple changes were a tidy of flower spreading which included mushrooms spreading too fast so I slowed it down.

VanessaE:
I actually don't remember the existence of such a group (let alone its purpose).

mods/flowers/init.lua
```diff
@@ -111,9 +106,7 @@ minetest.register_abm({
 			if not light or light < 13 then
 				return
 			end
-			if minetest.get_node(seedling).name == "air" then
-				minetest.set_node(seedling, {name = node.name})
-			end
```
PilzAdam: I'm not sure if this check for air can be safely removed. For example, fences can be above dirt_with_grass and may have a high enough light value.

PilzAdam:
@tenplus1 cleaning up the flower ABM has nothing to do with mushrooms growing too much. These are 2 separate issues.

Btw, your fix for the mushroom problem will only slow down mushroom growth, it does not fix the original problem that they will eventually fill the complete dark area they are in. A better fix would be to check the density of mushrooms in the area, like the flowers do currently. :-1: on your mushroom changes.

tenplus1:
@PilzAdam - Added mushroom count to stop them taking over area.

paramat:
:-1: Mushrooms need to completely fill an area for mushroom farming, they are for farming and a food source underground with limited dirt blocks.
Instead the thing to do is lower the light threshold that allows mushrooms to spread, to a value rarely occuring in mapgen. This will also extra challenge to growing them underground: players will have to 'turn the lights off' in their mushroom farms.

paramat:
```
		if not random_node then
			return
		end
		if random_node.name ~= "air" then
			return
		end
```
These can be combined into:
```
		if not random_node or random_node.name ~= "air" then
			return
		end
```
As the node is almost always existant there is no benefit to splitting these.

tenplus1:
The problem with mushroom spread is that in the wild biomes are covered with them during the night and it looks very unsightly, so limiting to no more than 4 per 3x3x3 area looks a lot better and still gives mushroom farms a good chance of working.  Will lower to 2x2x2 area if it helps.

paramat:
The chance for mushroom spread can then probably be left unchanged at 50, sofar tuned it for farming.

paramat:
Other changes to flowers are good though.

paramat:
Mushrooms love to grow on logs, this is often how they're farmed, so how about:
```
		if (minetest.get_item_group(node_under.name, "soil") ~= 0 or
                minetest.get_item_group(node_under.name, "tree") ~= 0) and
				minetest.get_node_light(pos, nil) <= 5 and
				minetest.get_node_light(random, nil) <= 5 then
			minetest.set_node(random, {name = node.name})
		end
```

paramat:
Good point, you can add 0.5 for time of day to get daylight value:
```
* `minetest.get_node_light(pos, timeofday)`
    * `timeofday`: `nil` for current time, `0` for night, `0.5` for day
```

sofar:
This should be squashed. I count 6 commits. ping me on IRC to walk you through it.

paramat:
:+1:

Gael-de-Sailly:
:-1: `group:flora_block` that is used in plantlife modpack:
[plantlife/poisonivy/init.lua 33](https://github.com/minetest-mods/plantlife_modpack/blob/2e0277e/poisonivy/init.lua#L33)

paramat:
Perhaps flora_block should be kept as at least 1 mod uses it, others might too.
Also i need to test this PR before +1.

mods/flowers/init.lua
```diff
@@ -187,9 +177,11 @@ minetest.register_abm({
...
 			return
 		end
-		if minetest.get_item_group(node_under.name, "soil") ~= 0 and
-				minetest.get_node_light(pos, nil) <= 9 and
-				minetest.get_node_light(random, nil) <= 9 then
+
+		if (minetest.get_item_group(node_under.name, "soil") ~= 0 or
+			minetest.get_item_group(node_under.name, "tree") ~= 0) and
+			minetest.get_node_light(pos, 0.5) <= 5 and
+			minetest.get_node_light(random, 0.5) <= 5 then
```
paramat: 3 lines above need indenting by one more tab to distinguish them from line 185.
The 2 light values actually need to be '<= 1' from testing in a dark area of rainforest, i did not realise rainforest could get this dark.

paramat:
screenshot_20160407_002052

This area of rainforest is so dark we need to change the light condition to '<= 1'. See line comment.

sofar:
Changing it to <= 5 would break existing mushroom farms, plus it wouldn't prevent people from farming them in jungles given that you've shown that light is that low there, so why change it?

tenplus1:
Mushrooms irl don't necessarily require absolute darkness to grow.

paramat:
0.4.13 stable used spores so we're free to change this. The original light threshold was '<= 9' which was too high, many places in mapgen have light at or below 9. '<= 5' was just a random number i came up with, it needs to be tuned.
I know mushrooms don't need complete darkness but here we are trying to stop mushrooms everywhere (on every node and tree) in shady forests. Try testing different light thresholds with ABM interval = 1 chance = 1.
Perhaps it's ok to have complete mushroom cover in the darkest places in rainforest?

tenplus1:
I tested light levels between 1 and 5 and for your new jungles 3 is the best option.

paramat:
Thanks, please could you fix the indentation issue and re-add the 'flora bock' check?

paramat:
:+1:

paramat:
An optimisation can be made here:
```
		if #minetest.find_nodes_in_area(pos0, pos1, "group:flora_block") > 0 or
				#minetest.find_nodes_in_area(pos0, pos1, "group:flora") > 3 then
			return
		end
```
It's best to have these 2 checks seperate with the group:flora check first as that is more likely to result in a 'return'. a 'flora block' 'return' only happens if poisonivy is present.

Flora tends to reach a steady state where it doesn't spread any further due to every flora having 4+ others nearby. This optimisation will therefore reduce 'find nodes in area' searches by 1 for most ABM actions.

Everything else looks good.

sofar:
3 is fine with me.

Ferk:
what about using the height to determine how much can they grow? So that mushroom farms only occur underground.

sofar:
That would make it impossible to make a mushroom farm in a tall building, or in a cave on a mountain. Or on a floatland.

paramat:
The optimisation will make a big difference as i've discovered that long grasses are also 'flora' and get dense in grassland.

tenplus1:
I still don't understand why we need the group:flora_block if it's only use is poison ivy...  It feels like an unecessary check which could further optimize the code if removed.

Zeno-:
The check will never even be performed if the first condition is true, so it's almost as optimised as not having the second... which is why paramat suggested swapping the order of conditions

tenplus1:
Optimizations done.

paramat:
I agree about the 'flora block' check being unimportant, i'll ask VanessaE.

paramat:
In the help section of github there are really good clear tutorials on how to fork a repo, store a copy locally, create branches for PRs and how to update PRs. I recommend working through this.
Once you have your PRs stored locally you will be able to update and rebase them within a minute, it's faster and easier than creating PRs remotely as you do now, or creating new ones for every change.

sofar's advice is good but probably overwhelming when following it at speed on IRC, the tutorials also leave out the more complex stuff.

paramat:
Well i'm going to request we remove that flora_block check @PilzAdam do you agree?

paramat:
@VanessaE please can we remove the check for 'group:flora_block'? As this adds a costly second 'find nodes in area' search to a frequently run ABM, and only for a mod that is most likely not used. I think it should be down to the poisonivy mod to do the extra work.

paramat:
Okay i'm over-ruled, the check has to stay.

kilbith:
I don't remember we ever adapted the upstream code **especially** to support some external mod...

sofar:
>@VanessaE please can we remove the check for 'group:flora_block'? As
>this adds a costly second 'find nodes in area' search to a frequently
>run ABM, and only for a mod that is most likely not used. I think it
>should be down to the poisonivy mod to do the extra work.

This is my concern.

Someone explain to me why plant life modpack can't override this method with a version of the same code that includes the check? That's what the override() method is for.

VanessaE:
I don't recall there being a feature to override an ABM.  The closest to that would be wrapping the current code (minus the flora_block check) inside a separate function that is then called from the current ABM.  Then a mod could redefine that function similar to what gloopblocks does (or did) with lava cooling.

The only problem then is, what if two or more mods need to do that?  Then you end up in a similar situation to what we had before the advent of cloning nodes with a deep table copy, or later override_item() - mods trying to create their own versions of existing features, and breaking each other because it was basically impossible for them to cooperate.

paramat:
Well i think it's acceptable to assume just plantlife will redefine this function.
Let's however propose this in a seperate PR to not drive tenplus1 crazy, this PR is okay for merge now :+1:

paramat:
I can see a couple of errors (not due to tenplus1) but perhaps we can merge this as is then i can fix the errors myself.
For flower spread:
```
		if under.name == "default:desert_sand" then
			minetest.set_node(pos, {name = "default:dry_shrub"})
		elseif under.name ~= "default:dirt_with_grass" then
			return
		end
```
This should 'return' after placing dry shrub?

For mushroom spread:
```
		if minetest.get_node_light(pos, nil) == 15 then
			minetest.remove_node(pos)
		end
```
This should 'return' after removing a mushroom killed by light?

tenplus1:
@paramat returns added

paramat:
Thanks, i can squash these when merging so that's not essential :+1:

paramat:
+1 from nore http://irc.minetest.ru/minetest-dev/2016-04-15#i_4584280

paramat:
9ece86cc7ada071f6eb64ff9f13e40f34a3c5ac0

PR #1013, opened 2016-04-04
Make cacti hurt players.

sofar:
We can trivially make players get hurt if getting close to cacti,
by using the lava `damage_per_second` property. In order to
allow players to actually get hurt, they need to be "inside" the
node. We therefore modify the collision_box so that this is
possible to actually do. We make the box large enough to not
allow the player to go in between 2 adjacent cacti, and high enough
that it's almost as high as a normal block, but still hurting the
player's feet.

This doesn't work unless we also do a better job in the engine to
actually see if toes, shoulders and forehead are actually in the
cactus block, since the engine currently only checks for the
horizontal center of feet/body/head, and not the extremities. You
can, without penalty, put your player face into lava and not
get hurt. This is obviously wrong, and I'm submitting a separate
engine change for that.

### DIFF
--- a/mods/default/nodes.lua
+++ b/mods/default/nodes.lua
@@ -891,6 +891,15 @@ minetest.register_node("default:cactus", {
 	groups = {snappy = 1, choppy = 3},
 	sounds = default.node_sound_wood_defaults(),
 	on_place = minetest.rotate_node,
+	damage_per_second = 1,
+	collision_box = {
+		type = "fixed",
+		-- Adjust collision box so it damages players that are too
+		-- close. We need to allow them to "enter" into the node a
+		-- little bit for this to actually happen. The top is low
+		-- enough that feet damage occurs if standing on it.
+		fixed = {-5/16,-1/2,-5/16,5/16,0.39,5/16},
+	},
 })
 
 minetest.register_node("default:papyrus", {


sofar:
Needs: https://github.com/minetest/minetest/pull/3961

nerzhul:
:+1:

kilbith:
Alternatively the cactus can me made a nodebox, sized like a regular node - 1/16px on X/Z. You'd take damages inside this 1/16px space and it'd allow to have proeminent spines on the sides for a thornier looking.

cactus

(BTW : a mesh model is really not necessary)

sofar:
@kilbith I played around with blender and made something similar yesterday, but it doesn't work at all with mapgen-planted cactus tree decorations.

Fixer-007:
Will be a nice feature, bring it on!

paramat:
:-1: Sorry but i don't think we should bother with minor damage like this, just like we don't make a furnace burn if too close. Also, cactus is often used as a building material and i feel players shouldn't be damaged by their creations.
-1 for the engine chage too.

benrob0329:
I feel the exact opposite, its all the little things that make a game (even a minimal one) seem well rounded and "polished".

Having cactus, and not having it hurt you seems rather backwards and incomplete (I mean, cactus are known for their spikes, after all) and I feel that not bothering with the "minor" things can take something great and make it seem rather melancholy.

ExeterDad:
Cactus hurts. My furnace does not hurt me when touched. In fact, most kilns that I have seen are (mostly) comfortable to the touch when in operation. It's by design that the heat is kept where it is most useful rather then it radiating high temperatures
.
I will be very disappointed if this feature collects dust until forgotten.

paramat:
Good points. My point about the MTGame furnace is that it has an open front with the fire close behind, this will burn.

sofar:
This isn't about cacti necessarily.

What this patch does is enable normal nodes to better damage the player, if the player is `partially inside` the node.

This is the mechanism that lava nodes uses to damage the player. It is extremely efficient as it involves no lua code other than handling the resulting damage, and is done during collision time.

Now, the current pre-patch code doesn't damage the player unless the player is `almost entirely inside` the lava node.

This isn't a discussion about MC-vs-MT. When I have my body 49% covered in a boiling lava fountain (sideways) and the old code `is not damaging the player`, because it only checks on the player median vertical axis. That is a game inconsistency - because the lava will hurt the same player even if his pinky toe even only goes 1cm into that lava downwards.

So, see it more as "fixing lava damage" that "coincidentally" also makes other damaging nodes doable.

paramat:
The highest and lowest checked points are 0.1 and 1.6 nodes above the feet, so these are already vertically a little way inside the player too although by not as much. Perhaps the highest point should be moved to 0.1 below top of the head (if it isn't already).
Because of the inconsistency i'm less opposed to the engine change than this PR.

sofar:
So, move it from `1.6` to `1.7` ?

paramat:
Well first we need to confirm the actual height of the player model, i may be wrong about it being 1.8.

paramat:
Vanessa says 1.768 nodes, closer to 1.8 than 1.7, i wonder if the damage point must be more than 0.1 nodes below top of head? I'm tempted to assume c55 made it that way for good reason, maybe let's ask him.

paramat:
Primarily it's the engine change i oppose.

tobyplowy:
@sofar +1

Zeno-:
Won't add in engine. Too bothersome.

paramat:
sfan5 is +1 http://irc.minetest.net/minetest-dev/2017-04-23#i_4893294
http://irc.minetest.net/minetest-dev/2017-04-23#i_4893298 rubenwardy: I support the concept of cactus damage

paramat:
https://github.com/minetest/minetest/pull/3961 has been replaced by https://github.com/minetest/minetest/pull/6611
For this MTG feature to work a further engine change would be needed for damage detection on player collisionbox edges.

paramat:
Closing for now, has concept approval but depends on a possibly problematic change in the engine.
'Problematic' as in for large custom players the number of checks would be very large.

PR #1034, opened 2016-04-14
Gui hotbar updates

KevDoy:
Thought it could use a more pixel based look like in other games. The
current one looks a bit strange with it’s 3D lighting.


KevDoy:
Please remove, this was a mistake.

PR #1038, opened 2016-04-15
Added on_blast to default chest

tenplus1:
Added on_blast function to default chest so that tnt explosions or mob blasts do not destroy it.  Also tidied some code.

### DIFF
--- a/mods/default/nodes.lua
+++ b/mods/default/nodes.lua
@@ -1463,16 +1463,17 @@ minetest.register_node("default:chest", {
 		minetest.log("action", player:get_player_name() ..
 			" moves stuff in chest at " .. minetest.pos_to_string(pos))
 	end,
-    on_metadata_inventory_put = function(pos, listname, index, stack, player)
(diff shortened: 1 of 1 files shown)


sofar:
I only see the cleanups.

sofar:
This isn't what we discussed on IRC. The mobs API currently "ejects" all items in the chest. This is what should be done as well.

sofar:
We should also make sure the chest itself is dropped.

tenplus1:
The chest will be destroyed in blast so would not drop, only it's contents.

sofar:
Why are you randomly ejecting items from the chest? The TNT API should take care of that. I think you actually want to return the items from the chest instead of dropping them.... This way the TNT API takes care of everything.

sofar:
You want to return the chest item as well as return value.

sofar:
I'll submit a patch that does this for all the "containers" in a nice and uniform way. Sorry for making you spend time on this - please close this.

sofar:
Replaced by #1039

PR #1039, opened 2016-04-16
Tnt+: Changes for TNT mod after #862

sofar:
These are queued changes for after pull #862 
- Drop items from chests, shelves (book, vessel), furnaces (and disable tnt destroying bones)

### DIFF
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -111,6 +111,21 @@ minetest.register_abm({
 
 
 --
+-- optimized helper to put all items in an inventory into a drops list
+--
+function default.get_inventory_drops(pos, inventory, drops)
+	local inv = minetest.get_meta(pos):get_inventory()
+	local n = #drops
+	for i = 1, inv:get_size(inventory) do
+		local stack = inv:get_stack(inventory, i)
+		if stack:get_count() > 0 then
+			drops[n+1] = stack:to_table()
+			n = n + 1
+		end
+	end
+end
+
+--
 -- Papyrus and cactus growing
 --
 
--- a/game_api.txt
+++ b/game_api.txt
@@ -241,6 +241,23 @@ tnt.register_tnt(definition)
 
 ^ Ignite TNT at position
 
+
+To make dropping items from node inventories easier, you can use the
+following helper function from 'default':
+
+default.get_inventory_drops(pos, inventory, drops)
+
+^ Return drops from node inventory "inventory" in drops.
+
+* `pos` - the node position
+* `inventory` - the name of the inventory (string)
+* `drops` - an initialized list
+
+The function returns no values. The drops are returned in the `drops`
+parameter, and drops is not reinitialized so you can call it several
+times in a row to add more inventory items to it.
+
+
 Screwdriver API
 ---------------
 
--- a/mods/bones/init.lua
+++ b/mods/bones/init.lua
@@ -123,6 +123,8 @@ minetest.register_node("bones:bones", {
 			return true
 		end
 	end,
+	on_blast = function(pos)
+	end,
 })
 
 local function may_replace(pos, player)
--- a/mods/default/furnace.lua
+++ b/mods/default/furnace.lua
@@ -260,6 +260,15 @@ minetest.register_node("default:furnace", {
 		local timer = minetest.get_node_timer(pos)
 		timer:start(1.0)
 	end,
+	on_blast = function(pos)
+		local drops = {}
+		default.get_inventory_drops(pos, "src", drops)
+		default.get_inventory_drops(pos, "fuel", drops)
+		default.get_inventory_drops(pos, "dst", drops)
+		drops[#drops+1] = "default:furnace"
+		minetest.remove_node(pos)
+		return drops
+	end,
 
 	allow_metadata_inventory_put = allow_metadata_inventory_put,
 	allow_metadata_inventory_move = allow_metadata_inventory_move,
--- a/mods/default/nodes.lua
+++ b/mods/default/nodes.lua
@@ -1474,6 +1474,13 @@ minetest.register_node("default:chest", {
 			" takes " .. stack:get_name() ..
 			" from chest at " .. minetest.pos_to_string(pos))
 	end,
+	on_blast = function(pos)
+		local drops = {}
+		default.get_inventory_drops(pos, "main", drops)
+		drops[#drops+1] = "default:chest"
+		minetest.remove_node(pos)
+		return drops
+	end,
 })
 
 minetest.register_node("default:chest_locked", {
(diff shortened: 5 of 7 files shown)


sofar:
Once #862 is merged, most of these commits will all disappear after a rebase.

tenplus1:
The furnace drops have "default_chest" instead of "default:chest"

mods/default/functions.lua
```diff
@@ -111,6 +111,21 @@ minetest.register_abm({
...
 --
+-- optimized helper to put all items in an inventory into a drops list
+--
+function default.get_inventory_drops(pos, inventory, drops)
+	local inv = minetest.get_meta(pos):get_inventory()
+	local n = #drops
+	for i = 1, inv:get_size(inventory) do
+		local stack = inv:get_stack(inventory, i)
+		if stack:get_count() > 0 then
+			drops[n+1] = stack:to_string()
```
kilbith: Use `drops[i]`, should be faster.

kilbith:
`tnt/init.lua` - lines 441, 458, 459, 466 have EOL trailing whitespaces.

mods/default/functions.lua
sofar: No, that would break the ability for this function to append drops - it was written so that it can be called multiple times - each time adding more items but not overwriting existing entries already in drops[]

sofar:
> The furnace drops have "default_chest" instead of "default:chest"

instead of "default:furnace", of course.

sofar:
@kilbith 

> tnt/init.lua - lines 441, 458, 459, 466 have EOL trailing whitespaces.

This PR is only about "TNT: Add on_blast to all nodes with an inventory" - the rest of the patches will disappear when #862 is merged. Since this patch doesn't modify tnt/init.lua, it's not relevant for this PR.

paramat:
Not sure if this the place for it but, is it necessary to nodeupdate within 'radius' if those nodes are destroyed? Not doing this could save a lot of nodeupdates.

sofar:
Rebased.

sofar:
@paramat technically it's possible for nodes to be blast resistant, but wallmounted or attached, or falling. I don't foresee big problems calling nodeupdate on "air" a bunch of times though, it should return really fast.

paramat:
Ok that's fine then.

paramat:
Is this ready for merge?

sofar:
Yes, I'm still contemplating on making the maximum entity velocity lower - that's the one patch that I haven't written yet, but it could get merged separately and doesn't influence any of these patches.

The player knockback is a bit experimental, it may blow a player into a block. Then again, if that happens they were too close to a TNT blast and did take significant damage :)

sofar:
Two more almost trivial changes

sofar:
I spotted a critical omission in the on_blast chest handling: books would lose all their metadata, oops.

This was a bug that you can actually spot in my video of 2 days ago - and that's exactly where I spotted it too (lol)

The simple fix is to properly make itemstacks and use ```to_table()```. It's folded into the first patch in this series.

paramat:
All tests good so far. It's possible to blow a player inside a block (damage disabled) but this is fun.
Let us know when this is in a good state for merge then we'll test and review.

paramat:
I have a request. I think the speed of smoke particles can be much lower, the fireball of an explosion tends to be much slower than the shrapnel, this would also make the explosion look more solid and smoky at it's centre, and less empty. I'll try some values.
Possibly smoke particles can have collision detection disabled as this is much more intensive?

sofar:
smoke particles have collision detection disabled. Only the "blast fragments" particles have it.

sofar:
My idea for TNT smoke is to make the particle textures animated and replace the smoke particles with something better and animated.

paramat:
Ah ok. For some reason i thought i saw smoke bouncing around.
I've tried halving smoke speed and looks much better:
```
		minvel = {x = -10, y = -10, z = -10},
		maxvel = {x = 10, y = 10, z = 10},
```

sofar:
I'm fine with that. I'd also lower the amount of smoke particles...

paramat:
Ok, if you make those changes i'll then test and re-review.

sofar:
@paramat tweaked it a bit and added.

paramat:
Will test.

paramat:
:+1:

paramat:
@PilzAdam @sfan5 @ShadowNinja please can anyone review this?
Tests ok but i haven't tested mob damage.

paramat:
@sofar I have an idea
```
minetest.register_node("tnt:boom", {
	drawtype = "plantlike",
	tiles = {"tnt_boom.png"},
```
Perhaps we can use 'visual scale' on the boom texture to make it larger? perhaps 300%?

paramat:
Looks good but there's a problem with it being scaled from the base, it appears too high, i'll adjust the texture.

screenshot_20160425_142944

paramat:
Edited texture 
tnt_boom_edit

sofar:
Maybe keep that for a next release? The bottom image does look decent!

C1ffisme:
It still doesn't look a lot like an explosion...

paramat:
Well i'll leave it up to you, it's simple and easy to add, and the current effect is pathetically small.

paramat:
It also results in something large to be lit-up by the boom effect, very effective at night time or in dark places.

sofar:
I'll look at it after work. It does make a big visual difference, I almost wonder if we should make it a particle instead to prevent it from lingering and requiring 2 node updates (but we still should emit light in the area somehow)

paramat:
Just realised this boom node shoulld be placed 1 node lower, then it will line up nicely with the original TNT. Because it's plantlike the visual scaling operates from the base of the node upwards. I'll test this.

paramat:
Works well with:
```
	minetest.set_node({x = pos.x, y = pos.y - 1, z = pos.z}, {name = "tnt:boom"})
```
But now i get undiggable TNT entities dropped because i guess the boom node is exterior to the centre of the explosion.

sofar:
This is working really well as a particle. The nice thing is that a particle's timing is really precise, whereas a node timer can be slow. So instead, I'd like to do it as follows:

1) tnt:boom becomes airlike, no texture
2) tnt:boom on_construct() has minetest.add_particle(), the particle itself is 30 large (3x a node) and is automatically centered at pos, and expires in 0.2secs.

this won't leave nodes behind since the node timer on tnt:boom removes itself.

sofar:
rebased, repushed including the boom particle.

I did not change the actual tnt_boom.png texture

sofar:
Updated particle time and nodetimer light removal time to 0.4s. Note that the nodetimer will likely not run before 1.0 second due to the default nodetimer_interval.

paramat:
:+1:

paramat:
screenshot_20160426_130206

paramat:
Approval from sfan5 http://irc.minetest.ru/minetest-dev/2016-04-26#i_4595376

paramat:
54b87e955de6f6e5f692d9131fcf72e4baaf7986 to 5cd4b5a23192e3a4e16b87ecea4a732cea36115d

PR #1048, opened 2016-04-21
Flowers: Fix waterlily node's duplication with stack of 1

kilbith:
Fixes https://github.com/minetest/minetest_game/issues/1046.

Cc: @paramat

### DIFF
--- a/mods/flowers/init.lua
+++ b/mods/flowers/init.lua
@@ -226,6 +226,7 @@ minetest.register_node("flowers:waterlily", {
 	sunlight_propagates = true,
 	groups = {snappy = 3, flower = 1},
 	sounds = default.node_sound_leaves_defaults(),
+	node_placement_prediction = "",
 	node_box = {
 		type = "fixed",
 		fixed = {-0.5, -0.5, -0.5, 0.5, -0.46875, 0.5}
@@ -235,12 +236,18 @@ minetest.register_node("flowers:waterlily", {
 		fixed = {-0.5, -0.5, -0.5, 0.5, -0.4375, 0.5}
 	},
 
-	on_place = function(itemstack, _, pointed_thing)
+	on_place = function(itemstack, placer, pointed_thing)
 		local pos = pointed_thing.above
 		local node = minetest.get_node(pointed_thing.under).name
 		local def = minetest.registered_nodes[node]
+		local player_name = placer:get_player_name()
+
 		if def and def.liquidtype == "source" and minetest.get_item_group(node, "water") > 0 then
-			minetest.set_node(pos, {name = "flowers:waterlily", param2 = math.random(0, 3)})
+			if not minetest.is_protected(pos, player_name) then
+				minetest.set_node(pos, {name = "flowers:waterlily", param2 = math.random(0, 3)})
+			else
+				minetest.chat_send_player(player_name, "This area is protected")
+			end
 			if not minetest.setting_getbool("creative_mode") then
 				itemstack:take_item()
 				return itemstack


PilzAdam:
Why is this necessary? What is the underlying problem for this hack?

kilbith:
The answer is somewhere in the lines of C++, but anyway returning the `itemstack` triggers the bug.

paramat:
Thanks, much appreciated, will test.

Zeno-:
(referring to bug report as well because the comments in it are relevant)

If this is a workaround for release only, how will we remember to look for the true cause after release?

HybridDog:
screenshot_20160422_100305
l didn't manage to duplicate the waterlily, l had one left in survival mode, placed it and then it was taken from inventory.
There was just a bug with node placement prediction, so it looks like another waterlily is under the placed one, after rejoining the other one is gone.
Maybe you only need to execute nodeupdate() if the last waterlily gets placed.

mods/flowers/init.lua
```diff
@@ -235,15 +235,22 @@ minetest.register_node("flowers:waterlily", {
...
 		local def = minetest.registered_nodes[node]
 		if def and def.liquidtype == "source" and minetest.get_item_group(node, "water") > 0 then
 			minetest.set_node(pos, {name = "flowers:waterlily", param2 = math.random(0, 3)})
 			if not minetest.setting_getbool("creative_mode") then
-				itemstack:take_item()
-				return itemstack
+				if itemstack:get_count() == 1 then
+					local inv = placer:get_inventory()
+					minetest.after(0, function()
+						inv:set_stack("main", placer:get_wield_index(), nil)
```
HybridDog: doing it this way is dangerous, if the player places it and then switches to another item, the wrong one gets removed

HybridDog:
How does placing waterlilies look like if the server lags?

mods/flowers/init.lua
kilbith: Indeed, it's better to localize before.

paramat:
> If this is a workaround for release only, how will we remember to look for the true cause after release?

We can keep the issue open when this is merged, i've added a comment in the issue.

paramat:
Tested :+1:

Ekdohibs:
The code is still dangerous, if the player inventory changes right after the player places the lilypad, then the wrong item will be removed. Isn't there a way to disable lily pad placement prediction instead? (Which is wrong anyway).

paramat:
Ekdohibs proposal sounds better to me, it would also stop the waterlily appearing underwater briefly.

kilbith:
Applied @Ekdohibs suggestion, which is much better.

Ekdohibs:
:+1: from me now

paramat:
Testing.

paramat:
Seems good :+1:

mods/flowers/init.lua
```diff
@@ -235,7 +236,7 @@ minetest.register_node("flowers:waterlily", {
 		fixed = {-0.5, -0.5, -0.5, 0.5, -0.4375, 0.5}
 	},
 
-	on_place = function(itemstack, _, pointed_thing)
+	on_place = function(itemstack, placer, pointed_thing)
```
HybridDog: the player could be used for testing if there's protection

kilbith:
Protection check added.

mods/flowers/init.lua
kilbith: Thanks for deleting this comment.

paramat:
:+1:

paramat:
16fe1efcd82a8274989962bf19f7d6f7befd3cfe

PR #1082, opened 2016-05-08
Mese texture

Ekdohibs:
See discussion in #1066 and #1074

mese


PilzAdam:
I think "MESE" should be actually readable on the texture.

Thomas--S:
@PilzAdam I can understand your point of view. It would make the game maybe a bit easier to understand for new players.

I am neutral if our new texture from above or the current MESE texture should win.

Ekdohibs:
@PilzAdam I am neutral on that; I suggested this texture because it is closer to the current one that the other that were suggested, and I like it better. I would also be fine with keeping the current texture.

MT-Modder:
I think if the texture's going to change it should be this one. It does exactly as #1066 requests, keeps the style but changes the "language".

mtango688:
Actually, it currently says "ME" and "SE" underneath. I'm not sure what the benefit of having a readable texture is, though. In some cultures that would read as "EE" and "MS" (which only further baffles me).

sofar:
I think the original name and texture was the biggest fuckup that this project has ever done.

sofar:
Actually, let me rephrase that. I think the choice of "mese" as a name for the block, and the texture showing these letters was the only fuckup this project ever made.

I don't think it's too late to fix it, either. This is a great step forward. :+1:

C1ffisme:
Do we really need two pull requests, and a discussion issue?

Eh... I don't think Minetest would be the same without Mese. It's been in the game since almost the creation of the game. (Before 2011.)

mtango688:
Does sentimental value really outweigh crude, unprofessional aesthetics?

paramat:
This having the same lettering graphics as the current one is i feel the problem, the current texture has ugly graphic design. Also i feel it should read ME, SE as the classic and be more recognisably 'MESE'. But this is all obvious from my PR :) Still an interesting submission.

HybridDog:
>Does sentimental value really outweigh crude, unprofessional aesthetics?

Of course, it does, crude, unprofessional aesthetics are the essence of a game.

mtango688:
@HybridDog Okay, then what this game needs is a crude, unprofessional overhaul of every other texture. lol

HybridDog:
>@HybridDog Okay, then what this game needs is a crude, unprofessional overhaul of every other texture. lol

no, don't waste your time doing this, add more nodes with new textures instead

everamzah:
mese
mese_preview

paramat:
^ That one is better.

sofar:
I can compromise, everamzah's version is a good step forward :)

benrob0329:
I agree.

Ekdohibs:
I think it is too

mtango688:
At first, I was meh but it's grown on me.

paramat:
I edited everamzah's version, just to remove unnecessary alpha channel, add 1 extra pixel of dark shading at the lower left corner, and converted to indexed for a smaller file.

everamzahmese

everamzahmese128

My submission isn't very popular, i would be happy with this one and would +1 it.

ExeterDad:
@paramat somehow aside from the mentioned pixel the entire appearance of your edited image is different. Perhaps the the indexing washed out the noise? About the pixel on the lower left, I'm not sure that was a improvement. @everamzah 's image looked as if the light source was above and a bit to the left. I can't tell what direction the light is coming from as the shadow shading is now impossible. That said, removing the unneeded alpha channel was a good call.

cdqwertz:
I made a mese-alphabet based on @everamzah 's texture:
mese_alphabet_scaled

benrob0329:
*claps*

C1ffisme:
@cdqwertz Hmmm... Good, but not all languages have a full 26 letter alphabet. (I think the French alphabet did not have a W for a while, Latin didn't have a W or a J, and the [Hawaiian language](https://en.wikipedia.org/wiki/Hawaiian_language#Orthography) has only 12 letters.)

sofar:
@cdqwertz please submit this in a separate issue (proposal: custom glyph alphabet for minetest - or something like that) so people can discuss use and issues in there.

Back to on-topic: @paramat: that texture is good.

paramat:
My edit should be identical to everamzah's apart from that shadow pixel, indexing should not change anything in a simple image.

paramat:
I'll update my edit to have identical shading to everamzah's.

HybridDog:
> indexing should not noticeably change anything in a simple image, it allows a palette of 256 colours. Many or most Minetest textures are converted to indexed to reduce file size.

The texture is 16x16, so it can't have more than 256 colours, i.e. indexed does only change file size, if every pixel has a different colour, indexed usually causes a higher file size.
Do not manually convert it to indexed, use adv- and optipng, they automatically do it and they also reduce the file size in other ways.

paramat:
See #1093 @Ekdohibs okay to close this PR?

paramat:
Closed with permission.

PR #1096, opened 2016-05-17
Change Drinking Glass Texture

### LINKED ISSUE #1089: Drinking Glass Texture
If you place a drinking glass it looks really strange:

screenshot_20160516_112659

I think we should fix that.

Thomas--S:
Doesn't affect the inventory image. Fixes #1089.

New:

Old:

Please notice, that this is my first actual PR. I'm quite new to git. I hope I've submitted this PR correctly.

I also changed the visual_scale to 0.9 and changed the selection box accordingly.

New image:


C1ffisme:
Your git seems to have worked correctly...

Thomas--S:
> Your git seems to have worked correctly...

Thanks.

sfan5:
:+1:

paramat:
:+1:

sofar:
Good, this will fit relatively well with the glass block texture.

tenplus1:
Would changing the visual_scale be ok as well since it's rather large when placed ?

Thomas--S:
I changed the visual_scale to 0.9 and changed the selection box accordingly.

New image:

C1ffisme:
@Thomas--S Good, but it's still about a meter tall...

paramat:
Please can the visual scale be 1.0? it's not much different and the pixels should be equal in size to those of bottle and steel bottle.

paramat:
Also, please can the inventory image be identical, like the other vessels? I'm not sure why we have seperate inventory textures when they're identical though.

paramat:
The 'inventory' textures are exact duplicates, please can you delete these as they are unnecessary?

paramat:
If we make these vessels smaller i think the best way would be to actually make smaller images within the 16x16 canvas, so that pixel size remains the same as all other surrounding textures. But that could be another PR.

C1ffisme:
@paramat I was thinking so too. Here's a texture made for fun, although it's still 1/2 a meter tall. (Well, the blocks are a bit less than a meter though...)

vessels_glass

Thomas--S:
@paramat: The inventory image for the drinking glass can't be the same as the node texture, because otherwise it would look really strange.

I will update this PR in about half a day.

Thomas--S:
Replaced by #1099. Thanks for your efforts.

paramat:
> The inventory image for the drinking glass can't be the same as the node texture, because otherwise it would look really strange.

Ok no problem.

Thomas--S:
Thanks, @paramat

PR #1101, opened 2016-05-20
save respawn entering bed, remove constant read_spawns

tenplus1:
Save respawn position when entering bed.  I also removed the constant beds.read_spawns() calls when a player joins as this is only required once https://github.com/minetest/minetest_game/pull/1102

### DIFF
--- a/mods/beds/functions.lua
+++ b/mods/beds/functions.lua
@@ -111,7 +111,14 @@ local function update_formspecs(finished)
 		if is_majority and is_night_skip_enabled() then
 			form_n = form_n .. "button_exit[2,8;4,0.75;force;Force night skip]"
 		end
-	end
+
+		if not is_sp then
+			minetest.chat_send_all(">> (" .. player_in_bed .. " / "
+				.. ges .. ") players are sleeping!")
+		end
+
+		beds.set_spawns() -- save respawn positions when entering bed
+end
 
 	for name,_ in pairs(beds.player) do
 		minetest.show_formspec(name, "beds_form", form_n)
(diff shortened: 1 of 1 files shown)


everamzah:
So with this, a player sets their respawn position whenever they right-click a bed?  I encourage the change from the previous behavior, which required a group majority to sleep and skip night in order to set respawn.  That very rarely ever happened.

What about right-clicking a bed in another player's protected area?  Is it alright to allow for any player to set their respawn at any other player's (protected) bed?

mods/beds/functions.lua
```diff
@@ -111,7 +111,14 @@ local function update_formspecs(finished)
 		if is_majority and is_night_skip_enabled() then
 			form_n = form_n .. "button_exit[2,8;4,0.75;force;Force night skip]"
 		end
-	end
+
+		if not is_sp then
+			minetest.chat_send_all(">> (" .. player_in_bed .. " / "
+				.. ges .. ") players are sleeping!")
```
kaeza: This will cause an error, as `ges` is a table.

sofar:
I'm positive to the idea. I've had issues setting bed spawns myself and it's confusing, this is simpler. I would even set bed spawns when it isn't night (right-click bed at day time to change spawn).

mods/beds/functions.lua
tenplus1: ```ges``` is a number.

kaeza: Welp. My bad. Didn't see the `#` in front.

PR #1105, opened 2016-05-21
add group grown and can_grow for grass growing abm

pithydon:
Allows one to add new types of dirt that grass can grow on without
editing the ABM.

### DIFF
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -384,12 +384,11 @@ minetest.register_abm({
 --
 
 minetest.register_abm({
-	nodenames = {"group:can_grow"},
+	nodenames = {"group:dirt"},
 	neighbors = {
-		"group:grown",
 		"group:grass",
 		"group:dry_grass",
-		"default:snow",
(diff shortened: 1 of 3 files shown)


sofar:
I'm not entirely ok with this. The group names you chose are confusing, and I don't like the string matching code in an often called routine. I prefer table lookups. I'm not a fan of this patch in this form.

pithydon:
It no longer adds group grown and can_grow, but does add group dirt.

sofar:
I'm less than enthusiastic, still. The added complexity and all the new groups are not nice - you add a lot of new groups just to make one very specific thing work. I don't think the dirt group was needed, but I'd rather see the ABM on `default:dirt` anyway (since that is what it's about here).

Sorry, I can't think of a better solution myself, other than something way over complex like a callback or table entry in the nodedef? Maybe `promotes_dirt_grow_to = "default:dirt_with_dry_grass"` ? Terribly ugly, as well (but far less complex)

PR #1111, opened 2016-05-24
Big simplification of slabs combination

sofar:
Combine slabs if identical based on orientations using a simple
lookup table if the nodes are identical.

Otherwise relies on place_node() to place the node, which properly
handles rotation compared to adjacent nodes already, and can
orient based on look_dir as well.

Largely based on kilbith's #807 PR. Slab combining and place_node()
usage by sofar.

### DIFF
--- a/mods/stairs/init.lua
+++ b/mods/stairs/init.lua
@@ -110,6 +110,11 @@ function stairs.register_stair(subname, recipeitem, groups, images, description,
 end
 
 
+-- Slab facedir to placement 6d matching table
+local slab_trans_dir = {[0] = 8, 0, 2, 1, 3, 4}
+-- Slab facedir when placing initial slab against other surface
+local slab_trans_dir_place = {[0] = 0, 20, 12, 16, 4, 8}
+
 -- Register slabs.
 -- Node will be called stairs:slab_<subname>
 
@@ -129,86 +134,61 @@ function stairs.register_slab(subname, recipeitem, groups, images, description, 
 			fixed = {-0.5, -0.5, -0.5, 0.5, 0, 0.5},
 		},
 		on_place = function(itemstack, placer, pointed_thing)
-			if pointed_thing.type ~= "node" then
-				return itemstack
-			end
+			local under = minetest.get_node(pointed_thing.under)
+			local wield_item = itemstack:get_name()
 
-			-- If it's being placed on an another similar one, replace it with
-			-- a full block
-			local slabpos = nil
-			local slabnode = nil
-			local p0 = pointed_thing.under
-			local p1 = pointed_thing.above
-			local n0 = minetest.get_node(p0)
-			local n1 = minetest.get_node(p1)
-			local param2 = 0
+			if under and wield_item == under.name then
+				-- place slab using under node orientation
+				local dir = minetest.dir_to_facedir(vector.subtract(
+					pointed_thing.above, pointed_thing.under), true)
 
-			local n0_is_upside_down = (n0.name == "stairs:slab_" .. subname and
-					n0.param2 >= 20)
+				local p2 = under.param2
 
-			if n0.name == "stairs:slab_" .. subname and not n0_is_upside_down and
-					p0.y + 1 == p1.y then
-				slabpos = p0
-				slabnode = n0
-			elseif n1.name == "stairs:slab_" .. subname then
-				slabpos = p1
-				slabnode = n1
-			end
-			if slabpos then
-				-- Remove the slab at slabpos
-				minetest.remove_node(slabpos)
-				-- Make a fake stack of a single item and try to place it
-				local fakestack = ItemStack(recipeitem)
-				fakestack:set_count(itemstack:get_count())
-
-				pointed_thing.above = slabpos
-				local success
-				fakestack, success = minetest.item_place(fakestack, placer,
-					pointed_thing)
-				-- If the item was taken from the fake stack, decrement original
-				if success then
-					itemstack:set_count(fakestack:get_count())
-				-- Else put old node back
-				else
-					minetest.set_node(slabpos, slabnode)
-				end
-				return itemstack
-			end
-			
-			-- Upside down slabs
-			if p0.y - 1 == p1.y then
-				-- Turn into full block if pointing at a existing slab
-				if n0_is_upside_down  then
-					-- Remove the slab at the position of the slab
-					minetest.remove_node(p0)
-					-- Make a fake stack of a single item and try to place it
-					local fakestack = ItemStack(recipeitem)
-					fakestack:set_count(itemstack:get_count())
-
-					pointed_thing.above = p0
-					local success
-					fakestack, success = minetest.item_place(fakestack, placer,
-						pointed_thing)
-					-- If the item was taken from the fake stack, decrement original
-					if success then
-						itemstack:set_count(fakestack:get_count())
-					-- Else put old node back
-					else
-						minetest.set_node(p0, n0)
+				-- combine two slabs if possible
+				if slab_trans_dir[math.floor(p2 / 4)] == dir then
+					if not recipeitem then
+						return itemstack
+					end
+					local player_name = placer:get_player_name()
+					if minetest.is_protected(pointed_thing.under, player_name) and not
+							minetest.check_player_privs(placer, "protection_bypass") then
+						minetest.record_protection_violation(pointed_thing.under,
+							player_name)
+						return
+					end
+					minetest.set_node(pointed_thing.under, {name = recipeitem, param2 = p2})
+					if not minetest.setting_getbool("creative_mode") then
+						itemstack:take_item()
 					end
 					return itemstack
 				end
 
-				-- Place upside down slab
-				param2 = 20
-			end
+				-- Placing a slab on an upside down slab should make it right-side up.
+				if p2 >= 20 and dir == 8 then
+					p2 = p2 - 20
+				-- same for the opposite case: slab below normal slab
+				elseif p2 <= 3 and dir == 4 then
+					p2 = p2 + 20
+				end
 
-			-- If pointing at the side of a upside down slab
-			if n0_is_upside_down and p0.y + 1 ~= p1.y then
-				param2 = 20
-			end
+				-- else attempt to place node with proper param2
+				minetest.item_place_node(ItemStack(wield_item), placer, pointed_thing, p2)
+				if not minetest.setting_getbool("creative_mode") then
+					itemstack:take_item()
+				end
+				return itemstack
+			else
+				-- place slab using look direction of player
+				local dir = minetest.dir_to_wallmounted(vector.subtract(
+					pointed_thing.above, pointed_thing.under), true)
 
-			return minetest.item_place(itemstack, placer, pointed_thing, param2)
+				local rot = slab_trans_dir_place[dir]
+				if rot == 0 or rot == 20 then
+					rot = rot + minetest.dir_to_facedir(placer:get_look_dir())
+				end
+
+				return minetest.item_place(itemstack, placer, pointed_thing, rot)
+			end
 		end,
 	})
 


sofar:
Replaces #807

sofar:
Fixes #1109

sofar:
@everamzah @shadowninja @paramat @webdesigner97 

Please test this, It was somewhat late when I wrote this but I think this is the proper solution with the minimal amount of code.

mods/stairs/init.lua
```diff
@@ -126,87 +129,35 @@ function stairs.register_slab(subname, recipeitem, groups, images, description,
...
-				-- Place upside down slab
-				param2 = 20
+				return itemstack
 			end
 
-			-- If pointing at the side of a upside down slab
-			if n0_is_upside_down and p0.y + 1 ~= p1.y then
-				param2 = 20
+			-- else attempt to place node with proper param2
+			minetest.item_place_node(ItemStack(wield_item), clicker, pointed_thing, node.param2)
```
Ekdohibs: So you can't place upside-down slabs if the node above hasn't a 20-23 param2?

sofar: You can place upside down slabs without issues. If you place any slab against the side of an already-upside down slab, it will also be upside-down.

You can also combine an upside down slab from below with another slab.

Ekdohibs: How do you place the first upside-down slab, then?

sofar: Can't. Need screwdriver once.

webdesigner97:
It's working fine for me, this is the behaviour I expected. Thanks for your quick reaction this problem, now I hope we can successfully merge this.

everamzah:
Tested and works as expected.  The only concern left is that using `set_node` and `item_place_node` bypasses the `register_on_placenode` callbacks (such as lowering satiation in hunger mod), which is present when using on_place.  I think this doesn't matter so much, though, if it is the case.

:+1:

Fixer-007:
Will be nice to have this.

PilzAdam:
:-1: if it's not possible to place the first slab upside-down without screwdriver.

sofar:
@pilzadam I'll see if I can add that.

Do you expect the player look direction to be the deciding factor for the slab orientation, or the orientation of the selected face?

0-afflatus:
> Do you expect the player look direction to be the deciding factor for the slab orientation, or the orientation of the selected face?

Personally, the orientation of the selected face.

sofar:
@pilzadam added placement "to" a face. 

If under the placement face is an existing slab, the slab is placed the same as the previously placed slab. 

But if the node pointed under is something else (a normal node), then the slab will be oriented appropriately. This allows upside down placement, or placement against a wall.

Tested and slab combining and extending slabs all work fine. Please give it a try.

mods/stairs/init.lua
sofar: Fixed in a push to allow upside down placement and placement against walls.

sofar:
@pilzadam, @Ekdohibs @ShadowNinja @paramat @sfan5 please review, all comments are addressed, including placing the first slab upside down without screwdriver.

Fixer-007:
Hmm, I may have some problems with it:
slab on slab (for some reason it is not placing it properly)-> https://i.imgur.com/AohJDvy.png
stair on slab (not sure if related but trying to place stair on that slab from different angles, and it stands the same like this -> https://i.imgur.com/TXCEyTx.png

sofar:
> Hmm, I may have some problems with it: slab on slab (for some reason it is not placing it properly)-> https://i.imgur.com/AohJDvy.png

on purpose. You're placing an slab on an upside-down slab, so it copies the orientation.

I could try and make an exception for this, but it would result in very messy and lengthy code.

> stair on slab (not sure if related

that's basically the same as the above issue.

Looking into the door problem atm.

sofar:
> Edit: found lua error, hold on a sec: Trying to place any kind of door between those slab-walls: https://i.imgur.com/tdnIbhG.png

This is an error in the Door mod. Fixed by #1119

Fixer-007:
Why is this still not reviewed?

sofar:
I think the only thing remaining is "placing a slab on top of an upside-down slab". Right now the placement orientation copies the existing stair/slab, while it should just place it normal side up. I think it can be added, it's one more exception case but not too complex.

sofar:
@pilzadam @Ekdohibs @paramat @sfan5 

Please review, all comments in the thread are addressed and I think this is complete.

paramat:
Ok.

Fixer-007:
```
<Fixer> sofar: lua error with pr 1111: http://pastebin.com/raw/XJyZ75tL
<Fixer> sofar: tried to add another slab to here: https://i.imgur.com/mVAIDAt.png
```

sofar:
Investigating... but right now I don't see what's happening there - what slab are you placing against?

everamzah:
Looks like mossy cobble.  Is it touched by xdecor/workbench?

Fixer-007:
Place mossycobble like on screenshot - place another one to the side of it...

sofar:
Seems to crash without this patch too? #1163

sofar:
Ahhh no, #1163 is only an issue for this PR.

sofar:
What is happening is that mossy cobble slabs are not craftable. There is no recipe item defined for mossy cobble so it is impossible for the code to combine mossy cobble slabs due to the fact that it has no idea what to combine it to.

I'm going to make it so that mossy cobble slabs are just not combinable, since they're an exception case.

Fixer-007:
I tested this PR and result is this: https://i.imgur.com/NYg5idc.png In most cases it was working nice, much simpler to place slabs and panels, very quick. There were some small cases were i put helper blocks to orient properly, but most of the time it is just works. :+1:

https://i.imgur.com/NYg5idc.png

sofar:
:+1: from myself.

mods/stairs/init.lua
```diff
@@ -129,86 +134,60 @@ function stairs.register_slab(subname, recipeitem, groups, images, description,
...
-						itemstack:set_count(fakestack:get_count())
-					-- Else put old node back
-					else
-						minetest.set_node(p0, n0)
+
+			-- place_node will in turn attempt to call on_rightclick()
+			return minetest.item_place(itemstack, placer, pointed_thing, rot)
+		end,
+		on_rightclick = function(pos, node, clicker, itemstack, pointed_thing)
+			local under = minetest.get_node(pointed_thing.under)
```
t4im: `under` isn't used afterwards

mods/stairs/init.lua
```diff
@@ -129,86 +134,60 @@ function stairs.register_slab(subname, recipeitem, groups, images, description,
...
+				-- combine two slabs
+				if slab_trans_dir[math.floor(p2 / 4)] == dir then
+					-- combinable slab?
+					if not recipeitem then
+						return itemstack
+					end
+					if minetest.is_protected(pos, player_name) and not
+							minetest.check_player_privs(placer, "protection_bypass") then
+						minetest.record_protection_violation(pos, player_name)
+						return
```
t4im: always return the leftover itemstack

mods/stairs/init.lua
```diff
@@ -129,86 +134,60 @@ function stairs.register_slab(subname, recipeitem, groups, images, description,
...
-					else
-						minetest.set_node(p0, n0)
+
+			-- place_node will in turn attempt to call on_rightclick()
+			return minetest.item_place(itemstack, placer, pointed_thing, rot)
+		end,
+		on_rightclick = function(pos, node, clicker, itemstack, pointed_thing)
+			local under = minetest.get_node(pointed_thing.under)
+			local dir = minetest.dir_to_facedir(vector.subtract(
+				pointed_thing.above, pointed_thing.under), true)
```
t4im: This would crash if `pointed_thing` was `nil` (see lua_api)
(assuming the previous redundant usage was already removed)

mods/stairs/init.lua
```diff
@@ -129,86 +134,60 @@ function stairs.register_slab(subname, recipeitem, groups, images, description,
...
+				elseif p2 <= 3 and dir == 4 then
+					p2 = p2 + 20
+				end
 			end
 
-			-- If pointing at the side of a upside down slab
-			if n0_is_upside_down and p0.y + 1 ~= p1.y then
-				param2 = 20
+			-- else attempt to place node with proper param2
+			minetest.item_place_node(ItemStack(wield_item), clicker, pointed_thing, p2)
```
t4im: Why not just return `minetest.item_place_node`'s return value? The following code seems redundant.

mods/stairs/init.lua
```diff
@@ -129,86 +134,60 @@ function stairs.register_slab(subname, recipeitem, groups, images, description,
...
+			if not minetest.setting_getbool("creative_mode") then
+				itemstack:take_item()
```
t4im: Why not reuse the return value of `minetest.item_place_node` instead?

t4im:
Why is this in `on_rightclick`? That seems extremely hacky. `on_place` would at least be limited to when you are carrying the right item, but this is called always, with anything, simply because it is available.

Some mods might choose to treat this node differently because of it, and it might cause a lot of bugs we cannot predict yet.
(most other `on_place` method pass through to this, which puts this at their mercy)

At the very least this makes usage unpredictable:  
Imagine a tool with right click functionality, that passes through `on_rightclick`, so that you still can use e.g. chests.
Right clicking a slabs would prevent tool usage without providing any type of feedback (like a formspec opening).

mods/stairs/init.lua
```diff
@@ -129,86 +134,60 @@ function stairs.register_slab(subname, recipeitem, groups, images, description,
...
+
+			if wield_item == node.name then
+				-- combine two slabs
+				if slab_trans_dir[math.floor(p2 / 4)] == dir then
+					-- combinable slab?
+					if not recipeitem then
+						return itemstack
+					end
+					if minetest.is_protected(pos, player_name) and not
+							minetest.check_player_privs(placer, "protection_bypass") then
```
t4im: Crash when placing into protected terrain, because `placer` is undefined.

Fixer-007:
Can you make it work for stairs and full blocks as well? For example, after I rotated some brick block, further blocks should follow its rotation...

paramat:
I feel that would be overcomplex.

I really need to review this PR.

everamzah:
It's really great!  It's a lot faster, and more combinations are possible.  Plus, right now there's this leftover slab that remains for a bit.  This doesn't suffer from that.

t4im:
> Plus, right now there's this leftover slab that remains for a bit.

That could simply be prevented by disabling placement prediction.
`on_right_click` doesn't do that by default, but that can be disabled for `on_place`, too with one more line.

paramat:
I guess this is a workaround for #1109 as the real issue there is an engine issue detailed in https://github.com/minetest/minetest/issues/4357

mods/stairs/init.lua
paramat: Yes, placer is undefined.

mods/stairs/init.lua
paramat: Yes.

paramat:
Seems good but t4im's line comments need addressing.

sofar:
Right, thanks for the comments, I need to address them. Will make some time for this.

mods/stairs/init.lua
sofar: will delete

mods/stairs/init.lua
sofar: fixed

mods/stairs/init.lua
sofar: I'll add a `nil` check for that.

mods/stairs/init.lua
sofar: agreed, also simplifies the code.

mods/stairs/init.lua
sofar: ack, fixed to `clicker`.

sofar:
Pushed a fix which addresses all the comments from @t4im (thanks).

I want to do a few quick tests on this - will get to that later today.

t4im:
The [question](https://github.com/minetest/minetest_game/pull/1111#issuecomment-229928493) why this is done in `on_rightclick` is still open.
I don't see what could be gained? It certainly has the potential to cause problems.

mods/stairs/init.lua
```diff
@@ -129,86 +134,58 @@ function stairs.register_slab(subname, recipeitem, groups, images, description,
...
+
+			if wield_item == node.name then
+				-- combine two slabs
+				if slab_trans_dir[math.floor(p2 / 4)] == dir then
+					-- combinable slab?
+					if not recipeitem then
+						return itemstack
+					end
+					if minetest.is_protected(pos, player_name) and not
+							minetest.check_player_privs(clicker, "protection_bypass") then
```
t4im: You should use `minetest.get_player_privs(player_name).protection_bypass` instead, if you just check one priv like this.
It is internally called by `check_player_privs`, too, which does some unnecessary set to list conversation, you don't need here.
`check_player_privs` also has a broken signature of return values (set expected and either string or list returned) and should probably be deprecated one day.

sofar:
Had to think about that for a bit, since it's been a while since I dove into this problem and patch myself.

If we'd rely on `on_place` entirely for slabs, placement prediction would place the second slab in the adjacent node, and so visually it would look very weird to begin with, since the new slab would be placed first, then the server would remove it and modify the target slab to a block.

I can't remember whether that was the only reason, though. Only thing left would be to convert it to `on_place` entirely and see whether it all still works.

mods/stairs/init.lua
paramat: > in the future

Why in the future, can this be corrected now?

paramat:
:-1: Thinking on this i now would prefer this to be done in 'on place' so that a function is not run on every rightclick performed in-game, even if 'on place' placement prediction makes something appear and disappear.

sofar:
I'll redo this and see if that can be done. The original code uses both as well, so when I wrote this patch originally I just didn't consider it to begin with. Give me some more time on this and I think I can work it out.

sofar:
Updated to entirely use `on_place()`. Needs a little bit more testing.

All the basic behavior is the same, with one exception:

Due to this now entirely relying on `on_place()`, this means that slab combining only works if `pointed_thing.above()` is either `empty` or is `buildable_to`. This is because we can not avoid the mandatory check to see if `pointed_thing.above()` is empty. If it isn't empty, `on_place()` will never get called. (test case: place slab in bottom half of node. Place full dirt node above slab. Place slab pointing to top of slab placed in first step - this last step does nothing)

I find this somewhat awkward, but not prohibitive for the patch. Most people combing slabs will not run into this corner case.

paramat:
> (test case: place slab in bottom half of node. Place full dirt node above slab. Place slab pointing to top of slab placed in first step - this last step does nothing)

No problem i wouldn't expect this to work.

This makes some assumptons about how a player wants to build but i find how it behaves fairly intuitively and better than current behaviour.

screenshot_20160913_005858

Just one thing, placing on 2 sides of a column produces horizontal and then vertical lines, perhaps better to always have horizontal? As horizontal is usually wanted more and this also matches the horizontal lines of the column sides.

mods/stairs/init.lua
```diff
@@ -129,86 +134,61 @@ function stairs.register_slab(subname, recipeitem, groups, images, description,
...
-			-- If it's being placed on an another similar one, replace it with
-			-- a full block
-			local slabpos = nil
-			local slabnode = nil
-			local p0 = pointed_thing.under
-			local p1 = pointed_thing.above
-			local n0 = minetest.get_node(p0)
-			local n1 = minetest.get_node(p1)
-			local param2 = 0
+			if wield_item == under.name then
```
paramat: How about
```if under and wield_item == uunder.name then``` in case an unknown node is 'under'? Pedantic i know but might avoid a crash (?).

paramat:
Line comment.

paramat:
Otherwise looks good.

t4im:
> This is because we can not avoid the mandatory check to see if pointed_thing.above() is empty. If it isn't empty, on_place() will never get called.

Ah, that explains a few things. That underlying behavior also prevents for example combining homedecor-style fences or handrails to corners on placement,  because you usually point pass the nodebox during placement at the nodes beneath them (Contrary to when placing slabs, so this is probably indeed a smaller issue here.)

Perhaps an item definition flag might help improve this in the future some time? To have `on_place` not ignore placement of certain items. (Although i don't think this needs to delay this PR)

mods/stairs/init.lua
sofar: no, it's entirely possible (pointing at an unknown node). So, I'll fix.

sofar:
Fixed up one comment by @paramat.

paramat:
b848e35ca59c9eea3dbe3df87cb14d6bfea93f9a

mods/stairs/init.lua
t4im: Can, just didn't seem important enough to stale for it.

PR #1116, opened 2016-05-29
Make a ton of crafted blocks rotatable.

sofar:
Allow many crafted nodes to be rotated in any way possible.

These blocks all have slab and stair versions, which can create awkward
patterns if placed together. By allowing these to be rotated players
can create new patterns and appearances that were not before possible.

Since this wasn't possible before, there won't be any effect
to existing builds, as param2 should always be '0'. The current
screwdriver mod also refuses to rotate and alter param2, so this is
safe to enable from now on.

Personally, since these are all *crafted* nodes to begin with, it
should be apparent that they can be rotated to begin with, but I can
see people may disagree from a simplicity perspective. It also may
affect param2 usage that other mods rely on, although I'm not aware
of any mods that do this.

Addresses #1113.

### DIFF
--- a/mods/default/nodes.lua
+++ b/mods/default/nodes.lua
@@ -202,6 +202,8 @@ minetest.register_node("default:cobble", {
 
 minetest.register_node("default:stonebrick", {
 	description = "Stone Brick",
+	paramtype2 = "facedir",
+	place_param2 = 0,
 	tiles = {"default_stone_brick.png"},
 	is_ground_content = false,
 	groups = {cracky = 2, stone = 1},
@@ -236,6 +238,8 @@ minetest.register_node("default:desert_cobble", {
 
 minetest.register_node("default:desert_stonebrick", {
 	description = "Desert Stone Brick",
+	paramtype2 = "facedir",
+	place_param2 = 0,
 	tiles = {"default_desert_stone_brick.png"},
 	is_ground_content = false,
 	groups = {cracky = 2, stone = 1},
@@ -252,6 +256,8 @@ minetest.register_node("default:sandstone", {
 
 minetest.register_node("default:sandstonebrick", {
 	description = "Sandstone Brick",
+	paramtype2 = "facedir",
+	place_param2 = 0,
 	tiles = {"default_sandstone_brick.png"},
 	is_ground_content = false,
 	groups = {cracky = 2},
@@ -268,6 +274,8 @@ minetest.register_node("default:obsidian", {
 
 minetest.register_node("default:obsidianbrick", {
 	description = "Obsidian Brick",
+	paramtype2 = "facedir",
+	place_param2 = 0,
 	tiles = {"default_obsidian_brick.png"},
 	is_ground_content = false,
 	sounds = default.node_sound_stone_defaults(),
@@ -439,6 +447,8 @@ minetest.register_node("default:tree", {
 
 minetest.register_node("default:wood", {
 	description = "Wooden Planks",
+	paramtype2 = "facedir",
+	place_param2 = 0,
 	tiles = {"default_wood.png"},
 	is_ground_content = false,
 	groups = {choppy = 2, oddly_breakable_by_hand = 2, flammable = 3, wood = 1},
@@ -539,6 +549,8 @@ minetest.register_node("default:jungletree", {
 
 minetest.register_node("default:junglewood", {
 	description = "Junglewood Planks",
+	paramtype2 = "facedir",
+	place_param2 = 0,
 	tiles = {"default_junglewood.png"},
 	is_ground_content = false,
 	groups = {choppy = 2, oddly_breakable_by_hand = 2, flammable = 3, wood = 1},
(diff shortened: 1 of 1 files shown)


sofar:
Feel free to mark this controversial :)

Fixer-007:
Finally! Good PR, I was waiting for this for a long time, especially those pesky slabs.. rotatable will make my build look nicer.

lightonflux:
Thank you.

Fixer-007:
Yes, looks good.

sofar:
obsidian, stone, desert stone are omitted purposely since they are natural blocks.

Straw doesn't look too bad as is.

I'm not touching rail or other non-cubic nodes in this PR.

C1ffisme:
I feel like this is a bit of overkill for a graphical bug that most of us have a tendency to ignore. Example:
screenshot_20160529_195521

yyt16384:
This will make it harder to place nodes that align up properly.

sofar:
@C1ffisme entirely different (and fixable) problem.

@yyt16384 yes, although I think the added freedom would outweigh that cost.

Ekdohibs:
Finally!
:+1:

paramat:
Seems good, just not stone cobble as those have non-directional patterns that look very similar rotated.

yyt16384:
@paramat I think it is not very good to assume their look with texture packs.

Also, this pr does not fix cases like this:

screenshot_20160530_180406

Another solution would be to add stairs with textures in another direction. It doesn't feel very right though.

paramat:
Good point, are many other cobble textures directional? My guess is that most are not particularly directional, cobble after all is essentially a mess of stones not as ordered as bricks. Not making them rotatable allows faster building with them. So i still think stone cobble should be removed from this PR, maybe desert cobble too since it is to some extent a mess of different directions.

The general stair texture issue doesn't bother me because having stair bricks in a different orientation looks good. If it's that important to have the stairs matching then rotate the brick nodes to match them.
I feel the issue shown in the screenshot is not enough of a problem to justify alternative stairs with rotated textures.

So my opinion is, for now, remove all cobbles from this PR. They could always be added later.

0-afflatus:
This isn't a graphical bug fix, it's a gameplay decision.
Having rotatable nodes allows for greater creativity - texture directions can be matched or deliberately differed. Many wooden buildings want vertical planks, for example.
The con is that you have to be much more precise about placement direction - and often in tight corners it is impossible to get right, so you either end up screwdrivering a lot or less patient builders (likely the majority) just end up putting down crazy-paving. The net result of the latter is a lot more messy builds in multiplayer worlds.
The pros outweigh the cons for me, but it would be great to be able to decouple placement direction from  rotateability for some nodes.

0-afflatus:
Also you can't and should not account for texture packs. This needs to work with default textures.

rubenwardy:
Does the direction you place a node decide its facedir? If so then :-1:
It should work exactly the same as before, but allow you to use the screwdriver if wanted.
I always disabled the code in moreblocks that made wood rotatable, as it made building things a pain.

yyt16384:
@0-afflatus
> Also you can't and should not account for texture packs. This needs to work with default textures.

I was just saying that making only desert cobble but not normal cobble rotatable because of default textures doesn't make sense. They should behave in the same way.

@rubenwardy
> I always disabled the code in moreblocks that made wood rotatable, as it made building things a pain.

Is this still the case now? The code is probably gone as I don't have this problem now.

C1ffisme:
@sofar It is the same genre ("Cracks don't line up") of problem though.

paramat:
> making only desert cobble but not normal cobble rotatable because of default textures doesn't make sense. They should behave in the same way.

Agreed.

Thomas--S:
To sum up the positions here:
Some people prefer the old behavior, as it aligns the nodes properly.
Some other people prefer the new behavior, as it allows more flexibility/creativity.

How about this:
If the node is placed and the user does *not* press the sneak key, the node automatically get the param2 value "0".
If the node is placed and the user does press the sneak key, the node gets rotated.

lightonflux:
> If the node is placed and the user does not press the sneak key, the node automatically get the param2 value "0".

Please don't use the sneak key for any more actions. It is confusing enough already (placing on chests, screwdriver, opening /command).

paramat:
No thanks, sneak key is irritating to use.

Thomas--S:
But the sneak key is already used for other building related things, e.g. placing leaves, placing blocks next to formspec-blocks, etc.

The "use" key (E) would also be ok for me.

rubenwardy:
Or place it with a param2 value of 0 always, and make the only way to rotate a screwdriver.

Thomas--S:
:+1: @rubenwardy 
This is a very good idea!

sofar:
@rubenwardy This has been something I've wanted in the form of a "placement" param2 value that can be hardcoded - i.e. when a player places the node, the param2 value is always the same, predefined value (0 or non-0).

https://github.com/minetest/minetest/pull/4177

This PR now requires this game change so that it can force `param2 = 0` for the affected nodes.

The nodes will all be placed normally with `param2 = 0`, but can all be rotated to any direction after placement.

yyt16384:
A temporary solution is to force a param2 in on_place. It probably won't have client side prediction though.

sofar:
@yyt16384 look at https://github.com/minetest/minetest/issues/4177

paramat:
I support sofar's approach.

paramat:
https://github.com/minetest/minetest/pull/4177 is merged.

rubenwardy:
No objections now :+1:

sofar:
Removed cobble, mossy cobble and desert cobble from this change, as requested (and I agree that they should just stay as they were).

paramat:
:+1:

paramat:
ba1ae07b4a7b59fd27614b3d93ddc7faca28e3cf

PR #1122, opened 2016-06-04
Update api.lua

camus14:
A little bug with the farming API that was giving me problems when registering a new plant.


yyt16384:
There is already #1115 which does this.

camus14:
Ok, sorry, didn't see it.

PR #1175, opened 2016-07-02
Sethome: Clean up loading I/O.

t4im:
It's faster and cleaner to read a complete file in and then operate on it via string operations, than having to read it number for number in a loop, causing unnecessary reads.
(The only reason not to read it all at once, would be if the file in question would be too big to fit in memory. But since it's stored there in the end anyway, there is no point to that. Even then you would rather use `io.lines`.)

A single iterated `string.match` per line is enough here.

Errors should never be thrown away, but printed to the log to aid problem solving.

### DIFF
--- a/mods/sethome/init.lua
+++ b/mods/sethome/init.lua
@@ -2,22 +2,16 @@ local homes_file = minetest.get_worldpath() .. "/homes"
 local homepos = {}
 
 local function loadhomes()
-    local input = io.open(homes_file, "r")
-    if input then
-		repeat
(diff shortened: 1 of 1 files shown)


tenplus1:
@t4im - you win: https://github.com/minetest/minetest_game/pull/669/commits/ab06c870c79d6145a2e8aa99fe786af7e417b0d7

t4im:
> @t4im - you win: ab06c87

Now you're just being silly.

t4im:
Covered by #669

PR #1191, opened 2016-07-12
Only register respawn callbacks if respawn is enabled

adrido:
Its only a very small optimization, but it is one.

I hope I followed the code style. If not, please complain :smile:

### DIFF
--- a/mods/beds/functions.lua
+++ b/mods/beds/functions.lua
@@ -173,19 +173,18 @@ end
 
 
 -- Callbacks
-
--- respawn player at bed if enabled and valid position is found
-minetest.register_on_respawnplayer(function(player)
-	if not enable_respawn then
-		return false
-	end
-	local name = player:get_player_name()
-	local pos = beds.spawn[name] or nil
-	if pos then
-		player:setpos(pos)
-		return true
-	end
-end)
+-- Only register respawn callback if respawn enabled
+if enable_respawn then 
+	-- respawn player at bed if enabled and valid position is found
+	minetest.register_on_respawnplayer(function(player)
+		local name = player:get_player_name()
+		local pos = beds.spawn[name]
+		if pos then
+			player:setpos(pos)
+			return true
+		end
+	end)
+end
 
 minetest.register_on_leaveplayer(function(player)
 	local name = player:get_player_name()


mods/beds/functions.lua
```diff
@@ -173,23 +173,23 @@ end
 
 
 -- Callbacks
+-- Only register respawn callbacks if respawn enabled
+if enable_respawn then 
+	minetest.register_on_joinplayer(function(player)
+		beds.read_spawns()
+	end)
```
SmallJoker: #1102 removes this callback.

mods/beds/functions.lua
```diff
@@ -173,23 +173,23 @@ end
...
+	
+	-- respawn player at bed if enabled and valid position is found
+	minetest.register_on_respawnplayer(function(player)
+		local name = player:get_player_name()
+		local pos = beds.spawn[name] or nil
```
SmallJoker: `or nil` doesn't make sense - why is it this way in the old code?

mods/beds/functions.lua
paramat: #1102 is merged.

rubenwardy:
Use git pull upstream master --rebase to make it so your PR is make on top of new changes.
You'll need to undo the two commits not by you, and then run the above command

```
git reset --hard 744b5e9
git log
# check that  	Only register respawn callbacks if respawn is enabled   is newest commit
git pull upstream master --rebase
# where upstream is the name of the remote pointing to minetest/minetest_game
# if it does not exist, run:
# git remote add upstream https://github.com/minetest/minetest_game.git

# Now it will tell you about merge conflicts
# Go to the file it mentions, and look for <<<<<<   and >>>>>>
# Fix the file so it is correct

git add .
git commit

git log
# check that the log is correct
# Should be the latest MTG log + your commit at the top

git push origin patch-1 -f
# where origin is the name of your github fork
```

adrido:
Thanks a lot rubenwardy :sweat_smile: 

but now I have another trouble, Im currently at the step git commit and the following colourful editor opens. 
git_editor
How is this program called? It does not show any name so I cant google how it works.
How do I write the commit message? WASD seems to control the blinking cursor.
How do I tell the program that I'm finished and submit the commit?
:confused:

rubenwardy:
That is VIM. You need to type :wq

rubenwardy:
Did you scroll down in that pic? There should be a commit message above all those lines starting with hashes

adrido:
nope :confused: but I think I was successful to reset all to the state as it was before. Ill try to rebase now again.

adrido:
Ha, I think its working now :laughing:  thanks a lot again rubenwardy

rubenwardy:
Yay :D 

Welcome to git

adrido:
The next time I do it like this: https://xkcd.com/1597/
Seems to be much simpler. :laughing:

paramat:
:+1:

sofar:
:+1: assuming it's tested - trivial IMHO

paramat:
d89bb69a22e7abe90c80e827e07ea20a2c633a35

PR #1202, opened 2016-07-18
Make trapdoor textures work as expected.

Thomas--S:
Until now, if the texture was not symmetric, the trapdoor textures looked impossible.
This PR fixes that problem. The top of the texture is now always on the hinge side of the trapdoor.

### DIFF
--- a/mods/doors/init.lua
+++ b/mods/doors/init.lua
@@ -600,8 +600,10 @@ function doors.register_trapdoor(name, def)
 		type = "fixed",
 		fixed = {-0.5, -0.5, -0.5, 0.5, -6/16, 0.5}
 	}
-	def_closed.tiles = {def.tile_front, def.tile_front, def.tile_side, def.tile_side,
-		def.tile_side, def.tile_side}
+	def_closed.tiles = {def.tile_front,
+			def.tile_front .. '^[transformFY',
+			def.tile_side, def.tile_side,
+			def.tile_side, def.tile_side}
 
 	def_opened.node_box = {
 		type = "fixed",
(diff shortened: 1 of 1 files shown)


paramat:
Please can you add before / after screenshots to explain?

Thomas--S:
I will provide the screenshots tomorrow. I hope this is ok for you.

sofar:
This should be OK - indeed they are now automatically rotated the wrong way, and I think this patch will improve it. :+1: if the screenshots confirm that.

Thomas--S:
Here is the screenshot:


I hope that this screenshot is good enough. If there are further questions, feel free to ask.

paramat:
I see it's to do with the handle :+1:

sofar:
Need to verify that the bottom texture also is correct (handle on the correct side). Looking at the code it should be OK, but it may help to verify this in game since it's easy.

paramat:
The bottom is shown in the screenshots. I'll check the backside of the open trapdoor though..

sofar:
I meant the bottom of the *closed* trapdoor - the handle should not be on the hinge side there as well.

paramat:
The closed trapdoor bottom is at the top of the screenshots?

Thomas--S:
The trapdoor in the left upper corner shows the bottom side of a closed trapdoor.

paramat:
In testing, the rear texture of the open trapdoor has the texture reversed left to right.
This is not noticeable very noticeable with your test texture, try testing with a texture that is not symmetrical left/right to see.

Thomas--S:
Thanks for the information. I will work on this problem in the afternoon.

Thomas--S:
@paramat The problem should be fixed now. Thanks for your help!

paramat:
Will test.

paramat:
:+1:

paramat:
72f4c6be483a52a1ad56c7de5a75f56938ef7db1

PR #1210, opened 2016-07-19
Changed fencegate open collisionbox

tenplus1:
Changed fencegate open collisionbox so that it doesn't interfer with falling nodes.

### DIFF
--- a/mods/doors/init.lua
+++ b/mods/doors/init.lua
@@ -720,7 +720,7 @@ function doors.register_fencegate(name, def)
 	fence_open.collision_box = {
 		type = "fixed",
 		fixed = {{-1/2, -1/2, -1/4, -3/8, 1/2, 1/4},
-			{-5/8, -3/8, -14/16, -3/8, 3/8, 0}},
+			{-5/8, -3/8, -14/16, -3/8, 1/2, 0}},
 	}
 
 	minetest.register_node(":" .. name .. "_closed", fence_closed)


paramat:
Problem is https://github.com/minetest/minetest_game/issues/1193#issuecomment-233013757

paramat:
"We could
1. Keep both collision boxes the same size. The node can fall to the ground. This looks colliding, unless the mesh is updated, but so will placing a solid node there right now
2. Add a placeholder node to the fence_gate. It can be made walkable, which will solidify the node. The opening fence needs to handle occupied spaces however."
@sofar 
Or until 2. we could do 1.

sofar:
If we fix the falling node problem, then this change is not needed.

tenplus1:
This change is tiny and stops falling nodes from glitching on an open fence gate, the whole point is to limit glitches UNTIL the main fix is provided...  Add this change already.

paramat:
Maybe i misunderstand but this PR makes the open collisionbox full height, which has this problem https://github.com/minetest/minetest_game/issues/1193#issuecomment-233013757

> That stops the node indeed from moving. But without there being a walkable node below, or the falling node being able to fall down to one, it won't solidify.

tenplus1:
That comment isn't right, if you make open gate full height it stops falling nodes on TOP of the gate which it's suppose to do (since it's in the way), but when the gate is closed the falling node continues to drop to the ground below...

paramat:
Indeed but does the node solidify into a normal node or remain as a stationary falling entity?

paramat:
Testing this PR, sand falling onto the closed gate transforms back into a node.

screenshot_20160724_174553

But in ths following 2 cases, while falling onto an open gate, the sand remains a stationary falling entity that can't be dug:

screenshot_20160724_174639

screenshot_20160724_174730

So this PR is no improvement, except visually removing the jittery motion.

PR #1222, opened 2016-07-28
fixed tiling issue

DonBatman:
Fixed a couple issues with tiling.


Rogier-5:
Original cobble to the right, new texture to the left.
cobble

Although the old cobble does not tile perfectly, I am not entusiastic about the new texture: it's much too regular. Please make it less regular. Or you could just fix the horizontal border, because that's really the problem with the existing texture.

Maybe you could also reword the commit message to be more specific about what you're changing.

paramat:
:-1: I find the cobbles too large and the crack is missing. Although the current texture has a horizontal line at node base that doesn't look too bad. See how we have just one cobble tiling horizontally and how this is enough to break the grid effect, perhaps you could make a very subtle change with just one cobble tiling vertically?

DonBatman:
I agree. Might try again later.

PR #1229, opened 2016-08-02
Bones: Iterate player inventory lists dynamically.

t4im:
Avoid hard-coded player inventory lists.
Expose `bones.player_inventory_lists` for mods to look up or change,
which player inventory lists are being dropped or placed into bones.

Addresses #1148, but without global callback registrations.
Unless more encapsulation is wanted?

### DIFF
--- a/mods/bones/init.lua
+++ b/mods/bones/init.lua
@@ -1,6 +1,8 @@
 -- Minetest 0.4 mod: bones
 -- See README.txt for licensing and other information.
 
+bones = {}
+
 local function is_owner(pos, name)
 	local owner = minetest.get_meta(pos):get_string("owner")
 	if owner == "" or owner == name or minetest.check_player_privs(name, "protection_bypass") then
@@ -159,6 +161,18 @@ local drop = function(pos, itemstack)
 	end
 end
 
+local player_inventory_lists = { "main", "craft" }
+bones.player_inventory_lists = player_inventory_lists
+
+local function is_all_empty(player_inv)
+	for _, list_name in ipairs(player_inventory_lists) do
+		if not player_inv:is_empty(list_name) then
+			return false
+		end
+	end
+	return true
+end
+
 minetest.register_on_dieplayer(function(player)
 
 	local bones_mode = minetest.setting_get("bones_mode") or "bones"
@@ -172,8 +186,7 @@ minetest.register_on_dieplayer(function(player)
 	end
 
 	local player_inv = player:get_inventory()
-	if player_inv:is_empty("main") and
-		player_inv:is_empty("craft") then
+	if is_all_empty(player_inv) then
 		return
 	end
 
@@ -191,19 +204,12 @@ minetest.register_on_dieplayer(function(player)
 	end
 
 	if bones_mode == "drop" then
-
-		-- drop inventory items
-		for i = 1, player_inv:get_size("main") do
-			drop(pos, player_inv:get_stack("main", i))
+		for _, list_name in ipairs(player_inventory_lists) do
+			for i = 1, player_inv:get_size(list_name) do
+				drop(pos, player_inv:get_stack(list_name, i))
+			end
+			player_inv:set_list(list_name, {})
 		end
-		player_inv:set_list("main", {})
-
-		-- drop crafting grid items
-		for i = 1, player_inv:get_size("craft") do
-			drop(pos, player_inv:get_stack("craft", i))
-		end
-		player_inv:set_list("craft", {})
-
 		drop(pos, ItemStack("bones:bones"))
 		return
 	end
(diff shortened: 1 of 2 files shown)


raymoo:
Originally my reason for suggesting callbacks was because you might have something "representing" inventory, but not actually being backed by an inventory. An example would be setting spells to use, where spells are normally represented as items. But I rethought this use case and it would work fine using normal inventories, so I don't have any use case that would work better with callbacks than with just an inventory list. I don't see the need anymore for a global callback table.

raymoo:
I just remembered a use case that requires a callback instead of just an inventory list to be specified. If having something in the inventory (e.g. armor) has an effect, you need to know when the item is removed so you can remove the effect. The best possible place to do this is when the item is removed, which cannot be done if the bones mod simply dumps the inventory into the bones.

t4im:
> you need to know when the item is removed 

That sounds like a generic player inventory callback job, not something for bones.
Bones isn't the only thing that can remove an item, and would only cover one single case.

raymoo:
I guess that's fair, I can open an issue asking for callbacks that run on any removal.

raymoo:
Actually, how would such callbacks work in the presence of set_list, get_list and friends? It's impossible to tell in general whether operations on inventory lists moved items in any sense, though it would be possible to perform a diff on the inventories on every set_list operation and see if things were removed or added. But that looks like it would make set_lists too expensive, doesn't it?

mods/bones/init.lua
```diff
@@ -214,21 +216,19 @@ minetest.register_on_dieplayer(function(player)
 	local meta = minetest.get_meta(pos)
 	local inv = meta:get_inventory()
 	inv:set_size("main", 8 * 4)
-	inv:set_list("main", player_inv:get_list("main"))
```
rubenwardy: are there any stats on how much slower it is to add per item compared to setting the main inventory first?

I imagine it'll be negligible

rubenwardy:
All seems okay. I'm not sure whether you should use ipairs or a for loop to iterate through the player_inventory_lists array. A for loop is twice as fast, but ipairs means you don't need to index the array [i].

This definitely DRYs the code out a little. @raymoo's concerns should be addressed, though

t4im:
> I'm not sure whether you should use ipairs or a for loop to iterate through the player_inventory_lists array.

That code runs only on player death. No point in premature optimizing for speed.
I intentionally stuck with the readable code here.

Raymoo's concerns can't be addressed as part of this PR or even MTG, they are out of their scope.
If the engine adds an "player inventory has changed" callback caller, bones can be update to comply with the api. Because a lot of inventory actions are bulk operations, they would probably need be called as an "end transaction" method.

mods/bones/init.lua
```diff
@@ -214,21 +220,19 @@ minetest.register_on_dieplayer(function(player)
...
-	for i = 1, player_inv:get_size("craft") do
-		local stack = player_inv:get_stack("craft", i)
-		if inv:room_for_item("main", stack) then
-			inv:add_item("main", stack)
-		else
-			--drop if no space left
-			drop(pos, stack)
+	for _, list_name in ipairs(player_inventory_lists) do
+		for i = 1, player_inv:get_size(list_name) do
+			local stack = player_inv:get_stack(list_name, i)
```
rubenwardy: what happens if you try to add an empty stack to a list, or drop an empty stack?

t4im: You mean something like `inv:add_item("main", ItemStack(""))`?
By design i'd imagine [nothing](https://github.com/minetest/minetest/blob/ae7fb9b5bea291f81a82912d42fc43b7e71702bf/src/inventory.cpp#L328).

I did see some dropped "empty" ItemStacks with the chainsaw the other day, so maybe there is some corner case, that however needs to be fixed engine wide then. What i'm assuming are empty ItemStacks, did not cause any further trouble, just some confusion.
handdrop

rubenwardy:
:+1: although untested (only read through)
will test on merge

paramat:
@t4im hasn't been around for a while.

rubenwardy:
I'm willing to maintain this if changes are needed. There's no merge conflicts so it _should_ be fine

paramat:
@rubenwardy i doubt t4im will be working on this before merge.

tenplus1:
Would it be worth increasing bones inventory size to host extra items from other inventories ?

Ekdohibs:
:+1: (although untested as well); as for an increase of bones inventory size, it can impact gameplay a bit, so I'd like input from others as well on this point.

paramat:
@rubenwardy or anyone, this needs adoption as t4im has been non-active since October (hope he's ok).

paramat:
http://irc.minetest.net/minetest-dev/2017-04-23#i_4893324 sfan5: 	paramat: bones pr might be nice but doesnt seem to be pressing (+1)

paramat:
Approved but apparently needs testing @rubenwardy ?

paramat:
@rubenwardy please could you test?

paramat:
@rubenwardy ~poke~

paramat:
@raymoo rubenwardy is busy and i don't understand this, please could you test the PR?

paramat:
Thanks! Can be merged.

PR #1239, opened 2016-08-07
animated nc rainbow

HybridDog:


paramat:
I'd prefer a texture strip, so that the strip isn't re-created by code each time.

HybridDog:
>I'd prefer a texture strip, so that the strip isn't re-created by code each time.

What about the rotate modifier?

paramat:
I'll test your PR to see the effect.

paramat:
screenshot_20160814_102537

^ Has an issue.

This is okay but would be trivial to do as a texture strip, which is preferable.
It's also about twice too slow, compared to the original https://www.youtube.com/watch?v=QH2-TGUlwu4

paramat:
I just realised, animated textures are not synced from mapblock to mapblock, this is the reason.

HybridDog:
>It's also about twice too slow, compared to the original

there's a link to nyan.cat, where l got the speed from, it needs to be changed because l took it from the wrong rainbow.

>There may be a setting for syncing animated textures

https://github.com/minetest/minetest/blob/master/minetest.conf.example#L571-L573

paramat:
The setting is global unfortunately, can't change that.

HybridDog:
Imo asynchronism doesn't really matter.

paramat:
Lots of nyancats cross mapblock boundaries, around a quarter. It looks really bad and can't be allowed.

PR #1263, opened 2016-08-20
Transparent ice

hasanalsamra:
Transparent ice:
ice
ice2


DonBatman:
Awesome. I have been meaning to do a pr for transparent ice. Now I don't need to.

hasanalsamra:
Thank you Don, but what do you mean by pr?

Wayward1:
pr= pull request

hasanalsamra:
Thx

RHRhino:
Has the transparency issue already been fixed?

hasanalsamra:
I don't think so

paramat:
Unfortunately we cannot use translucency due to issue https://github.com/minetest/minetest/issues/95
It may seem to work at first but visual bugs will happen.

LazyJ:
I've had transparent ice on LinuxGaming for a couple years now since I added it to my fork of splizard's [Snow mod](https://forum.minetest.net/viewtopic.php?p=143688#p143688). It looks cool but suffers from Minetest's problem with translucent nodes. The transparent ice becomes invisible in water. The tops of the ice nodes are visible but the sides disappear. If completely submerged, the ice nodes become invisible altogether from most angles when viewed from \*above\* the water. When viewed under water \*with\* the ice node, then the ice node is visible.

From above water surface:
Translucent ice, under water, viewed from above.


From below water surface:
Translucent ice, under water, viewed from under water.

Translucent ice behind and around flowing water:
Translucent ice becomes invisible behind flowing water.

hasanalsamra:
Yeah unfortunately i have not noticed this, hope it solved in a future releases, should i close this pull request?

paramat:
Done.

hasanalsamra:
Thank you paramat, seems to be an old issue and hard to solve.

Fixer-007:
It is possible to solve, but nobody did it for now.

hasanalsamra:
yes sure i don't think it is impossible to solve

paramat:
It takes some rewriting of render code, quite a big job.

hasanalsamra:
it seems so

hasanalsamra:
but i cannot understand the problem

C1ffisme:
It seems like this would be better if it was tagged "needs engine change" rather than "Won't add".

I have hopes that one day someone will come along and fix transparency sorting.

paramat:
Agrred.

hasanalsamra:
+1

Ferk:
It also happens in Minecraft, you cannot see water through ice or ice through water.

hasanalsamra:
You are right i don't think it is a major bug, i made some screenshots,

From back the water is invisible:
screenshot_20160823_222607

But from inside the water itself you can see the ice:
screenshot_20160823_222622

hasanalsamra:
It is just a feature any way......

paramat:
In some situations though the bug can become weird, like distant objects appearing in front of closer ones, so it's not subtle even though it seems so at first.

hasanalsamra:
Thank you for your time paramat :+1:

PR #1285, opened 2016-09-11
Destroy flammable items when in fire or lava

Ferk:
This is a follow-up from #662 that seems abandoned.

I attempted to include some of the suggestions.

The check for igniters will be done every 10 seconds, non-flammable items won't check for igniters.

Once a node in the group `lava` is detected in this check, the item will be removed.

Once an igniter is found, the value of its `igniter` group combined with `flammable` group value of the item will be used to determine the chance for the item to disappear.

There isn't much visual feedback though, other than a smoke puff when the item is removed, but I did not want to add dependency on fire just to apply fire textures... and I'm also not entirely sure about the best way to add flames to a wielditem drawtype (more particles?). But since the item is most likely already inside a fire or a lava node I'm not sure if it would be really necessary.

Any thoughts/suggestions?


paramat:
I think this is too complex. If there's no noticeable difference between an item waiting to ignite and one that is burning i think having these 2 stages is unnecessary, it can be reduced to 1 stage.
Items are very rarely inside an igniting node and items can be large in number, so this feature needs to be very lightweight.
The original PR seemed good if it checked much less often, just once every few seconds, then a combination of 'flammable' 'igniter' 'lava' could perhaps be used to calculate the chance of it being removed and emitting a smoke puff (which is a good feature).

Ferk:
Yes, the burning status was mostly in case it was wanted some further effect, but if we are not gonna do that then it's not needed. If nobody is concerned with that then I'll rather remove the burning state.

Any suggestions on how should be the combination for the chance to remove it? I changed it so ```flammable*(igniter+lava)``` being 1 would have 1/2 chance, a value of 2 would have 2/3 chance, 4 would have 3/4 chance, etc.

There's still the check for the item being flammable. There's no point looking for igniters around items that are not gonna be removed. Or should non-flammable items be removed as well?

paramat:
How about making lava the only thing that can destroy non-flammable items?

Ferk:
Ok, I've done just that.
I also factored out the "burn_up" action to a separate function.

paramat:
Thanks will test.

mods/default/item_entity.lua
```diff
@@ -0,0 +1,64 @@
...
+						self:burn_up()
+					end
+				end
+			end
+		end
+	end,
+}
+
+-- set defined item as new __builtin:item, with the old one as fallback table
+setmetatable(item, builtin_item)
```
t4im: This is setting `builtin_item` as a metatable (as the function says), not as a fallback table.
Unless `builtin_item` has a self referencing Index-metamethod, which i assume it doesn't, this won't work.

mods/default/item_entity.lua
```diff
@@ -0,0 +1,64 @@
...
+local builtin_item = minetest.registered_entities["__builtin:item"]
+
+local item = {
+	set_item = function(self, itemstring)
+		builtin_item.set_item(self, itemstring)
+
+		local stack = ItemStack(itemstring)
+		local itemdef = minetest.registered_items[stack:get_name()]
+		if itemdef then
+			self.flammable = itemdef.groups.flammable
```
t4im: Consider `flammable == 0`, which will pass the line 37 test, too.

mods/default/item_entity.lua
```diff
@@ -0,0 +1,64 @@
...
+	on_step = function(self, dtime)
+		builtin_item.on_step(self, dtime)
+
+		if self.flammable then
+			-- flammable, check for igniters
+			self.ignite_timer = (self.ignite_timer or 0) + dtime
+			if self.ignite_timer > 10 then
+				self.ignite_timer = 0
+
+				local node = minetest.get_node(self.object:getpos())
```
t4im: You could optionally use `get_node_or_nil` for an early return, as the following code (with all its table lookups) won't be useful for items in unloaded areas anyway.

mods/default/item_entity.lua
Ferk: All registered entities have the `__index` metamethod set up, [it's part of `core.register_entity`](https://github.com/minetest/minetest/blob/stable-0.4/builtin/game/register.lua#L98), which is used before in builtin for the builtin_item.

mods/default/item_entity.lua
Ferk: Good point. Done, thanks.

mods/default/item_entity.lua
Ferk: For some reason I assumed both `nil` and `0` evaluated to `false` in Lua, but it seems only `nil` does. Fixed.

mods/default/item_entity.lua
t4im: Ah, great then :+1:

mods/default/item_entity.lua
```diff
@@ -0,0 +1,67 @@
...
+	end,
+
+	burn_up = function(self)
+		-- disappear in a smoke puff
+		self.object:remove()
+		local p = self.object:getpos()
+		minetest.add_particlespawner({
+			amount = 3,
+			time = 0.1,
+			minpos = {x=p.x, y=p.y, z=p.z},
```
paramat: minpos = p,
?

mods/default/item_entity.lua
```diff
@@ -0,0 +1,67 @@
...
+			maxpos = {x=p.x, y=p.y+0.2, z=p.z},
+			minacc = {x=-0.5,y=5,z=-0.5},
+			maxacc = {x=0.5,y=5,z=0.5},
```
Ferk: I did it. Not sure why the code comments don't hide in the new github.

mods/default/item_entity.lua
```diff
@@ -0,0 +1,67 @@
...
+		-- disappear in a smoke puff
+		self.object:remove()
+		local p = self.object:getpos()
+		minetest.add_particlespawner({
+			amount = 3,
+			time = 0.1,
+			minpos = p,
+			maxpos = { x = p.x, y = p.y + 0.2, z = p.z },
+			minacc = { x = -0.5, y = 4, z = -0.5 },
+			maxacc = { x = 0.5, y = 5, z = 0.5 },
```
paramat: Sorry but please no spaces just inside the curly brackets.

paramat:
Please could you use the texture in tnt mod called 'tnt_smoke.png' and add it to default textures? The 'smoke puff' texture is really bad. Please can the smoke particles be larger too as i barey noticed them.

Fixer-007:
flint and steel will work too? I imagine case when I want to burn dropped items. Flint and steel won't create fire on nonwood blocks :/

paramat:
You could use flint and steel to add a permanent flame to a coalblock, then burn the items in the flame.

mods/default/item_entity.lua
```diff
@@ -0,0 +1,67 @@
...
+			self.ignite_timer = (self.ignite_timer or 0) + dtime
+			if self.ignite_timer > 10 then
+				self.ignite_timer = 0
+
+				local node = minetest.get_node_or_nil(self.object:getpos())
+				if not node then
+					return
+				end
+
+				-- Immediatelly burn up flammable items in lava
```
rubenwardy: Immediately

rubenwardy:
Looks good, I agree that the puff of smoke should be changed for a more noticeable effect

Ferk:
I increased the size of the puffs and they should travel now +8 nodes upwards.
Should there be a sound effect?

I think it would be best to replace the smoke_puff.png from the engine and change tnt to use that as well. It makes little sense to have two textures so similar. I created a PR for it minetest/minetest#4548

paramat:
Wow we duplicated PRs :)
No need for a sound effect i think, for now anyway. It wouldn't be heard inside lava and flames already make the burning sound.

paramat:
The 16x16 smoke puff will look much better for this use because these puffs are smaller than a node, the 32x32 puff is far too high resolution to look right.

How about making minpos = p, maxpos = p (because the starting point is not visible) and adding some y velocity variation so the puffs slightly seperate vertically?
I noticed you have not set min/max velocities? i feel these should be included even if zero.
It seems more suitable to have acceleration zero and vary velocity, as smoke tends to rise without accelerating either horizontally or vertically.

paramat:
I might have trouble with my engine smoke puff PR, and we might create something different for that use, so to speed this PR along it might be a good idea to add the 'tnt smoke' texture into default and call it 'item smoke'?

rubenwardy:
> add the 'tnt smoke' texture into default and call it 'item smoke'?

Sounds good. If you make it a different texture to the tnt one, then it would be possible for texture packs to texture them differently - although that's only a minor advantage, it's not a bad thing if they share

Ferk:
Sorry for the delay

> How about making minpos = p, maxpos = p (because the starting point is not visible) and adding some y velocity variation so the puffs slightly seperate vertically?

The starting point is visible, it results in separation between the smoke particles that is random but keeps constant as the smoke rises.

But I agree it makes not that much difference so I've removed the initial Y separation. I don't think it's good to add much velocity variation in Y, though. It looks better if the dissipation is mostly on the XZ pane.

> It seems more suitable to have acceleration zero and vary velocity, as smoke tends to rise without accelerating either horizontally or vertically.

That would cause the smoke to move laterally in a linear manner. The idea is that it should start up by moving vertically and only after a certain height start dissipating laterally.

I agree however that the initial Y velocity should be set and not accelerated, so I've changed that, but the lateral XZ movement should imho remain as acceleration only. I've increased it a bit to make it more noticeable.

> add the 'tnt smoke' texture into default and call it 'item smoke'?

Done. Also for now it's its own texture "default_item_smoke.png", not reused in TNT. I just did it myself in Gimp. Unlike the tnt_smoke texture this one uses alpha transparency, though.

Ferk:
Some screenshots of the smoke:

paramat:
> The min/maxpos variation results in separation between the smoke particles that is random but keeps constant as the smoke rises

Actually this seems a good idea now, i don't mind this being re-added.

> but the lateral XZ movement should imho remain as acceleration only

Yeah that's also fine.

> Unlike the tnt_smoke texture this one uses alpha transparency

We can't use translucency due to https://github.com/minetest/minetest/issues/95

I'll test this again, i'll probably +1 when translucency is removed.

Ferk:
> We can't use translucency due to minetest/minetest#95

Does that problem also affect entities?
The current 32x32 smoke_puff.png used in the engine uses alpha transparency too.

paramat:
I'm not completely sure it affects particles, https://github.com/minetest/minetest/issues/3761 suggests it might.
If the current smoke puff uses alpha it probably shouldn't.
How about for now not using alpha, then we can always update the texture later if it is discovered alpha is no problem?

Ferk:
The alpha channel was irregularly distributed so it looks much more solid without it. I've dug a few holes in the texture but not too many since then it started looking bad (more like a cobweb than a smoke puff). I guess it's ok like this:
'

Ferk:
Does PNG actually have support for any other kind of transparency that's not alpha channel? I'm not convinced that simply increasing the contrast in the alpha channel of the PNG (ie. only using 255 and 0 values) would make any difference.

The thing is that, unlike nodes, there's no `use_texture_alpha` setting for particles.
It's either you remove the channel entirely in the PNG and make the particles completely solid (which would make them square) or you have to use the PNG alpha channel which is anyway using tranlucency even if it's high constrast.

Ferk:
Even when removing the alpha channel completely the bug happens, see how the particles are not visible through water:

paramat:
Interesting. Ok, so particles have depth-sorting issues, this would explain disappearing raindrops in #3761 which should be re-opened.
The alpha channel is needed of course but for now best to keep alpha 0 or 255, your texture looks ok to me. Will test motion.

paramat:
There are 2 wys to do alpha, 'use texture alpha = true' and 'alpha = xxx', water uses the second, glass uses neither although it's texture obviously has an alpha channel. Does a 'use texture alpha' node hide particles?

Ferk:
@paramat I made a `use_texture_alpha=true` node before to test the tnt explosion and the tnt particles were not visible through it, so I guess both `use_texture_alpha` and `alpha` have the problem.

paramat:
Tested. All seems good, i just find the initial y velocity too high at 9m/s.

Ferk:
I reduced by nearly half the speed of the smoke and adjusted duration and acceleration accordingly.

I also added the `collision_removal = true` flag, without it indoor smoke would either go through the solid blocks of the ceiling or hit the ceiling and fall to the ground...

paramat:
Tested :+1:

yyt16384:
Can you add a setting to control this? Some people don't want their items to be destroyed by accident.

sofar:
At particle size 4-7, with texture size 16x16, the pixel size of the particle is up to 2.5 smaller than the rest of the artwork.

For consistency, I'd choose a fixed size for the particle and change the texture size accordingly: size 5 with 8px x 8px would do great.

paramat:
Yeah i agree with sofar, or maybe even make the 16x16 texture as large as a node? (or something inbetween these 2 suggestions?).

sofar:
Since this particle is shown when items burn up, and most items are about half the size of a node, I'd make am 8x8 texture myself.

But, if it artistically looks good and coherent, the size is not the critical factor.

paramat:
8x8 is ok with me.

Ferk:
Are you guys sure the resolution is inconsistent?
To me it looks much more inconsistent if I reduce it to 8x8

The items are already 1x16 and are fairly small, even at 16x16 the particles are lower resolution than the item texture.



Not saying that it looks bad, but it's not making the resolution more consistent.

Ferk:
I've committed a 8x8 texture for testing, I think it might be ok anyway. It wouldn't be a problem if you wanted me to revert it, though.

There's already inconsistent resolution between the item entity and the nodes. Particles have never followed any rule (tnt_smoke particle size even varies with the explosion strength). But that might not be a bad thing so long as it looks good, in particular for something so "particular" as particles.

paramat:
> For consistency, I'd choose a fixed size for the particle and change the texture size accordingly: size 5 with 8px x 8px would do great.

Could you use a fixed size and scale the puff to be half the size of a cubic node?
Texture looks fine to me.

Ferk:
Done
Without size variation it had lost part of its random look, so I added variation to the starting position to compensate. I also had to reduce the spread.

paramat:
screenshot_20161009_222145

Nice i think the puffs look better at matching resolution :+1:

sofar:
The smoke particles exist for a very, very long time. I'd prefer a much shorter length (2-3 seconds tops). They also travel really high as a result, and kind of stop near the vertical apoapsis, which looks a bit weird.

other than that, codewise this is fine.

paramat:
Yes that was my request, i was considering items thrown into deep lava, so asked for about 8 nodes of travel, to be seen.

paramat:
I agree the puffs slow down too much.

Ferk:
I've tweaked a bit more the values, this time the speed is 7m/s (middle ground between 5 which was said to be too slow and the previous 9 which was said to be too fast).

I've reduced the time to 2-3 seconds, the smoke still travels +8 nodes upwards (~12-14 to be specific).

I've adjusted deceleration so the longest living particle doesn't reach speed zero, though I still think the deceleration helps to give a more realistic movement making the trajectory describe sort of an asymptote.

Ferk:
Rebased (not sure why github says there's still a conflict) and reduced a bit the speed and deceleration so the movement is a bit more constant. The spread is now less noticeable, but I think it's ok, still feels smoky I think.

sofar:
This really, really needs a sound effect as well.

(also note it now has a conflict in `README.txt`).

Ferk:
To be honest, I'm not entirely sure what would be the most appropriate sound for it.
I added some hissing sound softer and with lower pitch than the one from the lava (edited based on [this](http://opengameart.org/content/steam-release-sounds))

sofar:
That's fairly reasonable for a sound. Can you attribute the author? e.g.  link to http://opengameart.org/users/bart

Ferk:
It's CC0 (public domain, basically) but why not.
Done.

paramat:
Sound and new puff motion both fine by me :+1:

sofar:
Still looks like particles go way too fast to me, and way too high.

This looks perfectly acceptable to me:

```
diff --git a/mods/default/item_entity.lua b/mods/default/item_entity.lua
index c0930f1..afaaaae 100644
--- a/mods/default/item_entity.lua
+++ b/mods/default/item_entity.lua
@@ -26,12 +26,12 @@ local item = {
                        time = 0.1,
                        minpos = {x = p.x - 0.1, y = p.y - 0.05, z = p.z - 0.1 },
                        maxpos = {x = p.x + 0.1, y = p.y + 0.05, z = p.z + 0.1 },
-                       minvel = {x = 0, y = 6, z = 0},
-                       maxvel = {x = 0, y = 6, z = 0},
-                       minacc = {x = -0.2, y = -1, z = -0.2},
-                       maxacc = {x = 0.2, y = -1, z = 0.2},
-                       minexptime = 2,
-                       maxexptime = 3,
+                       minvel = {x = 0, y = 2, z = 0},
+                       maxvel = {x = 0, y = 3, z = 0},
+                       minacc = {x = -0.2, y = -0.5, z = -0.2},
+                       maxacc = {x = 0.2, y = -0.6, z = 0.2},
+                       minexptime = 1.5,
+                       maxexptime = 2.2,
                        minsize = 5,
                        maxsize = 5,
                        collision_removal = true,
```

If you're worried about putting particles too deep under lava, you should find the node that's under air and put particles in those instead of putting them in 8-10 nodes deep lava. That should be a single `minetest.find_node()`.

Ferk:
The first version of my PR wasn't very different from that (I think it was even shorter).
I also thought that way it made more sense, but after the high-reaching smoke was suggested and I tested it, it stroke me as an interesting take, this way players can burn items for smoke signals, making chimneys, etc.
In addition to being the cheapest way to get it to come out of the lava, since particles are entirely client-side.
When indoors it should quickly disappear when it collides with the ceiling anyway, so the actual duration might even be shorter than your version that has slower ascending speed.

I don't mind changing it to the values you gave, but it would be good to hear more opinions. I don't want to be changing it back and forth all the time.

paramat:
Yeah i disagree with searchiing for a lava node under air, a long life particle is simpler and transfers some load to the client.
I think it's important for the puff to rise fairly high, roughly 8 nodes, makes it more fun and is only a small load on the client.
I'm ok with this as it is, any reduction in puff speed would be nice but i'm not insisting on it. We can always tweak later i think it's time to merge this. Thanks for your patience Ferk.

sofar:
@ferk you're the owner of the PR, your opinion counts, too.

I'm not opposed to making particles go up further, what I don't like is that they go up really fast - I feel smoke should puff up slowly and billow.

So if you want to make them go up higher, and potentially fill up a cabin with smoke, you should:

- make them last way longer (15-20 seconds)
- have a speed of ~ 1m/s upwards or so
- have no or negligable negative acceleration
- enable collisiondetection
- make smoke particles come out separately (longer spawner lifetime)
- don't remove on collision detection (so smoke fills up your cabin or cave!)

e.g. this looks nice to me, and still makes smoke go up 8+ nodes:

```
diff --git a/mods/default/item_entity.lua b/mods/default/item_entity.lua
index c0930f1..44452c3 100644
--- a/mods/default/item_entity.lua
+++ b/mods/default/item_entity.lua
@@ -23,18 +23,18 @@ local item = {
                })
                minetest.add_particlespawner({
                        amount = 3,
-                       time = 0.1,
+                       time = 1,
                        minpos = {x = p.x - 0.1, y = p.y - 0.05, z = p.z - 0.1 },
                        maxpos = {x = p.x + 0.1, y = p.y + 0.05, z = p.z + 0.1 },
-                       minvel = {x = 0, y = 6, z = 0},
-                       maxvel = {x = 0, y = 6, z = 0},
-                       minacc = {x = -0.2, y = -1, z = -0.2},
-                       maxacc = {x = 0.2, y = -1, z = 0.2},
-                       minexptime = 2,
-                       maxexptime = 3,
+                       minvel = {x = 0, y = 0.9, z = 0},
+                       maxvel = {x = 0, y = 1.1, z = 0},
+                       minacc = {x = -0.05, y = -0.01, z = -0.05},
+                       maxacc = {x = 0.05, y = -0.01, z = 0.05},
+                       minexptime = 8,
+                       maxexptime = 12,
                        minsize = 5,
                        maxsize = 5,
-                       collision_removal = true,
+                       collisiondetection = true,
                        texture = "default_item_smoke.png"
                })
        end,
```

Ferk:
@sofar I feel that's a bit too much in the other extreme, but testing it made me notice that the `collision_removal` is actually only effective when `collisiondetection` is also true (sorry, it's actually already in the docs but I didn't realize). In the previous state of the PR the smoke was actually going through the ceiling even thought `collision_removal` was true.

Probably it would be best to use `collisiondetection` without the removal, like in your patch. I noticed that after properly enabling the removal the particles were disappearing even when they were just slightly touching the lateral wall, which I would prefer to avoid.

With a very low acceleration I think  it's ok for the smoke to not disappear indoors immediately, but then I think a duration of 8-12 might be a bit too long.

I slowed them down to 2.5 and increased the duration to 4-6, I think this might be a good compromise. My previous attempt was actually a compromise in the opposite direction, as I thought you were asking for shorter duration and that had to translate to higher speed for the smoke if we still wanted 8+ height.

About the longer spawner lifetime, the idea was to spawn the 3 particles at the same time with small random displacement to create one single clumpy cloud. It also matches the sound effect better if the smoke comes simmultaneously.
There will be more smoke if there are more itemstacks anyway (that makes me think.. maybe the number of particles and spawner lifetime could depend on the amount of items in the stack?).

paramat:
> Probably it would be best to use collisiondetection without the removal

I agree.

1m/s would be too slow, will test this latest version but i'm fairly sure it will be fine for me.

paramat:
:+1:

sofar:
I tried several values and especially in a house-like setting with a ceiling at 3 or 4 nodes high, 2.5 m/s was way too fast. 0.5m/s was a bit slow. I found 1 m/s just right.

paramat:
Shall we merge this so that Ferk isn't changing it back and forth due to our differing opinions? Then we can discuss and tweak it later ourselves.

sofar:
Noooo, we should jerk him around some more.....

Ok, in all seriousness, it's minor comments compared to the whole. The collision stuff is better now, so, sure, merge as-is for now :+1:

However, I think realistically this isn't going to get changed any time soon, at least that's my experience, but, that's not Ferk's fault ;)

paramat:
bcb4426ec784ad076ac3abb8d933f030079a0bce

PR #1314, opened 2016-10-06
Use travis' container-based infrastructure

sfan5:
- we can't use sudo there
- build times should be even faster

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,10 @@
 language: generic
-
-branches:
-  only:
-    - master
-
-sudo: required
-
+sudo: false
+addons:
+  apt:
+    packages:
+    - luarocks
 before_install:
-  - sudo apt-get update
-  - sudo apt-get install -y luarocks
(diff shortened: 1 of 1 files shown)


xunto:
Tested it with my game. It looks like it faster without those changes. 

Without:  Elapsed time 28 sec
With: Elapsed time 35 sec

I restated it several times but nothing has changed.

xunto:
Or you can look at [this](https://travis-ci.org/minetest/minetest_game/builds/165636368) and [this](https://travis-ci.org/minetest/minetest_game/builds/165364954).

sfan5:
It seems that it's mostly better if you need multiple builds quickly after eachother.

sofar:
:+1:

squash and merge, I'd say.

sofar:
This is too banal to let sit around. If this breaks travis, we can just revert quickly anyway.

sofar:
b0ae488 (squashed merge)

PR #1318, opened 2016-10-08
Merge boost_cart as "carts" mod.

sofar:
This is all the working code from SmallJoker's boost_cart, poored into
a more suitable form for minetest_game.
- Mesecons and moreores stuff was removed entirely.
- Textures were all renamed and moved out of default/
- Updated license, readme.txt, attribution
- Changed code license to MIT, left artwork at CC0
- removed default:rail and made aliases for it
- :carts:rail is now carts:rail.
- localized entity def
- removed copper rail entirely

The mod namespace is still public, but I'm NOT declaring it an API. I'd
rather see it localized instead, for now. Any public interface in this
code is _experimental_ at best, and should be considered non-stable and
unsupported for now.


sofar:
Possibly solution for #1311 

This is a work in progress. I've taken a fairly sharp axe to the code and removed quite a lot. The API is unstable and I may end up making more things private. There are possibly other changes we should consider (e.g. add vector.floor() to builtin).

I'd like @smalljoker and @pilzadam to sign off on the proposed license to make sure everyone is on board with the proposed relicensing (to MIT/CC0).

sofar:
Fixed luacheck errors.

Fixer-007:
Will play with it.

rubenwardy:
One annoyance I have with boost carts is it's decelerration. It decelerates by decreasing acceleration, this is inaccurate and feels wrong and ends up with oscillation of the cart. I'd much prefer it if rails could define friction using meta data, and carts slow down using loss of energy

```
local friction = (meta and meta:get_float("rail_friction")) or def.rail_friction or 0.1
```

sofar:
@rubenwardy - in physics, friction is proportional to kinetic energy (hence quadratic with speed), so it's not necessarily a good method for the handbrake, although it may be a good measure for the brake rail and feel natural there. I'd prefer coding friction in the rail nodedef so we don't need to :get_meta().

rubenwardy:
> in physics, friction is proportional to kinetic energy (hence quadratic with speed), so it's not necessarily a good method for the handbrake, although it may be a good measure for the brake rail and feel natural there.

yeah, it's mainly the brake rail where it feels wrong

> I'd prefer coding friction in the rail nodedef so we don't need to :get_meta().

The meta was there to allow mesecon brake rails - but that can be done by swapping the node

Fixer-007:
Did few test runs on world saves (including roller-coaster from ESM), seems good from the first sight, nothing unusual. Thing I noticed is when you use both this carts and boost_cart mod at the same time, there will be dummy textures for some rails.

paramat:
You probably know this, but all the 'dtc' rails can be removed from textures.
I agree with removng copper rails, super soft metal.

paramat:
Weird that the speed modifiers alter acceleration, surely these should be altering velocity?

paramat:
Yeah the mixup of velocity and acceleration needs sorting out.
Power and brake rails should add or subtract an amount to/from velocity.
Velocity should be reduced by a very small amount per step for natural air/rail resistance.
Handbrake should reduce velocity by a larger amount per step.

The 'startstoprail' could be renamed 'stop_rail' as the start action is a mesecons thing. The recipe could perhaps be combining 2 brake rails.

paramat:
> I'd much prefer it if rails could define friction using meta data, and carts slow down using loss of energy

That's too complex, all that's needed is a fixed and small reduction of velocity per step.

paramat:
Trying out the original mod, the velocity changes seems surprisingly good, perhaps because 'new acc' is being initialised to 0 on each step, then modified to a value and active until the next step. Strange way to do it but results in a fairly correct behaviour.

paramat:
Concerning textures, the mod rails don't match the default rail texture, also, the yellow and black pixels run under the sleepers, which looks wrong, they should be on top somehow. I could help out with textures.

paramat:
The gravity acceleration on slopes is too weak, i find this much better `local acc = dir.y * -4`

tenplus1:
Now is a good time to change recipe so that it crafts 12 rails instead of 24 which is far too many.

SmallJoker:
The main reason why I used a new name was because the cart handles the rails on a different way. There's actually not much left from the original carts mod so it's amusing me that it still gets renamed back.

@paramat - There's also the possibility to change the point of speed when a player can't accelerate the cart anymore -> `carts.punch_speed_min = 7` (which is actually wrongly named and should be `_max` instead)

For my part, I'm fine with the MIT license.

sofar:
@smalljoker I didn't rename it "back" per se - I didn't want to name it "boost_cart" since I didn't want it conflict with it. "rail" or "rails" also didn't sound like a good name, and "cart" didn't fit either (boneS, bedS, doorS, etc..). So, I just ended up with carts as the best name again :)

Ferk:
Imho, the acceleration stored in the meta should be moved to the nodedef.
Otherwise if new versions of the cart mod wanted to tweak the values every world would have to remove and place again their rails.

I don't think this is the right use for getmeta, since the acceleration is linked to the rail node and isn't really changed after placed (vs a sign or a chest, which are valid uses).

In addition, `get_meta` is likely to be slower than a lua table lookup, which wouldn't require access to the database or change of context.

I still think it would be best to have an `on_rail_step` in the nodedef of the rails, as this will also allow much more flexibility for modding. For a mod to implement the "detectrails" from the original boost_carts mod it would require overriding the cart entity with this current implementation.

Ferk:
Maybe it would be interesting if in the engine it was possible to define what nodes an entity is able to physically move though. So that client-side we could prevent movement prediction from making the cart appear to move past the rails, sometimes it even goes inside the ground in downward slopes.

stujones11:
Overall, this is pretty much how I imagined carts should be included though some interesting points and ideas there from @Ferk
I might suggest that you use a b3d cart model rather than directx, you are welcome to use this one if it helps https://github.com/stujones11/minetest-railnet/tree/master/railcart/models iirc, it should be a drop-in replacement but it might need some tweaking

mods/carts/functions.lua
```diff
@@ -0,0 +1,226 @@
...
+		return {x=carts:get_sign(v.x), y=carts:get_sign(v.y), z=0}
+	else
+		return {x=0, y=carts:get_sign(v.y), z=carts:get_sign(v.z)}
+	end
+end
+
+function carts:is_rail(pos, railtype)
+	local node = minetest.get_node(pos).name
+	if node == "ignore" then
+		local vm = minetest.get_voxel_manip()
```
stujones11: I am really not convinced that using the LVM is beneficial here, this should be profiled.

t4im:
I wonder, wouldn't `moveto(pos, continuous=false)` allow for a more reliable and fluent movement, as the client can accelerate and stop at the next target, without having to wait for server updates?

Future changes to the engine could even expand on that and perhaps use `param2` data to predict turns and slopes.

stujones11:
> use param2 data to predict turns and slopes

iirc, param2 does not work as you might expect for 'raillike' nodes. That is param2 does not change when a straight rail changes to a T or curved rail.

Ferk:
> wouldn't `moveto(pos, continuous=false)` allow for a more reliable and fluent movement

It probably would. But then I guess you would have to either recalculate the ending position every so often or blindly trust that the rail didn't change by the time you end the movement cycle.

stujones11:
The trouble with `moveto` is that you cannot control the velocity. The further you move, the faster you go. I did try to do something similar with my experimental railcart mod l linked earlier.

t4im:
Good point. max velocity would be required as parameter first.

param2 might not even be necessary for prediction. The client could take start and end point and do a client-side path finding along connected raillike nodes and then move along that path.

The client is already informed of map updates, which it could use to invalidate and recalculate.

But all that goes too far off the scope of this PR.

stujones11:
> But all that goes too far off the scope of this PR.

Agreed, the mod is good enough for now and can always be improved at a later date, possibly with the help of some engine changes. If this is to make it into version 0.4.15 (or whatever) then I think it should be merged sooner rather than later. That way there will be sufficient time to fix any potentially blocking issues before release.

Ferk:
> The client is already informed of map updates, which it could use to invalidate and recalculate.

The recalculation can't be client side since it might involve mod-specific logic. what if the update causes a power rail in the path to turn into a brake rail, how does the client know what to do in the power rail?

I think that's too complex, it'll still be simpler to have the entity enforce entities that can only move in nodes of a particular group. Which would also help for aquatic creatures, boats, etc. The end result will still be the same: the cart will stop (client-side) as soon as it reaches a boundary (a node it can't step into) and hasn't received instructions from the server to change direction.

paramat:
> There's also the possibility to change the point of speed when a player can't accelerate the cart anymore -> carts.punch_speed_min = 7

Yeah this would be good set to 4, walking speed, so there's no speed advantage over walking.

sofar:
@stujones11 I think I'm going to make it an .obj instead - since there are no animations, that should suffice. Yes I can take a b3d and export it to obj of course.

I *am* thinking about retexturing it as well. It is a bit too bland for my taste.

sofar:
@paramat I'd like to give a minor speed improvement over walking since otherwise the initial joy of "oh I've built my first railroad" is going to be massively underwhelming players. 6 or 7 should be just fine, I think.

paramat:
Yeah fine by me.

paramat:
6 is ok, i just don't want players travelling around fast by just punchiing and never having to use power rails. The movement by punching can be considered pushing the cart up to running speed, so much faster than our walk/run speed of 4m/s is not good.

stujones11:
> I think I'm going to make it an .obj instead

Then you will need to create three of them and switch meshes since entities do not currently have pitch

Megaf:
Now we have to make the player model turn with the cart and make the cart transition in curves smoother.
Player models turn with boats, UFOs and cars, so it should be possible to make that in carts too.

Fixer-007:
Notice: if you are riding cart and you get kicked/disconnected, you will spawn in last place, but obviously without cart, which keeps moving to end position.

sofar:
@megaf no, there's nothing wrong with merging large features that aren't entirely polished, this can be merged without player rotation, and then fixed later, probably.

Of course, I'd like to see if we can address the problem already now, as well, but there are already a lot of good comments that need addressing, so it's up to available time :)

sofar:
Note: I'm letting this sit for comments for a few days before looking (and fixing) some of the comments that have come up until now.

sofar:
@paramat how about, when you're in a minecart, pushing speed is max 6, when you're not in a minecart, pushing speed is higher (e.g. 8)? < random thought >

mods/carts/functions.lua
sofar: This code seems to trigger only when an area is unloaded (it's effectively a forceload) e.g. when a cart goes off the edge of the loaded map. I don't see another way of dealing with this than to do a LVM this way...

Not sure where node metadata comes from, that seems unlikely to solve the "map is unloaded" problem?

paramat:
Well, i'd make it the same for both cases (which is simpler too), i don't think shoving a cart from outside should result in a high speed.

paramat:
I just realised, a player wouldn't be able to move fast enough to keep up with the cart to punch it up to a high speed.

Fixer-007:
player can punch it when inside the cart, really neat feature.

mods/carts/functions.lua
stujones11: If the rails set a meta value on placement then you can call `get_meta` on unloaded positions.to determine whether it is rail or not.

stujones11: If rails were to set meta value on placement then you could use`get_meta` on unloaded positions to determine whether it is a rail or not. Probably not worth it as entities tend to become unloaded before the hit unloaded chunks.

stujones11:
> Now we have to make the player model turn with the cart

Iirc boats and ufos turn with look direction, carts do not. If you look at the player in third-person then you will see that it does in fact turn with the cart.

> if you are riding cart and you get kicked/disconnected, you will spawn in last place

Not sure what you could do about that other than stop the cart if the rider suddenly dissapears. Still there would be nothing to stop someone else taking the cart in the meantime. This is not really an issue IMO.

sofar:
The model rotates with the player, but not the player look direction. This is a bit more tricky, since we'd want to do relative rotation of the player look direction (add/subtract 90 degrees at turns). I can see why it's not added to the code yet, for that reason.

stujones11:
@sofar I am not sure forcing player look direction is a good thing even if it is possible.

paramat:
>  The model rotates with the player, but not the player look direction

This is fine, just like boats, don't bother with this feature.

> if you are riding cart and you get kicked/disconnected, you will spawn in last place, but obviously without cart

This is no problem, just like boats.

sofar:
So I'm going to start making a few changes to this patchset.

I'll remove startstop rail: The power rail nicely increases speed, the brake rail decreases speed. That's nicely symmetric and so the startstop rail is really not needed.

That removes a bunch of textures.

I'll likely use the b3d posted by @stujones11 

I'll add placement sounds, those are currently missing. I might add a new set of sounds entirely.

paramat:
The stop rail is very useful to stop the cart at a terminus or station, at an exact point, without having to experiment with various numbers of brake rails. The number of brake rails needed to bring a cart to a halt at an exact point is dependent on cart speed, so precise stopping will not be possible.
However the stop rail should allow a punch to get the cart moving again.

stujones11:
How about using Gambit's cart texture from the pixel box tp? Much nicer than the current one IMO.

paramat:
Would be nice to have 1 a pixel row of wheels showing.

sofar:
@paramat I prefer for now to have the simplicity. Also, ending the tracks stops the train at exactly the last rail, so the problem isn't terribly hard to solve by users.

sofar:
@stujones11 I'll definitely look at that texture, now that I know it exists :)

paramat:
Ok, sometimes stopping at point not at a terminus is good though.

sofar:
Updated model (used @stujones11 blend file to start, retextured with pixelbox' textures, and made sizes for pixels consistently the same)

Working on sounds right now.

paramat:
Looks good.

sofar:
I've removed the usage of `nodemeta` entirely. It's expensive to have to query the metadata and unlikely that nodemeta will ever contain something different than the original `acceleration` value of the rail when it was registered.

Instead, I'm keeping it not in nodedef, but in a private carts.railparams table. This makes lookups fairly easy, we just need to do ` minetest.get_node(pos).name` lookup, which is cheaper than a `get_meta()`.

I didn't want to put this in `nodedef` since that table can be huge and there's very little reason to put it there - a private table that only has rail nodes works just fine and works with `carts.register_rail()`

mods/default/aliases.lua
```diff
@@ -22,7 +22,7 @@ minetest.register_alias("papyrus", "default:papyrus")
 minetest.register_alias("bookshelf", "default:bookshelf")
 minetest.register_alias("glass", "default:glass")
 minetest.register_alias("wooden_fence", "default:fence_wood")
-minetest.register_alias("rail", "default:rail")
+minetest.register_alias("rail", "carts:rail")
```
SmallJoker: \+ `minetest.register_alias("default:rail", "carts:rail")` to keep the compatibility with older worlds

mods/carts/rails.lua
```diff
@@ -0,0 +1,68 @@
+minetest.register_node("carts:rail", {
```
SmallJoker: Could be registered with `carts:register_rail` aswell.
The default rail groups are always the same (see function `carts:get_rail_groups()`) so additional groups in the rail definition could be adapted to those in the register function.

mods/carts/functions.lua
```diff
@@ -0,0 +1,216 @@
...
+	}
+	for k, v in pairs(def_default) do
+		def[k] = v
+	end
+	if not def.inventory_image then
+		def.wield_image = def.tiles[1]
+		def.inventory_image = def.tiles[1]
+	end
+
+	carts.railparams[name] = table.copy(railparams)
```
SmallJoker: `railparams or {}`

SmallJoker: `railparams` should be an optional argument, so
```lua
if railparams then
     carts.railparams[name] = table.copy(railparams)
end
```

Ferk:
You could also keep a `registered_rails` table with references to the nodedefs is you don't want to use `minetest.registered_nodes`, this would be consistent with how other mods do it, like the `farming.registered_plants`.

mods/carts/init.lua
```diff
@@ -0,0 +1,355 @@
...
+		-- Slow down or speed up..
+		local acc = dir.y * -1.8
+
+		local node = minetest.get_node(pos)
+		local railparams = carts.railparams[node.name] or {}
+		-- no need to check for railparams == nil since we always make it exist.
+		local speed_mod = railparams.acceleration
+		if speed_mod and speed_mod ~= 0 then
+			-- Try to make it similar to the original carts mod
+			acc = acc + (speed_mod * 10)
```
Ferk: Why not make registered `railparams.acceleration` "5" directly in the definition, instead of using "0.5" and then multiplying it by 10 on every step?

Also, it's misleading. Someone trying to crunch the numbers for their rail might think 0.5 increases the acceleration in 0.5 m/s2 when it actually increases it in 5 m/s2

stujones11:
Looking good, my only remaining concern is how merging this might affect servers/worlds that are  using existing cart mods. I guess the boost_cart mod will have some part to play in that but what will happen for those using PilzAdam's original 'carts' mod?

mods/carts/init.lua
SmallJoker: That has its source in the compatibility with the original carts mod, where the meta string is used.
Due to different handling of the acceleration in the both mods, this * 10 multiplicator was added to get almost the same result.

SmallJoker:
@stujones11 I think there will be a forum topic about this, at the least when the players notice this conflict. (Not sure if it throws a warning, but it should IMO)

mods/carts/init.lua
```diff
@@ -0,0 +1,355 @@
...
+		end
+		-- Up, down?
+		if dir.y ~= self.old_dir.y then
+			vel.y = dir.y * math.abs(vel.x + vel.z)
+			pos = vector.round(pos)
+			update.pos = true
+		end
+
+		-- Slow down or speed up..
+		local acc = dir.y * -1.8
```
paramat: If you build some slopes and watch the acceleration by gravity i thin you'll find something around `local acc = dir.y * -4` is best.

mods/carts/init.lua
```diff
@@ -0,0 +1,355 @@
...
+		local railparams = carts.railparams[node.name] or {}
+		-- no need to check for railparams == nil since we always make it exist.
+		local speed_mod = railparams.acceleration
+		if speed_mod and speed_mod ~= 0 then
+			-- Try to make it similar to the original carts mod
+			acc = acc + (speed_mod * 10)
+		else
+			-- Handbrake
+			if ctrl and ctrl.down then
+				acc = acc - 1.6
```
paramat: Ths needs tuning, i tried `acc = acc - 4`.  Maybe it should be strong enough to stop on a long slope?

mods/carts/init.lua
```diff
@@ -0,0 +1,355 @@
...
+	if vector.equals(cart_dir, {x=0, y=0, z=0}) then
+		return
+	end
+
+	local punch_interval = 1
+	if tool_capabilities and tool_capabilities.full_punch_interval then
+		punch_interval = tool_capabilities.full_punch_interval
+	end
+	time_from_last_punch = math.min(time_from_last_punch or punch_interval, punch_interval)
+	local f = 3 * (time_from_last_punch / punch_interval)
```
paramat: Perhaps 2 instead of 3 so that punch doesn't become a too easy to use way to travel, and because carts are heavy and hard to push up to speed?

mods/carts/init.lua
```diff
@@ -0,0 +1,355 @@
+
+carts = {}
+carts.modpath = minetest.get_modpath("carts")
+carts.railparams = {}
+
+-- Maximal speed of the cart in m/s
+carts.speed_max = 10
+-- Set to nil to disable punching the cart from inside (min = -1)
+carts.punch_speed_min = 7
```
paramat: Perhaps 5 or 6?

mods/carts/init.lua
paramat: Indeed i was suggesting a handbrake strong enough to slowly come to a stop when going downhill. Useful but might be problematic.

mods/carts/rails.lua
sofar: I saw this and yes, I'll likely see if we can do that.

mods/carts/init.lua
sofar: I'll probably just eliminate the 10x compat factor.

mods/carts/init.lua
sofar: It appears to want to stop downhill on brakerails, but then the "backwards slide" bug happens and speed appears to reset. It's weird.

sofar:
@stujones11 so far, I've renamed everything so that if people are using mods, they won't see any conflicts (they'll have 2 of everything). If we add aliases, we can elminate the duplicate set, but that doesn't seem right (people might want the extra features from the mods).

I'm of the opinion that if we don't break mods, we should just let the mods decide whether they want to alias parts or not.

sofar:
@ferk I don't want to keep a list of nodedefs for the entire rail node, just the info needed to run rails. It'll be faster to do the hash lookups, and there's no need to duplicate the data in `registered_nodes` either.

sofar:
FYI I'm still pondering how to nicely do sounds, but it feels horribly bad without them, so, I'll have to think this over some more.

Fixer-007:
@sofar On Just Test II admin added some sound, check it out if you like it, maybe it is good one (and with good licence, who knows).

stujones11:
FWIW I have tested this mod extensively on my rpi based Karsthafen map subways and have not had a single glitch so far. I think this is good enough to merge as-is, sounds can always be discussed and improved later, just use the closest default sounds we have available for now.

sofar:
I posted a PR with default metal sounds. If reception of that is OK, we can merge that later or before, either is fine with me.

However, playing movement sounds is something I really want to add, since footsteps are not played when you ride a train cart, and so it's entirely quiet.

tenplus1:
@sofar - this may help, has sounds when riding: https://github.com/tenplus1/boost_cart

sofar:
@tenplus1 interesting, I just came up with something myself yesterday. It looks a little bit different but works roughly the same. The trick is in the sound sample itself, so I'll try the sound file in your mod.

paramat:
To reduce file size make sure the sound is 22050Hz sample rate, as lofi sounds suit MT more. Many of our sounds are 22050Hz.

rubenwardy:
As long as it sounds good - no tin-y sounds please

sofar:
```
 $ file */sounds/*|grep 44100|wc -l
45
 $ file */sounds/*|grep 22050|wc -l
10
 $ file */sounds/*|grep 48000|wc -l
12
```

So the majority seems to be 44100 or 48000 Hz, only 14% is 22050Hz.

22kHz files sound bad really quickly, so I prefer not to use them. For short samples just a second or so long, I think it's acceptable.

paramat:
Ok, i meant for any long riding sounds.

sofar:
It really bothers me that the boat interaction and the cart interaction are not the same:

- boats are pushed with `up`, slowed down with `down`
- carts are moved with `punch`, slowed down with `down`
- boats are picked up with `punch`
- carts are picked up with `shift+punch`

I'm a fan of the punch-to-move from inside the cart, actually, but not a fan of `punch+shift` to pick up. but more importantly, i *really* think that the interaction with boat and cart should be as similar as possible.

rubenwardy:
Maybe make boats punch + shift to pick up, then allow them to be pushed whilst a player is not inside them?

paramat:
> I suggest making boats punch + shift to pick up, then allow them to be pushed whilst a player is not inside them?

Not sure, shift+punch is necessary for carts because they need to be punched from outside, boats do not and shift+punch to pick up is a little irritating.

> I'd prefer up to be used when inside the cart

Me too.

sofar:
The problem with punching a cart to move is that when someone is inside the cart, they're effectively invulnerable.

If we allow carts and boats to be pushed, maybe it should not be the punch interaction?

We can't unfortunately really `push` the cart with our body, which would be ideal I think.

Ultimately up/down makes more sense from inside the cart than punching to me.

perhaps this is a good use case for the `use` key binding:

- `use` = enter/leave cart
- `up/down` = `basic movement
- `punch` = unaltered, damages things
- figure out how to push carts with player body?

paramat:
> The problem with punching a cart to move is that when someone is inside the cart, they're effectively invulnerable.

Seems ok, they're stting in a metal box. Currently AFAIK players are invunerable when attached to an object anyway.
Punch seem intuitive and ideal for moving the cart from outside.
I don't mind punch moving the boat too.

Ferk:
Maybe the cart could drop when hit with a pickaxe or so? and let any other hit push it.

stujones11:
>  what actually feels a bit awkward is shift+punch to pick it up.

I completely agree but cannot think of an acceptable alternative. Regarding up/down keyboard control, I can see how that might be desirable in a creative game. How about making it configurable and dependent of game mode?

tenplus1:
We already have right-click placing cart so why not on_rightclick to pick it back up again ?

paramat:
Nah please no configurable controls, too complex.
Another problem with 'up' to move is that 'autorun' could be used to cheat and get constant forward motion.

stujones11:
I think 'cheating' is a matter of opinion and I would not see that as a problem in a creative game. Also configurable need not be complex at all, it either reads keys or it doesn't, simple as that. However, personally I am quite neutral on this feature.

Ferk:
I think configuration would also be confusing since it's server-side, this would make the control inconsistent for the player depending on which server or gamemode he is on.

> why not on_rightclick to pick it back up again ?

But on_rightclick is already used to get in/out the cart.
I still prefer the alternative of it dropping when hitting it with a pickaxe or maybe some other tool.

tenplus1:
@Ferk - using on_rightclick to get out again would be android device friendly and make players pickup cart instead of them laying around world.

sofar:
Having carts lying around in the world shouldn't be an issue - they are non-physical so you just travel through them.

zorman2000:
I know I may be late to the comments, but I really would like to make a point on this mod. First of all, really big thanks to @sofar and @SmallJoker for the effort on these mods and for inclusion into minetest_game.

@sofar , I know you said that the startstop rail is not needed, but to me, this would solve a huge issue. Right now, I have a few train systems on my personal worlds, and I have ended with some complicated mesecon circuit to make boost_cart stop on a specific place and start again. I know you say that both power and brake rails are symmetrical in the way they accelerate/decelerate the cart, but how would it work over large distances? Let me explain better.

If the physics of the original mod are not heavily changed (I have not tested but from conversation I assume so) then this means that, if I start a cart from zero speed using, say, 4 power rails, at some point, without more power rails in the way, it will stop if I don't push the cart anymore. If I want to build a cart that goes from location A to B without a lot of punching, I would have to set some recurrent power rails between A and B terminals to keep the cart going. Which means, in the end, that I just **can't** say for sure the cart will stop with 4 brake rails at exactly the rail I want it stop.

The startstop rail is something that would solve this problem perfectly, because it means it will stop no matter what speed it will have. Even if the start is not automatic, at least it solves the problem of stop and is great for building exact stations. Therefore, I ask, please include the startstop rail, at least in the stop form.

Thanks! And thank you for your work!

sofar:
@zorman2000 currently the merged code has no (zero, nothing, nada) support for mesecons.

I intend to make the code as neutral as possible:

- not do things that other mods should do (e.g. mesecons signals or turn off/turn on)
- not add lots of rail variants that only add marginal extra features (e.g. a rail that accelerates, a rail that brakes, a normal rail, and that's it. Anything else is just a variant of those 3 rails)
- Either provide an API or be entirely removable by mods (e.g. the existing boost_cart should stay functional and should be able to remove the mtg rails and cart).

So yes, there won't be a way to stop a cart perfectly on a spot except *ending the rails*. Sorry. Having precise rails like that is something I consider terrain for mods.

paramat:
I'd like a simple stop rail too. Obviously not a mesecon one. Punch to start moving again.
It's good to have an automatic way to take a journey, and having a non-terminus auto-stop is useful, it is also a feature of the original and probably much used and wanted..

zorman2000:
@sofar Yes, I perfectly understand the lack of support for mesecons and actually agree about that. But as @paramat  said, I would actually like to see at least a simple stop rail. This would be extremely useful for building long railways.

Ferk:
I agree with the lack of direct support for mesecons but I think an `on_cart_step` would open up for mods to create their own start-stop rail or any other kind of rail (even mesecons if they wanted). It would also not be very expensive to add, when not used it's just an if test that is checked only once per rail.

    if railparams.on_cart_step then
        railparams.on_cart_step(node, self)
    end

Without something like this, the MTG carts mod is unlikely to be a boost_carts replacement in servers.

tenplus1:
@Ferk - If it helps here is the current boost_cart mod updated with the changes inside this pull, with riding sounds AND mesecon detector rails remaining and working.

zorman2000:
A simple API function such as @Ferk  described would be useful in this case, and is part of @sofar solutions. If a simple stop rail isn't really wanted, a basic API would help then us to create stop/detector rails. I would rather stick to use the default MTG carts mod than replacing it with something else.

stujones11:
While I would welcome some basic api features, I do not think this mod should even try to be a replacement for boost_cart. What is important is that it should not be made difficult for mods to extend or override it's functionality, @zorman2000 if you use the boost_cart mod then you are not replacing it with something else, only a more fully-featured version.

Ferk:
Sorry, when I said "replacement" I didn't mean that it should be a reimplementation of the same. I don't think the mod should have every single thing that boost_cart has.

But a mod that is simple can still be a replacement for a complex one if it has a better design and can be optionally extended. I don't think it would be as interesting to just have "yet another incompatible cart mod" that is bundled in MTG but that basically needs to be replaced so that "yet another rail api" can be done by other mod.

sofar:
My idea is to assure that `carts.register_rail()` works as an API for boost_cart to register start/stop rails and detector rails. I'm not sure if we ever need to disable the cart entity in minetest_game, but with overrides it should be possible to upgrade the cart and the default power and brake rails.

The last thing remaining is what to do about special rails that want to execute some handler. I think the only reasonable solution is a callback, fortunately we can pack that in the rail param table, and then handle in the cart `on_step()`. That would make all sorts of special rails possible (switchable direction, carts that fly off the track, detector rail for mesecons or even digilines, and the strangely popular start/stop rails)

Ferk:
> @Ferk - If it helps here is the current boost_cart mod updated with the changes inside this pull, with riding sounds, 10 second empty cart drops AND mesecon detector rails remaining and working.
> https://github.com/tenplus1/boost_cart

The detectorails don't get activated when the cart steps on them in that version, currently.

> My idea is to assure that carts.register_rail() works as an API for boost_cart to register start/stop rails and detector rails. 

I agree with that goal. The problem is that currently detector rails cannot be implemented with register_rail, unless the on_step of the entity is overridden.

zorman2000:
> The last thing remaining is what to do about special rails that want to execute some handler. I think the only reasonable solution is a callback, fortunately we can pack that in the rail param table, and then handle in the cart on_step()

This. This is perfect. As long as there's a callback, it means we can use the original MTG entity cart for everything without replacing the whole mod. If this gets in for release then I'm cool with the mod as it is. Thanks @sofar !

sofar:
Significant changes:

- lowered cart speed to 7 max and 5 pushed by hand
- added metal sounds (may conflict with metal sound PR, oops, need to fix this)
- event handler for rails (on_step in railparams)
- API documentation

sofar:
I did not update the control scheme. While I would like to, I don't see any consensus for a better scheme. If any, I think we should consider altering the boat control scheme instead perhaps.

Fixer-007:
dislike lower speeds... especially for powered one... travel will be much longer, with original to travel 10000 you need about 24 minutes... not it will be even more...

sofar:
@Fixer-007 agreed that I'm making it slower on purpose. Yes, players will not like this. It's not appropriate for minetest_game to assume players are travelling 10k on a regular basis, that's a multiplayer problem.

Besides, the max speed is tunable by a simple config, so setting it low by default is much more likely to make things a bit harder.

Compare this to the default boat speed as well. It needs to be in balance.

mods/carts/functions.lua
paramat: Running into ignore is highly unlikely anyway.

Fixer-007:
+ powered speed should be higher
+ when cart stops at powered rail (or any other?) - the sound of cart is still been played

mods/carts/init.lua
paramat: Krock approves. Also means more precise pushes if you just want to nudge the cart a short way.

mods/carts/init.lua
paramat: Needs increasing.

paramat:
Lower speed seems okay (but i will test), i almost suggested 8 earlier. Considering 'on step' is around 10-20 per second (?) it reduces chance of missed rail detection.

mods/carts/init.lua
paramat: * 10 needs removing.

mods/carts/init.lua
paramat: Needs to be more powerful, but perhaps just not quite enough to stop going downhill, to avoid the bug.

mods/carts/rails.lua
```diff
@@ -0,0 +1,59 @@
...
+minetest.register_alias("default:rail", "carts:rail")
+
+carts:register_rail("carts:powerrail", {
+	description = "Powered rail",
+	tiles = {
+		"carts_rail_pwr.png", "carts_rail_curved_pwr.png",
+		"carts_rail_t_junction_pwr.png", "carts_rail_crossing_pwr.png"
+	},
+	groups = carts:get_rail_groups(),
+}, {acceleration = 0.5})
```
paramat: 5

mods/carts/rails.lua
```diff
@@ -0,0 +1,59 @@
...
+})
+
+carts:register_rail("carts:brakerail", {
+	description = "Brake rail",
+	tiles = {
+		"carts_rail_brk.png", "carts_rail_curved_brk.png",
+		"carts_rail_t_junction_brk.png", "carts_rail_crossing_brk.png"
+	},
+	groups = carts:get_rail_groups(),
+}, {acceleration = -0.5})
```
paramat: -5

paramat:
Added line comments, which are not showing up here for some reason.

paramat:
Intentional that metal node PR has been combined into this?

sofar:
No, I can remove it but then it'll fail to load since I made it use default.node_metal_sound_defaults().

I'll either remove it or rebase... this pr has the old crappy sounds in it anyway.

mods/carts/rails.lua
sofar: yeah, I'll remove the scaling factor here for both of these...

paramat:
I see.

sofar:
It's not a bad time to fix bad UI interaction, but there's also nothing 
wrong with merging as-is and fixing it later.

sofar:
Rebased, since metal sounds are merged now.

sofar:
Merged https://github.com/SmallJoker/boost_cart/commit/b7412

Attempted to fix sound (need to handle sounds in exit conditions, otherwise sound doesn't stop looping).

sofar:
I would appreciate it if people can check if the sound-not-stopping issue is resolved with this latest version.

Fixer-007:
Will check

sofar:
The two "stuck cart" bugs reported above have been present in the code as long as I can remember. Perhaps @SmallJoker can comment.

mods/carts/init.lua
```diff
@@ -0,0 +1,396 @@
...
+		else
+			-- Handbrake
+			if ctrl and ctrl.down then
+				acc = acc - 1.6
+			else
+				acc = acc - 0.4
+			end
+		end
+
+		if self.old_dir.y == 0 and not self.punched then
```
SmallJoker: `self.old_dir.y ~= 1` to prevent this "jiggle"

mods/carts/init.lua
```diff
@@ -0,0 +1,396 @@
...
+	if self.old_pos then
+		-- Detection for "skipping" nodes
+		local expected_pos = vector.add(self.old_pos, self.old_dir)
+		local found_path = carts:pathfinder(
+			pos, expected_pos, self.old_dir, ctrl, self.old_switch, self.railtype
+		)
+
+		if not found_path then
+			-- No rail found: reset back to the expected position
+			pos = expected_pos
```
SmallJoker: `pos = vector.new(self.old_pos)` should fix the stuck carts

sofar:
Updated with @SmallJoker's fixes. Carts now stay "slanted" if they come to a halt on slopes, but they still wobble backwards....

Fixer-007:
Will check

Fixer-007:
* cart stuck when you punch it going uphill at low speed is **fixed**
* "fix" for brake rails has big negative side effect, I can't move off from such brake rail even furiously clicking at it, even starting on neighbouring simple rail and going through 1 brake rail is close to impossible.

Fixer-007:
freeze soon

paramat:
In 3 weeks, 7th December.

paramat:
Haven't tested sounds yet but i think they should be soft, like footsteps, not loud as in Minecraft, as long journeys would become annoying. Minecraft's approach is noisy, rattly old minecarts on bad track, i would prefer ours to be more like smooth-running trains.

sofar:
Sounds definitely need more work, or we should consider merging without sounds (maybe placement sounds only?)

paramat:
Merging without riding sounds is fine by me, i still need to test your latest version.

Fixer-007:
also don't forget about brake rail problem

paramat:
In testing i'm not keen on the riding sounds, i think no sound would be better than this.
Ideally it would be best to have a long sample that repeats, but that also cuts out quickly when necessary, not sure how to do this though.

paramat:
Note there are 6 comments of mine that need addressing.
Once we sort those out perhaps we can merge, without riding sounds, to get a couple of weeks testing before freeze?

paramat:
I'm working on making the textures consistent: use the new rail colours for the original rails. Will submit these later.

paramat:
https://github.com/paramat/rails
Here are the updated rail textures.
The new rail grey colours are used for the old textures..
I renamed the straight rails to 'rail_straight_ ..' so that all rails organise into groups of 3 in a folder.
2 rails were RGB, converted to indexed.

sofar:
Included 2 fixes:

- removed the stopping wiggle entirely.
- Updated all textures thanks to @paramat.

I'll work on the sound next.

mods/carts/init.lua
```diff
@@ -0,0 +1,413 @@
...
+		self.old_dir.y = 0
+	elseif vector.equals(vel, {x=0, y=0, z=0}) then
+		rail_sound(self)
+		rail_on_step_event(railparams.on_step, self, dtime)
+		return
+	end
+
+	-- stop cart if velocity vector flips
+	if self.old_vel and (((self.old_vel.x * vel.x) < 0) or
+			((self.old_vel.z * vel.z) < 0)) then
```
SmallJoker: What happens when you drive upwards and it should begin to roll backwards?

mods/carts/init.lua
```diff
@@ -0,0 +1,413 @@
...
+		else
+			-- Handbrake
+			if ctrl and ctrl.down then
+				acc = acc - 1.6
+			else
+				acc = acc - 0.4
+			end
+		end
+
+		if self.old_dir.y ~= 1 and not self.punched then
```
SmallJoker: The code within this `if` became useless due addition of https://github.com/minetest/minetest_game/pull/1318/files#diff-9812a7ca8951c9d004454c5c7c339f40R198

mods/carts/init.lua
```diff
@@ -0,0 +1,413 @@
...
+	local f = 3 * (time_from_last_punch / punch_interval)
+
+	self.velocity = vector.multiply(cart_dir, f)
+	self.old_dir = cart_dir
+	self.punched = true
+end
+
+local function rail_on_step_event(handler, obj, dtime)
+	if handler then
+		handler(obj. dtime)
```
SmallJoker: `s/./,/`

mods/carts/init.lua
sofar: removed. also updated `acceleration` constants by 10 to compensate.

mods/carts/init.lua
sofar: the bug is entirely gone as of the last version.

mods/carts/init.lua
sofar: Ok, setting this to `2`

mods/carts/rails.lua
sofar: fixed, as mentioned earlier.

mods/carts/rails.lua
sofar: fixed, as mentioned earlier.

mods/carts/init.lua
sofar: Got that fixed too, will push a fix in a sec.

mods/carts/init.lua
sofar: ew, bad!

sofar:
- api fix (typo)
- fixed push force (lower)
- carts going uphill too slow on a normal rail will now turn around and roll back down the hill unless they're pushed!

mods/carts/init.lua
SmallJoker: Yep.

Fixer-007:
BUGS!
* make simple straight rail, ride it, punch it once, wait before stop, now punch in opposite direction - it won't go.
* make simple rail running uphill to some flat area, start going up, but allow it to slowly roll back, start punching it again on slope when you going backwards, it will partially fall through rail and wobble.
* if you stand on brake rail - repeatedly punching it will not make you run, but punching once will, but with very slow speed and you will need to repunch.
* sound is cycled very fast? Why don't you play it fully?

paramat:
Perhaps use a sound timer that uses seconds instead of globalstep cycles, that restarts a long (2-3s) sample as the sample nears it's end. But of course 'stop' the sound instantly when necessary (velocity 0).
This means the velocity controlled gain will lag by up to 2-3s but i don't think that's a big problem since the riding sound is low volume.

sofar:
> make simple straight rail, ride it, punch it once, wait before stop, now punch in opposite direction - it won't go (critical bug).

Fixed this. Removing the old velocity when stopping the rail fixes this.

sofar:
> if you stand on brake rail - repeatedly punching it will not make you run, but punching once will, but with very slow speed and you will need to repunch, If you stand on big row of brake rail ... good luck punching through that (bad brake rail mechanics).

I don't mind this behavior. It's a brake rail - it should be hard to push a cart across it, and punching it every second or so (not fast) works fine.

mods/carts/init.lua
```diff
@@ -0,0 +1,414 @@
...
+
+		-- no need to check for railparams == nil since we always make it exist.
+		local speed_mod = railparams.acceleration
+		if speed_mod and speed_mod ~= 0 then
+			-- Try to make it similar to the original carts mod
+			acc = acc + speed_mod
+		else
+			-- Handbrake
+			if ctrl and ctrl.down then
+				acc = acc - 1.6
```
sofar: I kind of like the fact that it takes a while to stop - real trains don't stop on a dime anyway - it takes a lot of time for a train to halt.

sofar:
Much better sounds and sound handling.

I've moved the sound handling loop outside of the rail movement code so we can use real dtime and not a slice of it. This makes sound handling much more reliable. The sound samples are now 1.5 seconds long but we cycle them after 1.0, which leaves some room for lag. The new sounds are from an actual handcar, so they're quite authentic and appropriate.

Volume of the sound is tweaked - slow riding makes less sound.

Also added cart placement sound using default metal sounds.

mods/carts/init.lua
sofar: removed

sofar:
Removing the WIP flag. Maybe there are still some minor issues with it, but I'm not planning any major changes.

mods/carts/init.lua
paramat: Ok.

paramat:
screenshot_20161120_015115

Looks like the indexed alpha texture bug.
Also inventory / wield image for straight rail is the old texture.
New sounds are good, handcar is perfect.

mods/carts/init.lua
paramat: Trying this again something around -3 seems good.

paramat:
Perhaps i'll update the textures myself

paramat:
Fixed textures https://github.com/paramat/rails

paramat:
Well i'm completely bemused.
At least the green rails are gone, so it's worth using the RGBA textures.

mods/carts/init.lua
```diff
@@ -0,0 +1,403 @@
...
+		minetest.after(0.2, minetest.sound_stop, handle)
+	end
+	local vel = self.object:getvelocity()
+	local speed = vector.length(vel)
+	if speed > 0 then
+		self.sound_handle = minetest.sound_play(
+			"carts_cart_moving", {
+			object = self.object,
+			gain = (speed / carts.speed_max) / 2,
+			loop = true,
```
sofar: The overlap will help to make the sounds better under laggy servers.

sofar:
Pulled @paramats texture updates, and removed carts_rail.png entirely (substituted with carts_rail_straight.png everywhere.

However, I find the new rails to be very sterile. There is no color/brightness variation anywhere in the rail, which the old rail texture had. The sleepers are fine, but the rails themselves are not.

sofar:
I've reworked all the textures to incorporate the old rail style which had some minor rail color variation, and it looks much less sterile to me.

mods/carts/rails.lua
```diff
@@ -0,0 +1,59 @@
+carts:register_rail("carts:rail", {
+	description = "Rail",
+	tiles = {
+		"carts_rail_straight.png", "carts_rail_curved.png",
+		"carts_rail_t_junction.png", "carts_rail_crossing.png"
+	},
+	inventory_image = "carts_rail.png",
+	wield_image = "carts_rail.png",
```
sofar: Fixed

mods/carts/init.lua
paramat: Ok.

mods/carts/init.lua
paramat: I was considering acceleration on a slope, at 45 deg this will be 0.707 * g, 
`new_acc = vector.multiply(dir, acc)` It seems dir is not a normalised vector but can have a horizontal component 1 and a vertical componenet 1, so on a slope acc is multiplied by 1.414.
So acceleration should be set to 0.707 * 0.707 * g = 0.5 * g = 4.9.
But this is ignoring wind resistence and friction, and is almost equal to power rail acc, so how about
`dir.y * -4` as mentioned above?

paramat:
Just discovered that using mip map with anisotropic filter helps a lot with the disappearing sleepers.

sofar:
I'll revisit brake rail and downhill acceleration, it's indeed not pretty right now.

sofar:
I've reduced the brake rail strength and the power rail strength. These were way too powerful. I've also increased the gravity strength, making downhill acceleration stronger than the brake rail but also stronger than a power rail.

This results in the downhill issue being gone: a Brake rail won't be able to stop a cart on a downhill slope - it slowly accelerates, and this avoids the whole downhill jiggle issue.

As a result, now you can't push a cart up more than 3 nodes by hand.

Both these effects are IMHO desirable. We want to encourage players to add powerrails to get uphill, and have them think about slopes when they build.

paramat:
`local acc = dir.y * -4.0` is good.

I feel strongly that power rails need to be at least as powerful as gravity acceleration on a slope, to allow long climbs over large vertical distances. Due to how Minetest's vertical size long continuous slopes are essential.

With your latest change, power rails would be needed to reach maximum speed on the flat, then the slope would also need continuous power rails, but the maximum length of a slope would be limited to 6-7 nodes, with the cart slowing to a crawl near the end. This will be frustrating and impractical and remove a lot of fun.

If power rails are made +4 then with continuous power rails speed remains constant on an upward slope, the cart will continue to climb at the speed it which it hit the base of the slope, which seems reasonable behaviour.

I would prefer that brake rails are -4 to match the power rails, as they are matched currently.

I will +1 this PR if these changes are made.

sofar:
OK, I don't mind acceleration being slightly faster than what it is right now. Set at 4.

paramat:
:+1:

paramat:
1e691c442907d6bb5602bc34fdbb3c9a7436c223
Congratulations.

paramat:
I made some minor changes on merge:
Removed EOF whitespaces from 3 files.
Changed max speed stated in README to 7n/s.
Changed credits for rail textures to 'modfied by sofar' instead of by me.

sofar:
Thanks! Looking forward to fixing the bugs people find next ;)

PR #1322, opened 2016-10-11
3D Torches

octacian:
3D Torches

The title says it all. This PR makes torches 3D. Aside from using a mesh drawtype instead of torchlike, the only major difference is that you cannot place torches on the ceiling (like @BlockMen's [torches](https://forum.minetest.net/viewtopic.php?t=6099) mod).

The textures and models are from @BlockMen's [torches](https://forum.minetest.net/viewtopic.php?t=6099) mod, and much of the code I was able to write as a result of looking through the mod.

I am working on making torches drop when placed on ceilings (instead of simply staying in your inventory), adding particles, and torch decay (latter two optional).

**Note:** <s>three commits need to be squashed.</s> **EDIT:** I figured out how to squash already-pushed commits! Yay!


paramat:
We can easily squsah on merge if you can't squash.

Ferk:
+1 on the concept, but maybe particles and torch decay should be separate PRs. Better first just get the 3D torches right.

For example, you are missing an LBM or some way to get the old torches to properly get in position, so loading old worlds that do not have this change does not result in a weirdly perpendicular torch.

I seem to recall there were also some bugs discovered on sofar's variation of BlockMen torches. Perhaps that's happening for yours too?

octacian:
@paramat thanks..

@Ferk that's what I was thinking at this point. Torch decay will be a different PR for sure. Could you give me more information of these bugs? From my testing, I don't think there are any issues, however, I'd far rather be safe than sorry.

Fixer-007:
I think particles should be in this PR (it is just visuals)...

Ferk:
@octacian I think this might be one of the things, try displaying a toch in an xdecor itemframe:



I know xdecor is not an MTG mod but it's a popular one, i'd guess this has something to do with the on_place callback, maybe there's a way for it to not interfere with the action of other mods.

mods/default/models/torch_wall.obj
```diff
@@ -0,0 +1,48 @@
+# Blender v2.73 (sub 0) OBJ File: 'mt_torch_full_vertices3.blend'
+# www.blender.org
+mtllib torch_wall4.mtl
+o Cube_Cube.001
+v 0.061653 -0.554493 -0.328908
+v 0.061653 -0.442208 -0.381835
+v -0.059258 -0.554493 -0.328908
+v -0.059258 -0.442208 -0.381835
+v -0.059994 -0.948766 -0.143618
```
t4im: Okay... So you rotated the former geometry to keep it simple, i assume. But that vertice and the other face's V16(L20) are practically already in the middle of the node below. Are you sure about this?
Looks kinda odd now.

mods/default/models/torch_floor.obj
```diff
@@ -0,0 +1,48 @@
+# Blender v2.73 (sub 0) OBJ File: 'mt_torch_full_vertices3.blend'
+# www.blender.org
+mtllib mt_torch_full_vertices3.mtl
+o Cube_Cube.001
+v 0.061653 -0.496733 0.061674
+v 0.061653 -0.496733 -0.062460
+v -0.059258 -0.496733 0.061674
+v -0.059258 -0.496733 -0.062460
+v -0.059994 -0.497234 0.497315
+v -0.059994 -0.497234 -0.498102
```
t4im: Gap seems small enough to not be noticed; but have you considered having the torch actually stick inside the node a little bit here, like with the wall-torch? Or would that be too hard for avoiding z-fighting?

mods/default/models/torch_floor.obj
```diff
@@ -0,0 +1,48 @@
...
+v 0.062686 0.507706 0.497315
+v 0.062686 -0.497234 0.497315
+v -0.494900 0.507706 -0.063149
+v -0.494900 -0.497234 -0.063149
+v 0.497295 0.507706 -0.063149
+v 0.497295 -0.497234 -0.063149
+v -0.058217 0.134520 0.060216
+v -0.058217 0.135872 -0.061813
+v 0.061944 0.135872 -0.061813
+v 0.061944 0.134520 0.060216
```
t4im: Don't get me wrong. The vertice count isn't actually why i'm asking. It's just that i'd expect some weird graphical intersecting going on with some texturesets.

paramat:
I won't miss torches on ceilings.

t4im:
But what about ceiling-torches, that are already in the world?

sofar:
Is there any reason not to just take github.com/minetest-mods/torches?

- I specifically maintained it with the idea of making it ready to merge
- It has been cleaned out and made as minimal as possible
- It fits minetest_game style
- It has many bugs fixed
- Ceiling torch is working
- New model which looks better with default torch animated textures
- Retains texture pack compatibility
- The wieldlight needs to be stripped prior to merging (it's a nice feature, but requires wieldview to be useful, so I'd just delete that prior to merging).

And, much more important, this code is running on a few servers in production, so it's getting tested.

The only minor bug remaining is itemframes, but this isn't unsolvable (I just haven't put in any time).

sofar:
Nothing personal, but I've made an alternative PR that competes with this one: #1325.

I've sat on the code for 6 months and lots of people have helped me test it, so it would be silly not to at least have it compared.

octacian:
@sofar no specific reason. I didn't know @BlockMen's original mod was still being maintained in any way.

I've been trying to make it as simple in code as possible, to avoid less bugs, much like you've been doing. I've tried to follow MTG style, but I've never done any direct contributing to either MT itself or MTG so I'm not sure how well I did in that area.

Ceiling torches are something that I personally really dislike. That's why I don't have it here; most other people I've spoken with (not just @paramat) have disliked them as well, and wanted them removed. I'm also going to make the ceiling torches seemingly fall down, by dropping an item.

With the information @t4im has given me, I will fix the itemframe issue day-after-tomorrow (I'm away from home ATM). Possibly sooner, but I almost guarantee by then.

As you can see, I'm using both textures and models from the original mod, and personally am quite happy with them. I'll look into yours as well though. So far, my PR retains all texture pack support. I'll also be using fire and smoke particles, and I've got an idea to maintain texture pack support for those as well. I'm thinking about getting my friend to make a new model for wallmounted torches, with a type of holder making it seem more realistic, and look cooler.

Overall, I'd be reasonably happy with either PR being merged, but I think that there are some details that I've mentioned above that would make me prefer my work (particles, ceiling, etc...).

sofar:
My work was listed in the thread at #1308 which you commented on 2 weeks ago.

t4im:
Ah, the model is from Blockmen. Overread that before. But i still wonder, why that weird #-shape instead of a simple stick with lowered top. Any limitations with animated frames?

octacian:
@sofar I completely forgot about what you mentioned in that issue, sorry :D 

I've looked over your PR further, but I still prefer no ceiling torches, and particles.

octacian:
@t4im Not that I know of. But I know almost nothing about animations and UV mapping, so... You, somebody else, or @XtremeHacker would have to explain it to me.

t4im:
The current model isn't formed like a torch, just texture transparency hides that fact from you:

bm_torch_geometry

mods/default/nodes.lua
```diff
@@ -1534,13 +1512,56 @@ minetest.register_node("default:torch", {
...
-		wall_side = {-0.5, -0.3, -0.1, -0.5 + 0.3, 0.3, 0.1},
+		wall_top = {-1/16, -2/16, -1/16, 1/16, 0.5, 1/16},
+		wall_bottom = {-1/16, -0.5, -1/16, 1/16, 2/16, 1/16},
 	},
 	groups = {choppy = 2, dig_immediate = 3, flammable = 1, attached_node = 1},
-	legacy_wallmounted = true,
 	sounds = default.node_sound_defaults(),
+	drop = "default:torch",
+	on_place = function(itemstack, placer, pointed_thing)
+		local above, under = pointed_thing.above, pointed_thing.under
```
octacian: How do I check if the node under has an on_rightclick? IK how to get the node under, but I'm not sure how to check it's full nodedef. I tried `minetest.get_node(under)`, but that only returned it's name an 2 params. Maybe I need to look at the params, or there may be another table inside that I'm not seeing, but IDK.

t4im: `minetest.registered_nodes[name]`

Provides you a nodedef, but it can be `nil` if you're looking up an unknown node.

You could also use `minetest.get_node_or_nil()` here to get the name, which will return `nil`, if you right click an unloaded node. While not strictly necessary here, due to lack of node setting; it's still a bit cleaner imo.

t4im: Just take a look at 32f7051eb85bfe54ac21475be6366db7aff8743d :-)

sofar:
While I am a fan of adding particles, I think it should be a separate PR, and added independently whether this or the other PR is merged. I personally prefer to wait until client-side modding becomes available, which will make particle decorations like torch smoke much more efficient.

I'm against removing the ceiling torch: There is just no way that you can make that change smoothly (backwards compatible, not breaking any existing builds) on any existing server. Dropping potentially thousands of nodes as entities on servers and home builds is just going to annoy players. Yes, the ceiling torch is weird (to say the least), but it's there and we should keep it - I do see it used plenty on lots of servers, so players do appreciate it's style.

t4im:
The ceiling torch could probably be made to burn upwards with the same texture, but a different mesh; would that help?

sofar:
@t4im yes, but large parts of the flame would be inside the stick part, which is why I kept mine just upside down, for now. It would basically become a 3rd node with separate mesh if you'd really want to
fix it, and I didn't want to take it that far (for now).

paramat:
To be clear, i'm not asking for removal of celing torch, just that if it went i wouldn't miss it. Sofar's reasons for keeping it are good though, it probably should be kept.

XtremeHacker:
@octacian I don't know UV (Un)-Wrapping, yet...
On-topic I personally like upside down torches, for two reasons: 
1: places you can't put  torch on the side, 
2: for making chandeliers.
3: as @sofar mentioned, it could (and I'm sure would) break a lot of server, from lag caused by all the torches that would all of a sudden drop as entities.

octacian:
Ok. I've heard enough :P 

I'll keep ceiling torches. Once I'm back home, it'll take me about 2 minutes. 

This morning I got item frames working with the information @t4im gave me (thanks!). I'll push the changes once I'm home. 

I'm going to switch to @sofar's mesh as it will definitely be nicer. Sofar's code is near identical to mine (with the last two changes), so I'll use some of Sofar's code as well since its been tested and proven.

octacian:
Closing. See #1325 by @sofar.

It has undergone much more testing, and is overall better fit for many of the server scenarios. I still have to finish fixing the issue with itemframes (make the code simpler), possibly change the model, and re-enable ceiling torches. However, @sofar has already done all of this in his PR, so it's pretty much ready to go.

PR #1325, opened 2016-10-12
Merge 'torches' mod as default:torch.

sofar:
This merges the current state of the well-maintained and tested
`torches` mod as I've maintained it for the last 6 months.

This started out as a thorough cleanup of 3D torches by blockmen,
where some of the initial code still remains.

The models were redone entirely from scratch and have been extensively
tested with dozens of animated textures converted with mcimport,
and look a lot better than the original 3D Torches mod.

The ceiling torch is retained and functional. The `wieldlight`
addition that the torches mod has was removed, since it relies
on wieldview to look decent. This can stay external mod code.

I've opted to move the torch nodes to a separate file. It's not
a lot of code but nodes.lua is already huge, and I wanted to
retain the copyright header and some of the readme.txt notes,
and this was the easiest way of doing it.


sofar:
Shrug.

replaces #1322. Or, it's an alternative, well tested solution at least.

sofar:
sreenshots are in the forum post:

https://forum.minetest.net/viewtopic.php?f=11&t=14359

sofar:
Fixed luacheck issues found.

est31:
Finally torches will be 3D, really like that.

Ferk:
I can reproduce the same issue with xdecor item frames that [I saw with the other PR](https://github.com/minetest/minetest_game/pull/1322#issuecomment-253020565).

I think setting an inventory_image for "default:torch_wall" would fix it. I know it's not something MTG should care about, but I think it doesn't hurt to have the inventory image for torch_wall the same as for torch, if it solves solves issues for other mods.

t4im:
Even if it would be a workaround for itemframes, it would not help the locked chest or other items, that are right clicked server side.

You generally want your custom `on_place` to behave as similar to the default `on_place` as is reasonable. Passing through right clicks does not impact the functionality of the torch, but will make it work with other nodes, just as if you would be wielding a cobble stone.

sofar:
@ferk, @t4im

Latest commit version fixes the itemframe problem: We essentially implement a better version of the `item_place` routine that doesn't put the wall torch into xdecor:itemframe, but passes the normal torch instead. This makes the torch properly show in xdecor itemframes, as well as properly drop if you punch the itemframe.

mods/default/torch.lua
```diff
@@ -0,0 +1,132 @@
...
+	drop = "default:torch",
+	selection_box = {
+		type = "wallmounted",
+		wall_top = {-1/16, -2/16, -1/16, 1/16, 0.5, 1/16},
+		wall_bottom = {-1/16, -0.5, -1/16, 1/16, 2/16, 1/16},
+	},
+	sounds = default.node_sound_wood_defaults(),
+	on_place = function(itemstack, placer, pointed_thing)
+		if pointed_thing.type ~= "node" then
+			-- no interaction possible with entities, for now.
```
t4im: The serverengine actually doesn't call `on_place` for entities.
I'm not 100% sure though if a modified client could bypass `on_secondary_use`, calling this instead. But other than that, you'd not need the check at all.

paramat:
I would prefer this torch PR due to the extensive development and testing.

t4im:
I'd prefer it too, the model seems better and it already includes a LBM.

sofar:
I just verified that if people have `torches` already installed, this change won't break any in-game torches for them. Minor detail, of course, but should help people move from `torches` away to this code if it gets merged.

Megaf:
Where's the screenshots for this?

t4im:
> Where's the screenshots for this?

*awkwardly points to the [top](https://github.com/minetest/minetest_game/pull/1325#issuecomment-253114516)*

sofar:
@megaf follow the forum link I posted above....!

t4im:
torch_sofar

sofar:
@t4im? that's not the model in this patch.

tobyplowy:
@sofar I have a question about the Torch model doesn it have These two Planes Crossing each other If Not a Lot of texture packs use that, its important It's there

t4im:
@sofar that's torch_floor.obj taken directly out of this PR; of course without textures applied yet; just some quick colored materials

Thomas--S:
Wouldn't it be better to add a new `torchlike_fancy` drawtype similiar to [this](https://github.com/minetest/minetest/blob/1079aeaa13a94c67c1e0a0bd1c461e14841a1eec/src/content_mapblock.cpp#L1028)?

octacian:
Nice. I've closed my PR, as yours seems ready to merge, while I would still have to complete a few things in my own. This version is much better anyways, as it's been in use for 6 months, while I spent 3 hours on mine.

After testing, I think the model makes a bit more sense. Everything but the model is pretty much the same as mine at this point, but things seem a bit more solid here. I do think the model should be redone again, but that would be a completely different PR.

mods/default/torch.lua
sofar: good point

mods/default/torch.lua
```diff
@@ -0,0 +1,132 @@
...
+			return itemstack
+		end
+
+		local under = pointed_thing.under
+		local node = minetest.get_node(under)
+		local def = minetest.registered_nodes[node.name]
+		if def and def.on_rightclick then
```
sofar: Don't we then want to pass on the `nil` through if that's the case? It's up to the itemframe to account for a `nil` placer (e.g. a "bot" places an item in an item frame).

`and ((not placer) or (placer and not placer:get_player_control().sneak))`

mods/default/torch.lua
```diff
@@ -0,0 +1,132 @@
...
+			return def.on_rightclick(under, node, placer, itemstack,
+				pointed_thing) or itemstack, false
```
sofar: OK, I'll take it out then.

sofar:
@thomas--S that is indeed an option, and I think I can do that as well.

huh, come to think of it, that would massively simplify this, and make client placement prediction flawless, plus remove any code changes to `mtg` at all.

*ponders*

sofar:
I've pushed this with 2 changes:

- remove the stub entity code (it's not needed)

- Addresses @t4im 's comment. If a mod right clicks, with player is `nil`, we pass `nil` through. If a player right clicks, make sure they don't have the sneak button pressed (force-place).

mods/default/torch.lua
```diff
@@ -0,0 +1,128 @@
...
+		local node = minetest.get_node(under)
+		local def = minetest.registered_nodes[node.name]
+		if def and def.on_rightclick and
+			((not placer) or (placer and not placer:get_player_control().sneak)) then
+			return def.on_rightclick(under, node, placer, itemstack,
+				pointed_thing) or itemstack, false
+		end
+
+		local above = pointed_thing.above
+		local wdir = minetest.dir_to_wallmounted({x = under.x - above.x, y = under.y - above.y, z = under.z - above.z})
```
sfan5:
```lua
local wdir = minetest.dir_to_wallmounted(vector.subtract(under, above))
```

mods/default/torch.lua
```diff
@@ -0,0 +1,128 @@
...
+		local fakestack = itemstack
+		local retval
+		if wdir <= 1 then
+			retval = fakestack:set_name("default:torch")
+		else
+			retval = fakestack:set_name("default:torch_wall")
+		end
```
sfan5: In practice `set_name` never fails (unless you pass an empty string)

mods/default/torch.lua
sofar: sure!

sofar:
Fixed up @sfan5's requested changes.

Ferk:
Just brainstorming:

> Wouldn't it be better to add a new `torchlike_fancy`

Perhaps it would be more reusable if the `mesh` drawtype ([here](https://github.com/minetest/minetest/blob/1079aeaa13a94c67c1e0a0bd1c461e14841a1eec/src/content_mapblock.cpp#L1755-L1793)) could receive a set of different meshes/textures to use based on facedir value (similar to how tiles can be defined in an array depending if you want the same repeated or different ones for top/down/sides).

I wonder if something like that could also be used for things like the open chest PR to simplify it and just use the same node that switches mesh. Though that would probably be hacky.
Too bad there's only param2 available, it could have been another use for the `meshoptions` paramtype.

sofar:
Removed `, false` being passed back as 2nd param

JurajVajda:
FYI: with old 2D torches you could do shift+click to place torch on nodes what would normaly trigger action on just "click". In other words you cannot place this 3D torch on i.e. Protector block - normally Protector opens formspecs but with shift+click it was possible to put a torch on it.

t4im:
> do shift+click to place torch on nodes what would normaly trigger action on just "click

Are you sure you tried this PR instead of the former mod?
https://github.com/minetest/minetest_game/pull/1325#discussion_r83055428 was particularly targeted to allow that behavior with right clicking.

sofar:
@JurajVajda it seems you tested the first revision of this patch:

I just tested this patch (latest version) and was able to:

1- place a torch on the side of a furnace, and bookcase (shift-right-click)
2- access the contents of the furnace and bookcase with a torch in my right hand (right-click)

So, shift-clicking to place is fully working.

rubenwardy:
:+1: code looks okay. I quickly tested and it behaves as expected:

* Old wallmounted auto changes to new torch (via lbm) - didn't check any other placements
* Still drops on dig
* wall/floor/ceiling placements all work

JurajVajda:
@sofar thats good to hear ;) i will make sure i have the latest and greatest then..thanks

sofar:
@JurajVajda the nice thing to do is to re-test it and confirm that it now works properly.

twoelk:
I really like the look of this version. It was sometimes strange when you could not see the torch from certain angles. This is way better. 
Would be nice if we had something more realistic for the placement under the ceiling some day. Didn't we have some hanging lamp for this in some old version? 

So I presume I cannot place any torches on beds, glasses, bottles, doors or apples because these are entities from the torches viewpoint? 
So I cannot place these torches on or onto Sokomines marker sticks but why not on the boundary blocks although I can do so with other things. Also while the bookshelf works, attaching torches to the wiki block only opens the formspec. The same happens with the wardrobe mod. I can even place beds on these blocks but not the torch. The mods may be outdated, so for better testing, is there any other node in mt-game that is similar enough to the torches to compare placement issues? Do the wiki and wardrobe mod and others need updating to solve the issue? 

Just for the records I could add the old torch to beds, doors, markers, wikiblocks and wardrobeblocks but not to vessels and apples. At least not without the replacer :-P

sofar:
> So I presume I cannot place any torches on beds, glasses, bottles, doors or apples because these are entities from the torches viewpoint? 

You can place these on doors, beds etc. just fine. No, doors are not entities from any perspective.

Remember to shift-right click to place on these nodes.

sofar:
@twoelk 

> Just for the records I could add the old torch to beds, doors, markers, wikiblocks and wardrobeblocks but not to vessels and apples. At least not without the replacer :-P

I think, just like @JurajVajda, that you were testing the first commit of this PR. Please update to the latest version of the code in this PR and test again (with shift-right click to place on special nodes).

I'm not sure what is up with apples and vessels, but the `on_place()`of the torch never gets called, so the problem isn't with the torch code, but it appears apples and vessels.

t4im:
The engine might just not allow placing attached nodes to e.g. plantlikes.

JurajVajda:
@sofar would like to re-test it but didnt found the way how to clone this PR.. i think i have to cherry pick ? ..or just checkout to the merge commit number...will try something like that i guess.

sofar:
```
curl -L -O https://github.com/minetest/minetest_game/pull/1325.patch
git am 1325.patch
```

sofar:
@sfan5 @paramat any further comments?

paramat:
Will test.

paramat:
screenshot_20161024_041733

I feel fairly neutral on this. One thing that bothers me is how the upside-down torches have flames that burn downwards. I'd prefer the stick is angled, perhaps using the angled stick used for wall torches, then the flame can be upright, notice the flame itself burns at an angle so if angled away from the stick it wont overlap much with the stick.

Fixer-007:
Or simply download torch branch of sofar's repo. I prefer this - much faster and easier.

Fixer-007:
Placing to the side of the block is unsmooth - first it blinks horizontally and then to correct position.

Thomas--S:
May I ask, why we still put effort in this PR? Wouldn't it be more useful, if we put these efforts in implementing the [`torchlike_fancy` drawtype](https://github.com/minetest/minetest_game/pull/1325#issuecomment-253539209) – provided that everybody is ok with this. (But according to the three likes, I think there aren't too many opponents.)

est31:
@Thomas--S that was discussed in the engine before, but all attempts to add such a drawtype failed. I think the game should first add it via @sofar 's approach, and later on when we want we can still add such a drawtype to the engine, but [people were critical about it](https://github.com/minetest/minetest/pull/1328#issuecomment-44064570).

sofar:
> Place torch on or to the side of someones bone (it is important that it is not placed bone) - remove bones - torch will hang in its position (including in water).

Bone removal handler needs to call `nodeupdate` - bug in bones?

> Placing to the side of the block is noticably unsmooth - first it blinks horizontally and then to correct position - looks annoying to me.

Yes, sadly placement prediction doesn't help. I could disable placement prediction, but that would mean placement of upright torches will be subject to lag.

> Old wield model in hand, old inventory pic

on purpose, it's recognizable.

sofar:
I'm with @est31 that I'd like this to merge first before we make a special drawtype... Having this PR get far enough that people are positive about the visuals will help do the drawtype later.

sfan5:
:+1: Code looks good, I also agree that we need this generally.

paramat:
:-1: The upside-down torch needs reworkiing.

sofar:
It's doable to redo the upside down torch, but it needs a new mesh, and therefore a new node, which will complicate the code (although marginally, it needs handling in placement and in the LBM).

I can certainly do so. Just need a few days to get to it :)

Thomas--S:
@sofar @est31 Sounds reasonable. :)

tobyplowy:
Oh man I'm so excited thank you @sofar for this

tobyplowy:
@sofar when will this be ready for Merge? (Just A Question I'm really excited About This)

sofar:
I've taken a step back to think this PR over a bit more. I'm not convinced that this is the right way forward given the added complexity to handle a 3rd model for upside-down torches (a lot of work for a very uncommon case, I'd say), and coding it up as a client-side builtin seems far more efficient (plus, it'd be essentially free).

So, I'm doing some reflection on this idea.

paramat:
Perhaps just 2 crossed planes are enough for the torch stick?

tobyplowy:
@paramat please no, if you do that torches will look crap again then This whol ting has no Point (why don't we get rid of The Upside Down torches almost no one uses them and they don't really look that great in builds) I didn't even notice them before this topic

sofar:
@paramat It's not the model that makes it complex, it's the fact that we need a new node.

I'm committed to keeping the upside down torch btw. It's already in many worlds and shouldn't just disappear.

tobyplowy:
@sofar I know this may sound like a stupid question but could be upside down torch be turnd into a lantern? It would Look More realistic Than An Upside Down torch but the fact that the torch magicly turns into a Lantern is not really realistic :\

paramat:
Well feel free to remove 0.4.15 milestone, that's up to you, 1 month to freeze.

sofar:
@tobyplowy anything is possible, but changing it into a lantern is as bad as removing it IMHO.

Ferk:
If you are planning to make this in the engine, maybe it's best to do that in one step without having a release that adds 2/3 nodes for 3d torch orientations.

Otherwise there would have to be additional aliases and compatibility logic in the version that implements the new method, one for the worlds that contain the 0.4.15 3D torches and another for the 2D torches pre-0.4.15

tobyplowy:
@sofar I know but it would look a lot Better for building thow :)

sofar:
OK, reading the discussion above, and after having explored making a drawtype, I think we should *merge this PR as is*.

Several others have pointed out that making a drawtype at a later stage would not pose any compatibility issues (1 simple ABM would suffice to adjust existing maps with either old or these torches), and that merging a drawtype later in the engine at a later stage would be just fine.

So, given the two approvals and the general support, we should pursue merging this PR.

As for the upside-down torch, I'm of the opinion that the current model isn't ideal, but the simplification is acceptable enough for many (it's a huge improvement compared to the old torches anyway).

I'm certainly willing to try and add a new mesh for the upside down torch, but given that this code has been tested so well already, I'd like to do that at a *later time* and get this code in now, and then add a new mesh later once we iron out the placement kinks.

Sound ok?

paramat:
> Several others have pointed out that making a drawtype at a later stage would not pose any compatibility issues (1 simple ABM would suffice to adjust existing maps with either old or these torches),

Such an ABM should be avoided if at all possible, ABMs are intensive and have to remain for a very long time to properly convert old worlds. There's no justification for it as it's just a case of properly finishing the PR.

> I'm certainly willing to try and add a new mesh for the upside down torch, but given that this code has been tested so well already, I'd like to do that at a later time

I oppose this. the new upside-down torch is much worse than the old 2D one.

Even if this means missing release the upside down torch model should be completed, so i'm removing the milestone.

Fixer-007:
What you gonna do with that unsmooth torch placing to walls? It looks kinda crappy for a moment when you place it, can you just hide first step and show it when it corrects its position?

est31:
@sofar @paramat Isn't that a task for LBMs anyway?

paramat:
Indeed i wondered why sofar wrote 'LBM', i'd still rather avoid those though.

sofar:
I meant LBM, not ABM.

sofar:
@Fixer-007 disabling placement prediction seems worse to me, since then it will be painfully apparent that you have lag. I don't think anyone else has complained about that yet.

sofar:
Updated and now includes a separate model for the ceiling torch.

I've played a bit with texture offsets and this comes out fairly decent - about 1 px of the flame comes out at the bottom, the rest at the top of the torch. Obviously, parts of the flame are hidden by the stick part, that's just unavoidable.

I've opted for a fairly steep angle so that it looks like it hangs, and not an angle that makes it look like it's horizontal, which looks less appealing.

paramat:
Will test.

paramat:
:I approve of the ceiling torch +1.
Haven't reviewed the code though.

sofar:
renamed the LBM to a nice and short `default:3dtorch`.

retested the conversion LBM and it's all OK.

Fixer-007:
>disabling placement prediction seems worse to me, since then it will be painfully apparent that you have lag. I don't think anyone else has complained about that yet.

I've seen it on some videos too, it is not me only, I think I will be at least the one with OCD about it.

sofar:
The problem with disabling placement prediction and lag on torches is that it leaves you dark in caves until the server finally delivers you the map update packet.

Now, it may look weird that the torch "changes" on the wall or ceiling, but it's much more useful to players that they can see in the dark quickly than that it looks seamless, because after all these are torches.

Until we get the core engine to have a drawtype for this, we can't solve it. Perhaps client side mods can do a better job at placement prediction, but I wouldn't bet much on it.

Fixer-007:
What about this workaround: just don't draw the torch at first stage, so it lights the cave but torch will appear in its final state only.

sofar:
That would mean adding a 4th node, and more handling code for all this.

Just... no.

paramat:
I agree with sofar, this is fine as-is.

rubenwardy:
My approval remains

------

I'd prefer if the wall torches were aligned with the voxel grid, and attached to the wall using a fitting. Like this but more of a torch than a candle:

torch

But I don't mind that much, and if other developers don't either then I'm fine with it

tobyplowy:
👍

Marnack:
The merged torches implementation seems to only work when converting from the latest versions of the Torches mod.  Here's a potential improvement for the implementation of minetest.register_lbm() in torch.lua that supports converting torches from an earlier version of the Torches mod.

After opening a large world in 0.4.16 that I had created a few years ago in MT 0.4.12 that had used the Torches 2.0 mod, I noticed that all of the torches placed on a horizontal surface (e.g. the ground, the top of a fence post) were now upside down.  After some investigation, I traced this back to the MT 0.4.15 release and this particular code merge.  As a test, I modified minetest.register_lbm() in games\minetest_game\mods\default\torches.lua to be the following:

```lua
minetest.register_lbm({
	name = "default:3dtorch",
	nodenames = {"default:torch", "torches:floor", "torches:wall", "torches:wand"},
	action = function(pos, node)
		if node.name == "torches:floor" then
			minetest.set_node(pos, {name = "default:torch",
				param2 = 1})
		elseif node.name == "torches:wand" then
			if node.param2 == 0 then
				minetest.set_node(pos, {name = "default:torch_wall",
					param2 = 4})
			elseif node.param2 == 1 then
				minetest.set_node(pos, {name = "default:torch_wall",
					param2 = 2})
			elseif node.param2 == 2 then
				minetest.set_node(pos, {name = "default:torch_wall",
					param2 = 5})
			elseif node.param2 == 3 then
				minetest.set_node(pos, {name = "default:torch_wall",
					param2 = 3})
			end
		elseif node.param2 == 0 then
			minetest.set_node(pos, {name = "default:torch_ceiling",
				param2 = node.param2})
		elseif node.param2 == 1 then
			minetest.set_node(pos, {name = "default:torch",
				param2 = node.param2})
		else
			minetest.set_node(pos, {name = "default:torch_wall",
				param2 = node.param2})
		end
	end
})
```

In addition to correcting the conversion of torches from the Torches 2.0 mod placed on horizontal surfaces to standard torches, the code also properly converts torches from 2.0 that were placed on vertical surfaces (i.e. sides of nodes) to standard wall torches.

For reference, here is the original version of minetest.register_lbm() in MT 0.4.15:

```lua
minetest.register_lbm({
	name = "default:3dtorch",
	nodenames = {"default:torch", "torches:floor", "torches:wall"},
	action = function(pos, node)
		if node.param2 == 0 then
			minetest.set_node(pos, {name = "default:torch_ceiling",
				param2 = node.param2})
		elseif node.param2 == 1 then
			minetest.set_node(pos, {name = "default:torch",
				param2 = node.param2})
		else
			minetest.set_node(pos, {name = "default:torch_wall",
				param2 = node.param2})
		end
	end
})
```

I've not done any coding in Lua before so I'm sure my changes could be handled better but, as a proof-of-concept, it shows how the torches:wand and torches:floor nodes of the older Torches 2.0 mod might be converted to the new standard torches of minetest_game.  I have not tested this with a world using the newer Torches 3.0 mod (other any other torches mods) so this could potentially be an insufficient solution.  However, this change did allow me to successfully convert the torches in my old world to the new standard torches.

Of course, there may not be many people still running with the older version of the Torches 2.0 mod to make this a worthwhile update to minetest_game.  In that case, they could update their local copy of torches.lua as I did.

paramat:
^ @sofar

sofar:
It's easy enough to consider merging a PR - so, please submit one with the change?

PR #1331, opened 2016-10-15
Use unregister_item() to remove tnt:tnt if tnt is disabled.

red-001:
This allows for disabling the default TNT and using other mods that register TNT using tnt.register_tnt()


Ferk:
Why not just prevent "tnt:tnt" from registering in the first place?
You can still leave the register_tnt definition untouched for other tnt mods to use, instead just make  [the `tnt.register_tnt` call that uses it](https://github.com/red-001/minetest_game/blob/02df34771a819e4e22ba5cdfd609722a3f203b8e/mods/tnt/init.lua#L591-L595) conditional.

tenplus1:
Current dev additions have 'disable_tnt = false' not register tnt:tnt and allow the use of tnt functions and tnt:tnt_burning for mods...  No need to unregister a node.

PR #1335, opened 2016-10-20
Sapling LBM timer check

tenplus1:
This checks saplings for already running timers before starting a new one.

### DIFF
--- a/mods/default/trees.lua
+++ b/mods/default/trees.lua
@@ -87,7 +87,7 @@ minetest.register_lbm({
 			"default:pine_sapling", "default:acacia_sapling",
 			"default:aspen_sapling"},
 	action = function(pos)
-		if minetest.get_node_timer(pos):is_started() == false then
+		if not minetest.get_node_timer(pos):is_started() then
 		        minetest.get_node_timer(pos):start(math.random(1200, 2400))
 		end
 	end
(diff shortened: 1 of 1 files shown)


sofar:
This isn't needed. 

Conversion only happens once, ever, and the penalty of restarting a timer (which, being a conversion, is highly unlikely if not impossible) is so incredibly tiny that the extra check is just a waste of cpu time.

This code runs when (1) you have saplings in a block that (2) were added before the new timer-based sapllings. In that case, we *must* start the timer since those saplings would otherwise never grow again. The extra check just confirms that timers were not running on old saplings.

Saplings planted with later mtg versions won't ever get converted, because this ABM will have already ran on the block that they are planted on.

sofar:
After talking with @tenplus1 on IRC, it seems that he is having issues with timers not being active after unloading and reloading an area, or restarting the server.

This is either a weird custom code issue or an engine bug. In both those cases the above patch is not needed (or wanted), so we should likely close this issue.

I was unable to reproduce the issue he has with this code:

https://gist.github.com/sofar/f6daa81cdba74ed2754be2877f8aa979

Fixer-007:
i was able to reproduce with tenplus1 guidence, there is might be a bug in engine related to that.

t4im:
> https://gist.github.com/sofar/f6daa81cdba74ed2754be2877f8aa979

Wouldn't it make more sense trying to get the node timer of an definitely unloaded node to test this?
If you have to punch a node you're usually loading it anyway most of the time, ignoring the border cases you could enforce instead.

PR #1357, opened 2016-10-30
Use opaque texture for top+bottom of iron bar

Wuzzy2:
This PR changes the textures of the iron bar. I always thought the iron bar looks very strage/ugly/unreal on the top and bottom.

Before:
Iron bar, before the change

After:
Iron bar, after the change


sofar:
It's good but it's all exactly the same color. Can you maybe weave in a bit of the grey color variation to make it a bit less monotonous?

Other than that this is fine!

Wuzzy2:
I don't really know how I could do the texture correctly, sorry. :-(

sofar:
ah, I just saw, you're not even modifying the texture anyway.

Let's do that in a different PR then. I'll look into it.

sofar:
:+1:

paramat:
:-1: The combination of 3D top bar and flat texture is more ugly. It's a flat texture so obviously unrealistic, it's alos visually consistent with ladders, plantlike etc. We have to stop moves like this that are trying to suggest flat textures are unnacceptable. This adds render load too.

paramat:
We could improve the existing texture by adding subtle shadowing below the top bar, to give it more 'substance', i'll even do this myself. If the top and bottom bars bother you why doesn't the rest?
I think 3D bars should be a mod.

paramat:
Oops i misunderstood, removed my objections.

The top needs a little colour variation so it has the same appearence as the side of the top bar.
As long as this is addressed soon :+1: 
I can't even see where the 'grey' texture is used in the code.

paramat:
screenshot_20161031_052558

xpanes_bar_top

xpanes_bar_top128

This works.

paramat:
Hang on, i'll redo that texture, i don't like how the dark pixels line up.

paramat:
screenshot_20161031_061957

xpanes_bar_top

^ xpanes_bar_top.png

xpanes_bar_top128

Feel free to use it for this commit.

Wuzzy2:
@paramat: Texture is used. `xpanes_grey.png` was deleted since it is not used anymore.

mods/xpanes/README.txt
```diff
@@ -16,3 +16,6 @@ xyz (CC BY-SA 3.0):
 
 Gambit (CC BY-SA 3.0):
   xpanes_bar.png
+
+paramat (MIT):
```
paramat: License for textures is always CC BY-SA 3.0, please could you change it?
Also, for perfection, add my name in license.txt after line 35.

paramat:
Thanks, added a line comment, sorry. :+1: with the line comment attended to.

paramat:
:+1:

sofar:
Excellent! :+1:

PR #1360, opened 2016-11-01
sethome.get: Copy a return value

Rui-Minetest:
Currently, the following code works.
```lua
sethome.set("singleplayer", {x=1, y=2, z=3})

local pos = sethome.get("singleplayer")
print(pos.x, pos.y, pos.z) -- 1 2 3

pos.x = 4

local pos2 = sethome.get("singleplayer")
print(pos2.x, pos2.y, pos2.z) -- 4 2 3
```
This PR fixes it.

### DIFF
--- a/mods/sethome/init.lua
+++ b/mods/sethome/init.lua
@@ -41,7 +41,12 @@ sethome.set = function(name, pos)
 end
 
 sethome.get = function(name)
-	return homepos[name]
+	local pos = homepos[name]
+	if pos then
+	 	return vector.new(pos)
+	else
+	 	return nil
+	end
 end
 
 sethome.go = function(name)


mods/sethome/init.lua
```diff
@@ -41,7 +41,7 @@ sethome.set = function(name, pos)
 end
 
 sethome.get = function(name)
-	return homepos[name]
+	return table.copy(homepos[name])
```
Rui-Minetest: vector.new

mods/sethome/init.lua
```diff
@@ -41,7 +41,7 @@ sethome.set = function(name, pos)
 end
 
 sethome.get = function(name)
-	return homepos[name]
+	return vector.new(homepos[name])
```
Rui-Minetest: should check nil

sofar:
This just boggles my mind, sorry, returning a table `and` a table?

Can we just do:

```
if pos then
        return vector.new(pos)
else
        return nil
end
```

instead? That way there's a way to detect that a player has no home, and mods may want to unset homes as well (e.g. the bed is destroyed, or some other mechanism).

stujones11:
@sofar you don't even need the else clause, the function will return `nil` anyway ;-)

Rui-Minetest:
I think my code and your code are equivalent.

http://lua-users.org/wiki/ExpressionsTutorial
See an `and` section.

sofar:
@Rui-Minetest  That is just an illustration of how confusing that `and` statement is....

stujones11:
Since this is valid Lua I don't think the PR can be rejected because it looks confusing. I must admit I was not aware of this feature either but I do not think it is one I will be adopting.

sofar:
@stujones11 I wouldn't reject it just for that either. But I'll complain again :)

Rui-Minetest:
Fixed the confusing style.
Goodbye the code which is difficult to read.

sofar:
:+1:

Ekdohibs:
:+1:

Ferk:
Personally, I actually find simple uses of `and` more readable than adding yet another `if`-`then`-`else`-`end` level of indentation. Just so long as it's not a complex expression.

`and` never really meant additive conjugation, it's just a logical gate. `and` and `or` have been used like this for ages in several programming languages.

stujones11:
I am in favour of code readability but the return nil part is really not necessary, whatever...

sofar:
computationally, branches in `or` and `and` statements are identical to `if` branches.

PR #1371, opened 2016-11-06
Make hand in creative mode golden

Wuzzy2:
This is a small graphical tweak.
If creative inventory is enabled, the (wield)hand will become golden. It looks like this:

Golden hand

I did this just for fun, as a nod to the super powers your get in Creative Mode. :D

Known problem: The player model texture is not modified. So the wieldhand does not match in color with the character texture in Creative Mode. Changing this texture as well would probably too complex and/or hard to implement. This would create a big chaos with custom player textures, so I decided not to bother with it.

Tell me if you like the idea or not. This was just a quick silly idea and I am not sure myselves if it really shold be added.


red-001:
+1

________________________________

sfan5:
...why?

Iniquitatis:
@sfan5, +1.

sofar:
I don't mind the idea per se (I think it's good to have a visual confirmation that you're in creative), but I had hoped that we could actually use the player skin instead, and this would conflict with that.

RHRhino:
I like the idea of a visual feedback for creative, but I'm unsure if a "golden hand" really helps players to understand that they are in creative. The chance that they just get confused and think that this is a graphic bug is quite high.

paramat:
Cute and fun idea, but i feel it's best to keep the normal appearence, especially as we would ideally want to auto-render the hand with the current player skin.

Wuzzy2:
@RHRhino: It's not even meant as an usability feature. I did it for teh lulz. xD

DS-Minetest:
wouldnt it be better if it was a mese hand, gold is heavy
or make the hand grey so that the player thinks he would be a nyancat :cat2::rainbow_flag:

C1ffisme:
At first I thought this was about making the default skin yellow, so that it wouldn't imply that the player is either white or black, like how github changed it's thumb textures not to be "racist".

:+1: By the way.

tobyplowy:
@Wuzzy2 this is really funny and all but I don't think it's a good thing -1 from me :/

paramat:
Cute but disapproved.

PR #1392, opened 2016-11-17
Improve arms and orientation of character model

Rogier-5:
The body was rotated 180° while the rest wasn't, leading to
strange results when setting bone positions manually.

The arm bones were at an angle, so that the bones were difficult
to orient correctly, as unintuitive combinations of angles
would be needed to get the default orientation, or any other
acceptable orientation. Moreover, a simple rotation of the arm,
e.g. forward, would require all angles to be adapted, instead
of just one.

See issue #1376


Rogier-5:
Sample images:

Left = before, right = after:

screenshot_20161113_000527 screenshot_20161113_001521

Pointing sideways is *much* better:
screenshot_20161113_000500screenshot_20161113_000722

Pointing forward:
screenshot_20161113_001242screenshot_20161113_000921

paramat:
screenshot_20161118_010926_1

New appearence when digging: shoulder is lower due to new arm rotation point. Arm is still slightly angled when digging.
When walking it seems the arm rotates around the new 'shoulder-cube centre' rotation point too.
:+1:

stujones11:
The latest model looks good and seems to be free of warnings, however, I have noticed and error in the 'body' vertex group. If you select only the body from the vertex groups menu then you also get parts of the arms and legs selected. This error is very likely present in the existing model and not your fault but it really should be corrected.

On the whole I think this is a definite improvement but it is going to affect some mods, albeit only visually, like my own wield3d for example. I am happy to update as this will make attachment positions a lot easier to calculate and much more intuitive to modify.

paramat:
The error found by stujones11 should be addressed, then i will +1.

Rogier-5:
>>  If you select only the body from the vertex groups menu then you also get parts of the arms and legs selected.
 This error is very likely present in the existing model and not your fault but it really should be corrected.

> The error found by stujones11 should be addressed, then i will +1.

I propose that this be done later, in a different PR. For the following reasons:
- The problem was pre-existing, and not introduced by my changes (as far as known).
- It apparently (?) only affects the contents of the blender file, not the final model (.b3d file). So the  desirability of my changes so far is higher than the importance of this additional fix.
- As it doesn't affect the exported model, it is more a maintenance-like change.

Note that if it were a simple change that I could easily make, I'd fix it right away - or rather: it would already have been done by now. but ATM I seems like I can't (maybe because of the older version of blender that I currently use), so if the reasons above are not convincing, then maybe the additional circumstance that I most likely won't be able to fix this issue until I upgrade blender, which won't be in the short term, may be sufficient to accept the continued presence of this error in the model for the moment...

Unless of course somebody else fixes it, but then it'd be a different PR as well.

stujones11:
@paramat it turns out that error is in the existing model but it should be fixed at some point at least. I am doubtful that it will fix any of the other model related issues.

@Rogier-5 I am happy to make those changes as I have also just found a 'fix' for this long-standing issue https://github.com/minetest/minetest/issues/2813 I will explain (if I can) later.

I will attach a zip file asap

stujones11:
Here is the model with fixes applied for both of the above issues, you can incorporate the changes or I can make a subsequent PR, though I would really rather not have to.

[character.zip](https://github.com/minetest/minetest_game/files/600940/character.zip)

paramat:
> I propose that this be done later, in a different PR. For the following reasons:

Ok that's fine :+1:

Rogier-5:
@stujones11 I tested your new model, and all looks OK as far as I am concerned. Please do describe both changes you made, so that I can add that info to the commit text. As the commit cannot speak for itself (file can't be diff-ed), IMO the commit message should speak for it...

... binary / non-diffable formats are terrible...

stujones11:
@Rogier-5 I have corrected the body vertex group which was trivial but I noticed that the hand switching of arm attachments was only present in the walk animation. After a lot of hair pulling and experimentation, it turns out that it flips out when the body rotation is zero on all axis and the walk animation is the only sequence when the body bone does not move. My modification was to make the body bone move only +/-0.0001 blender rotation units during the walk animation and you can prove this works with my wield3d mod to see the 'hand-switching' fixed

Rogier-5:
So IIUC the root cause is probably that irrlicht mixes up things? And that your change somehow prevents it from getting confused ?

Then of course, the question is whether irrlicht gets confused because of an irrlicht bug, or because of a model bug. But that doesn't need to be answered now.

stujones11:
> So IIUC the root cause is probably that irrlicht mixes up things?

If not irrlicht then I'd say it must be the exporter(s), I can't find any more errors in the player model and you get similar behaviour with simpler models. While my changes might not fix the actual root cause, it does fix the problem from the end user's perspective. I will continue to look into this as time permits but I don't think any more can be done by merely editing the model.

paramat:
Let us know when this is ready for merge.

stujones11:
For me it is good for merge any time, I can submit my changes in another PR. I just thought it might save a bit of commit spam if these very minor but worthwhile changes could be included here.

paramat:
Combined or a separate PR is fine for me, i'll leave it up to Rogier-5 to decide.

paramat:
@Rogier-5 ?

Rogier-5:
@paramat: Work and other activities have occupied me the last few days... I may be able to push a new version today, or else tomorrow. It'll include @stujones11's changes.

Rogier-5:
@paramat, @stujones11: I pushed a new version. The final versions of the files are the ones from [character.zip](https://github.com/minetest/minetest_game/files/600940/character.zip) (from [this comment above](https://github.com/minetest/minetest_game/pull/1392#issuecomment-261643222)).

rubenwardy:
This looks like a good change, although I can't say I know much about meshes.
It seems like this could break mods due to different bone placement - but this isn't a problem as it seems to be broken already (From the before pictures)

stujones11:
@rubenwardy I am not sure of any mod other than my own that is likely to be affected by this. Rui's player animation, maybe? I my case the only difference is that the hand wielded item is tilted slightly but this will be a joy to fix now that the numbers make sense. Also the end of that annoying hand switching, afaik, your CTF players should appreciate that ;-)

paramat:
I don't understand models in depth so @sofar could you have a quick look at this to see if it's reasonable?

paramat:
Testing again it seems good, but one thing bothers me, the arm seems to get thinner when digging or when pointing out to the side:

https://cloud.githubusercontent.com/assets/6202053/20242699/85274a8a-a938-11e6-8cae-650f3e61406a.png

screenshot_20161124_042439_4

Maybe it's an illusion.

sofar:
This seems OK. 

- Tested with `player_animation` and works without any issues.
- Tested with `cozy` and sitting animation now has wrong body offset, making player float above the ground. Since everemzah is AWOL, we'll have to fix it up anyway, so I don't consider it a major issue.

:+1:

Rogier-5:
I'll look into the sitting animation. I shouldn't have been affected though.

I think that the arm thickness was pre-existing, as I took care not to change anything about the meshes. In the old image, to me it seems that even though the top of the arm is level with the center of the head, the bottom of the arm does not seem level with the bottom of the head...

Anyway, I'll have a look at the arm as well.

I don't suppose anybody reading this is aware of a diff tool for blender models ?

Rogier-5:
@sofar: I tested with `cozy`, and I don't see any difference in the sitting animation: in both cases, the character sits on the ground, with no discernible gap. I used an updated and freshly compiled minetest tree, with an updated minetest_game tree.

New model:
screenshot_20161124_194047

As for the arm width: the scale of the character's pixels is not identical in all directions and for all body parts:
- The head is 4 blender units high, and 8 pixels, so 1 pixel is 0.5 blender unit
- The body is 4 blender units wide x 2 deep. It is 8 x 4 pixels so unsurprisingly the pixel size is also 0.5
- The arms and legs are 2 x 2 blender units, and 4 x 4 pixels, so unsurprisingly the pixel size is also 0.5

However
- The height of the arms, legs and body is 6.75 blender units each, and 12 pixels, so the pixel height is 0.5625 instead of 0.5.

That means that those pixels are actually rectangular instead of square, and that the arms, when not in default position, are never going to line up with the body's pixels. Unless of course the character were to be resized in some of the directions.

paramat:
Excellent thanks, that explains it, we could resize the model in another PR then, the arma, leges and body really should be 6 units long to make the pixel sizes consistent.
The legs pivot at the midpont of leg top and body base, similar to how the arms used to, not really a problem.
This can be merged.

Rogier-5:
> Excellent thanks, that explains it, we could resize the model in another PR then, the arms, leges and body really should be 6 units long to make the pixel sizes consistent.
> This would make the model shorter so collisionbox and damage detection points may then need changing.

Currently, all dimensions seem to have been carefully chosen (1, 2, 4, 12.5, 6.75, etc. No values like 1.0017, 6.74983, etc.). In addition, the body, legs and arms have exactly the same height. I can hardly imagine all of this happened by accident or by oversight. Presumably, it looks better this way: if the body and legs were shortened, the head would become relatively larger (i.e. the character will be more child-like), and the model might even start to look fat.

> The legs pivot at the midpont of leg top and body base, similar to how the arms used to, not really a problem.

I noticed that as well. I think moving the rotation point of the legs down 2 pixels makes the different manual leg orientations look worse than they do now. Alternatively, the legs could be extended by 2 pixels into the body, but the extension would have to be hemispherical to avoid the legs protruding from the body when at an angle. In that case, the consequence is of course that the model loses a bit of its block-like nature...

paramat:
Yeah the leg rotation is fine.

The body and limbs were deliberstely 12.5 of course, but i think that was a mistake, 12 is not much difference and is no different to Mnecraft so will look fine.

paramat:
fd87b04f3c510fd61797c6e79b76830f03f485b7

PR #1393, opened 2016-11-17
Add sample texture with cape

Rogier-5:
I propose to add an example texture with cape.

screenshot_20161117_104149

NOTE: this will _not_ be the default texture. Just an additional png file that is included as an example of how to add a cape to the model.


DS-Minetest:
i dont think, this has to be reported in minetest game

sofar:
I'm ok with adding this template, and adding it here.

But please use `_` and not `-` in file names.

C1ffisme:
Paramat talked about removing the cape altogether, but I don't remember what happened to that conversation.

Rogier-5:
@sofar: I should have realized that... Fixed.

DS-Minetest:
with this some players will just take the texture and replace it
maybe it would be good to
- make the cape to the texture by default
- or make a setting to enable capes

Rogier-5:
@DS-Minetest I think that many people will prefer a character without cape. So I think it's better to not have a cape by default. However I personally wouldn't mind if the default character had a cape.

Having a cape or not is just a matter of choosing the right texture. See [this mod](https://forum.minetest.net/viewtopic.php?id=4147). I don't know if it would be possible to actually disable just the cape. Or an arm or leg for that matter. Anyway, if it's doable at all, then that would be a minetest-core change.

paramat:
> make the cape to the texture by default
or make a setting to enable capes

I disapprove of both,

Since this is an example of how to add a cape to the player skin, if something like this is added perhaps better to include a large template with all texture sections labelled, like:

skinzones

However i'm still unsure whether something like this belongs here.

My general objection to capes is because they are a copy of something very Minecrafty, capes in Minecraft are very much associated with Minecon exclusives, so capes seem a little like distasteful copying, like Minecraft adding Mese block.

DS-Minetest:
i agree @paramat's opinion to make a template like this
but capes arent really typical for minecraft. thats not comparable with mese

rubenwardy:
Wouldn't this be better on the wiki?

paramat:
Indeed i doubt this is the right place for templates.

paramat:
ShadowNinja's opinion http://irc.minetest.net/minetest-dev/2016-11-28#i_4751640
Closing due to 3 -1s.

PR #1422, opened 2016-11-27
Fix ItemStack method call for keys.

sofar:
Fixes (I hope) #1420 (bad method call).


Fixer-007:
Fixes the problem. Side effect: if I place another empty locked chest and try dig it with that key - locked chest will reappear after digging.

PR #1428, opened 2016-11-29
Default: allow character to be more feminine (depending on texture)

Rogier-5:
Add long hair and breasts.

The longer hair is divided into sections horizontally and vertically, so that different hair styles and lengths are possible.

Both features can be enabled by changing the texture only.

female-1 female-2 female-3 female-4


DS-Minetest:
this doesnt look very feminine i think...

sfan5:
( ͡° ͜ʖ ͡°)

SmallJoker:
What do the players say? There are already many skins used in-game that look feminime. The Sam model, as it currently is, seems well fitting for everybody.
[Old action game pendant model](http://www.gamerheadlines.com/wp-content/uploads/2014/02/Tomb-Raider-original-game.jpg)

Rogier-5:
> this doesnt look very feminine i think

@DS-Minetest In what way do you mean ?
I hope the model is at least more feminine than the existing model, even with the adapted male skin I used ?

I tried to make it more feminine, but still compatible with the current model (no second model needed). If you have an idea about how I could do better, I welcome your suggestions !

Maybe a completely different female model is preferable. The advantage of this is that a skin change is sufficient. A disadvantage is that a skin change is needed :-)

> What do the players say?

I hope to hear from them.

@SmallJoker I *am* aware of the existing female skins (of course). This is not about a female skin, but primarily about a more female form and long hair that is attached to the head instead of sticking to the body. I just happened to use a modified Sam skin. I'm sure few people will like this particular skin.

Rogier-5:
A few more images showing how the hair is attached to the head (with full length hair):
female-5 female-6 female-7 female-8 female-9

paramat:
Sorry i think this looks realy bad, and not half as cute as a current female skin. :-1: 
Breasts on player models just don't look right, especially sloped ones as here.
Bothering with this sort of detail is like adding a groin bulge for male players, which would be horrific.
This also relies on the cape feature which is likely to be removed or replaced with full-body overlay.
Hair should be defined in the texture, not as an add-on cuboid, it has infinitely more artistic detail and variation that way.

One idea i do like is having thinner arms, 3x3.

rubenwardy:
I think the boobs are too much detail for such a voxel model.

The hair doesn't look quite right to me, not sure why

zorman2000:
As a player, I strongly believe we should have a default male and female texture. However, I'm not entirely sure this is the correct approach. It doesn't look horrible, but the breasts do look out of place. You could make the head a little bit shorter, and the legs and arms and body in general a little bit less wide. And then of course, we need some female texture. Maybe one with the clothes color inverted (blue shirt, green skirt or pants?)

In addition, it would be great if players could choose what character (male/female) to play with when logging to the game, but that is another PR.

paramat:
Because current female skins are already extremely cute and attractive i feel we don't need more than 1 player model.

NathanSalapat:
Wouldn't this model break all existing textures, I imagine having extra mesh would mean a different UV map, which would render all existing textures useless.

Would be better to just have a way to choose your skin in game, similar to what several mods already do.

DS-Minetest:
@NathanSalapat 
it seems like it adds extra mesh and places the uv map of it on the empty spaces of the current uv map.
the existing textures leave these areas transparent and the woman body parts are invisible

sofar:
The implementation leaves a lot of issues, but I'm absolutely not opposed to any of this.

The biggest problem to be solved first isn't "should we add more characters or skins", but "how is the player going to change skins" instead.

So, unless someone comes up with solutions, we're not going anywhere. Suggestions? Sure:

- a method through which a different character model can be chosen
- a method through which a different character texture can be chosen
- a "character creator" that, at first login, allows characters to choose skin and model
- a randomly assigned skin & model so that new players don't all start out looking the same.

think about it, before sending more models/skins.

DS-Minetest:
can you let the players choose the models and skins from a database?
like [this one](http://minetest.fensta.bplaced.net/) but also with models
maybe make a setting of what skin and model should be used
and a setting for servers and singleplayer server to load from the database (of course with need of internet)
is something like this possible?

sofar:
No, an external database is not an option. It forces us to do one of these things:

- provide a database, or
- use network connections from lua, or
- use network connections from the engine

And there would be no way around it - players would have no choice to disable any of this, and it would just make stuff break easier and be insecure.

DS-Minetest:
then maybe make it like in teeworlds _(another game)_?
we have a `sounds`_(in the past `testsounds`)_ file, right?
can we do something like this with skin textures?
suggestion:
you can set the skin with a setting for the skin name _(eg. "bla.png")_
every player who doesnt have a skin with this name in his/her file sees the player with "character.png"
else if you have the texture you see the player with your texture named "bla.png"

paramat:
I'm not opposed to making it possible for a mod to provide multiple models to choose from, but am opposed to having a fixed alternative model as here.

stujones11:
IMO, alternative skins/models do not make much sense in a singleplayer game and most servers already have their own skin changing systems in place. I agree with @paramat, this sort of thing is best left to mods.

C1ffisme:
If it's done well, I'd take a female player model, but holy crap, I think I'd rather have one of those 'emo teenager' skins, than that.

I like Minecraft's approach better: Rather than trying to add breasts to a female character in an abnormally cubic world, just make their arms thinner.

About the hair, the current suggestion makes the hair way too long. How about just putting a ponytail over the character's shoulder?

stujones11:
Sam, iirc, was meant to be asexual anyway, perhaps a shortening model a little would help to emphasize that?  #1411

paramat:
Yeah Sam is either sex or neither. I intend to make the texture cuter, bigger eyes, slightly longer hair.

stujones11:
> just make their arms thinner.

I have to admit that the three pixel wide arms of the 1.9 format do look proportionally better for both male or female skins. The problem is that is would break all existing skins with no clean method of conversion. It should also be noted that the skin format 1.9 supports semi-transparency, which minetest does not.

pithydon:
Fabio skin anyone?

tobyplowy:
@pithydon Yes Me All the way baby!! XD sorry

C1ffisme:
@stujones11 I think the idea behind the thinner arms was that instead of doing breasts, Alex (the newer female character) would have a larger body (And head, I guess) relative to her limbs. But because she would be slightly smaller, her body and head would he the same size as Steve's.

If anything, I would think a larger problem would be that Sam's skin is a blinding colour of white, but that problem is out of scope for this pull request. Although I will note that Alex's skin is also a bright colour of white, probably because at the time she was made, there was no need for interracial skins because you could already just download one off of the internet or make your own.

octacian:
I agree with @paramat's comment a while back, the current model is just fine even for girly skins. And, the breasts look just weird. Like weird. Would be better without them altogether.

However, as sofar said, before doing something like this a way has to first be provided for players to switch skins and customize their skin. I wanted to do this a while ago, but realized it was beyond my knowledge as C++ would surely be required for the skin editor and probably even to just open a file to import a skin. Maybe not for the latter, but IRDK.

On the actual player model side, I think it's far more important to work on something like supporting the more 3D (in appearance) skins from Minecraft 1.9.

pithydon:
You could add a skirt.

feminine_mesh

DS-Minetest:
Why that beard?

tobyplowy:
@DS-Minetest xd

DS-Minetest:
Perhaps the beard is only there to show that it also can be more masculine.
I actually like the skirt. They could look really nice, also animated.

tobyplowy:
Jeezz this issue should be called Sam gets plastic surgery XD

tobyplowy:
@DS-Minetest the best thing you don't even need to change his name because it's a boy and a girl's name

tobyplowy:
@Rogier-5 this is my attempt
girl tex
[character girl.zip](https://github.com/minetest/minetest_game/files/723888/character.girl.zip)

tobyplowy:
@Rogier-5 also the model I made is compatible with custom skins

C1ffisme:
@tobyplowy That's actually pretty decent.

@sofar I like the idea of a character creator, although I'd still like to apply my own skins to the character if possible, and it would be good if subgames and mods could customize it. (Would it be possible to put a skin in a folder, and upload it to the server you are connecting whenever you join? Or would that cause issues?)

tobyplowy:
@C1ffisme thx

DS-Minetest:
@tobyplowy This looks actually not bad, good texture and hairs, also the breasts look ok but as @paramat wrote:
> Breasts on player models just don't look right, especially detailed ones as here.
Bothering with this sort of detail is like adding a groin bulge for male players, which would be horrific, or adding a bum-bulge.

tobyplowy:
@DS-Minetest I added breasts and make the arms 3 pixels wide instead of 4 (you can always get rid of the breasts)

tobyplowy:
@DS-Minetest no breasts 
girl tex no boobs

[character girl no b00.zip](https://github.com/minetest/minetest_game/files/724440/character.girl.no.b00.zip)

PR #1530, opened 2017-01-26
Add paint API

juhdanad:
This commit tries to standardize palette-based painting.

In order to make your node paintable, you have to create two functions: one to convert palette index to color, and an other to convert color to palette index.

There is also a default palette: `default_r6g6b6.png` with pre-defined converter functions.

If your node uses this palette, it will be automatically paintable with the dyes.


juhdanad:
This is what the r6g6b6 palette looks like. I hope it will be sufficient for most purposes (however not all colors are accessible with the standard dyes, but this screenshot was made with just mixing dyes)
palettecolors

pithydon:
SO MANY COLORS!

C1ffisme:
@juhdanad Is it possible to make the darkest orange one more brown?

juhdanad:
@C1ffisme see the image file in this commit. That contains all available colors.
The darkest orange node is in fact yellow, because there is no such dark orange on the palette.
You can register your own palette, however.

C1ffisme:
@juhdanad Okay.

mods/dye/init.lua
```diff
@@ -1,3 +1,28 @@
+-- List of palette handlers
+
+palette_handlers = {}
```
kaeza: This should be local or put into `dyes` namespace.

game_api.txt
```diff
@@ -725,6 +725,25 @@ Example of one shapeless recipe using a color group:
...
+
+ * `index_to_rgb(param2) -> red, green, blue` returns the current paint's color
+    from the palette index.
+    0 <= `red`, `green`, `blue` <= 255
+ * `rgb_to_index(red, green, blue, oldparam2) -> param2` returns a palette
+    index for the given color. If the palette does not use all bits of
+    `param2`, the remaining bits are copied from `oldparam2`.
+    0 <= `red`, `green`, `blue` <= 255
+
+To register a palette handler, assign it to `palette_handlers[<palette_name>]`.
```
kaeza: For consistency, you should add a `register_palette_handler` function.

kaeza:
Just a minor point: naming the texture `rgb666` would be better than `r6g6b6`. The former is a more widely used notation.

tobyplowy:
@juhdanad +1

pithydon:
The way this colors wool makes it not look like wool.
Maybe something like this will work https://github.com/pithydon/wool

sofar:
@pithydon what is needed is better texturing. Not the "overlay greyscale with blasting colors". You can make intricate and rich textures with this method, but most of the colored base textures I've seen are just 
"grayscaled" versions which is terribly flat and uninteresting.

Making 63 nodes is the exact thing we want to avoid, too.

paramat:
Actualy, any kind of colourisation, whether texture modifier or hardware colourisation, tends to result in lower quality textures, because it is a single hue applied to one texture with a fixed contrast and brightness. Current wool textures have many different hues and each one has contrast and brightness tuned, which is why they wil always look better than colourised textures.

paramat:
I prefer that a paint API is a separate mod as i see it as a fairly specialist feature. I also think palette based painting is likely to benefit from a variety of implementations so i'm unsure about making this an 'official' API.
I appreciate this contribution though and am sure it's well coded.
Dyes needs some rework too.
Sorry 👎

benrob0329:
If the textures are made specifically for colorization, they could likely be made to look quite good. I think its worth trying, as painting is very handy for building.

juhdanad:
Yes, I plan to improve on coloring, until that this PR is better closed.

PR #1544, opened 2017-02-04
Make drop logic work better for non-8-step-plants

Thomas--S:
Meant as a replacement for #1406 

Now using @paramat's formula:

`base_rarity =  8 - (i - 1) * 7 / (def.steps - 1)`

****
### Old version of this PR
```lua
local base_rarity = 1
if def.steps ~= 1 then
	base_rarity = ( def.steps - 1 ) / ( 0.9 * i + 0.1 * def.steps - 1 )
end
```
Now the drop possibilities result in a linear function with the two fixed points `(1|0.1)` and `(def.steps|1)`.
@paramat What is you opinion about this?

### DIFF
--- a/mods/farming/api.lua
+++ b/mods/farming/api.lua
@@ -330,12 +330,16 @@ farming.register_plant = function(name, def)
 
 	-- Register growing steps
 	for i = 1, def.steps do
+		local base_rarity = 1
+		if def.steps ~= 1 then
+			base_rarity =  8 - (i - 1) * 7 / (def.steps - 1)
+		end
 		local drop = {
 			items = {
-				{items = {mname .. ":" .. pname}, rarity = 9 - i},
-				{items = {mname .. ":" .. pname}, rarity= 18 - i * 2},
-				{items = {mname .. ":seed_" .. pname}, rarity = 9 - i},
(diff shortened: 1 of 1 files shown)


paramat:
> When i = 1 rarities should be 8 and 16.
When i = def.steps rarities should be 1 and 2.

This is no simpler than my formula and doesn't match previous behaviour :-1:

Thomas--S:
Let me explain how I came to this formula:
farming

I made a new PR for two reasons:

- In the old PR I had quite a lot of problems with git. I didn't have the time to solve the problems.
- I made the new PR as a completely new suggestion. I wanted to see what you think about this new formula. If it is disliked, I plan to use your formula, although I personally don't like it very much.

paramat:
Ok no problem opening a new PR then :]

Thomas--S:
@paramat: Good news for you: PR updated to your formula.

paramat:
Ok thanks. I was about to comment that i still disapprove of your formula because it is unintuitive and very difficult to mentally see how it works, as demonstrated by your explanation above. I appreciate your effort though.

paramat:
:+1:

paramat:
Assuming approval by @sofar https://github.com/minetest/minetest_game/pull/1406#issuecomment-263102702

PR #1550, opened 2017-02-07
 Creative: Add /gamemode

zaoqi:


SmallJoker:
This is and shall not be Minecraft :-1:

C1ffisme:
@SmallJoker Changing gamemode within the game is incredibly useful, though. Examples being:

- Make a PVP map and switch to survival without having to close the server to do so.
- Put admin players in creative mode contrast to normal players.
- Switch to creative mode to fix something destroyed by a glitch, then switch back like it never happened.
- ...
- Trolling?

red-001:
👍

Fixer-007:
We will probably have creative priv, will be enough at first.

sofar:
I would like to see a move away from privileges, and using the new player meta attributes instead.

privileges should be reserved for low level, and they don't adequately reflect some of the meanings that we can assign to free typed attributes.

e.g. a server with a mixed game mode can switch mode from "survival" to "adventure" to "pvp" to "creative" to "admin" and only ever test 1 meta attribute.

With privileges, this would require 3 or more privileges to implement, and make for some really awkward code checking all these bits.

We already have problems in minetest where the 'protection_override' priv is wreaking havoc on nicely readable code.

PR #1552, opened 2017-02-08
Leafdecay: node timer based implementation, API.

sofar:
This implements a node-timer based leafdecay mechanism, and exposes
an API to use it in mods.

The API is documented in game_api.txt.

`default.register_leafdecay(leafdecaydef)`

`leafdecaydef` is a table, with following members:

```lua
{
	trunks = { "default:tree"}, -- nodes considered trunks
	leaves = { "default:leaves", "default:apple"}, -- nodes considered leaves
	box = vector(), -- consider this area for finding leaves or trunks
}
```

The algorithm will drop `leaves` items in the area if no `trunks` are found
in the `box` sized area around the position of the leaf. If a node listed
in `leaves` has a group `leafdecay_drop > 0`, then the item is dropped,
otherwise the item is removed but not dropped.

The algorithm also implements a value `default.leafdecay_speed` (default
15) which can be modified to increase or decrease of the leaf decay. The
algorithm will vary the actual speed a bit to introduce randomness.

EDIT by paramat.

This implements the idea in #1545 :

"Event based leave decay instead of ABM.
I know leave decay is a highly discussed feature, because it effects the performance.
I don't know if it was discussed before to use some kind of event based approach. If a tree node is removed (dig, replaced, explosion...) each leave node in a radius of 4 should check for tree nodes and decay if no tree node is near. To "simulate" the old behaviour a random node timer could be used"

Originally leafdecay used a 'trunk cache' to try to reduce processing, this was removed a while ago due to the cache using huge amounts of memory, to compensate the ABM was slowed down but is still very intensive. Also, we have recently incressed the default active block radius so the ABM has become more intensive.

99% of the time the ABM is running for no reason, leafdecay only needs to happen when trunks are removed nearby.
When the ABM is removing leaves it is slow because it has to be tuned to not be too intensive, because it runs all the time.

The nodetimers ensure that all leaves disappear within the specified time of 12sec, instead of a few always remaining for a long time due to random removal (If desired the timer length can be as small as we want for very fast removal of leaves, it can even be instant with alternative code).

### DIFF
--- a/game_api.txt
+++ b/game_api.txt
@@ -658,20 +658,36 @@ default.player_get_animation(player)
 Leafdecay
 ---------
 
-To enable leaf decay for a node, add it to the `leafdecay` group.
+To enable leaf decay for leaves when a tree is cut down by a player,
+register the tree with the default.register_leafdecay(leafdecaydef)
+function.
 
-The rating of the group determines how far from a node in the group `tree`
-the node can be without decaying.
+If `param2` of any registered node is ~= 0, the node will always be
+preserved. Thus, if the player places a node of that kind, you will
+want to set `param2 = 1` or so.
 
-If `param2` of the node is ~= 0, the node will always be preserved. Thus, if
-the player places a node of that kind, you will want to set `param2 = 1` or so.
+The function `default.after_place_leaves` can be set as
+`after_place_node of a node` to set param2 to 1 if the player places
+the node (should not be used for nodes that use param2 otherwise
+(e.g. facedir)).
 
-The function `default.after_place_leaves` can be set as `after_place_node of a node`
-to set param2 to 1 if the player places the node (should not be used for nodes
-that use param2 otherwise (e.g. facedir)).
+If the node is in the `leafdecay_drop` group then it will always be
+dropped as an item.
 
-If the node is in the `leafdecay_drop` group then it will always be dropped as an
-item.
+`default.register_leafdecay(leafdecaydef)`
+
+`leafdecaydef` is a table, with following members:
+	{
+		trunks = {"default:tree"}, -- nodes considered trunks
+		leaves = {"default:leaves", "default:apple"},
+			-- nodes considered for removal
+		radius = 3, -- radius to consider for searching
+	}
+
+Note: all the listed nodes in `trunks` have their `on_after_destruct`
+callback overridden. All the nodes listed in `leaves` have their
+`on_timer` callback overridden.
+
 
 Dyes
 ----
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -325,47 +325,64 @@ default.after_place_leaves = function(pos, placer, itemstack, pointed_thing)
 	end
 end
 
--- Leafdecay ABM
-
-minetest.register_abm({
-	label = "Leaf decay",
-	nodenames = {"group:leafdecay"},
-	neighbors = {"air"},
-	interval = 2,
-	chance = 10,
-	catch_up = false,
-
-	action = function(pos, node, _, _)
-		-- Check if leaf is placed
-		if node.param2 ~= 0 then
-			return
+-- Leafdecay
+local function leafdecay_after_destruct(pos, oldnode, def)
+	for _, v in pairs(minetest.find_nodes_in_area(vector.subtract(pos, def.radius),
+			vector.add(pos, def.radius), def.leaves)) do
+		local node = minetest.get_node(v)
+		if node.param2 == 0 then
+			minetest.get_node_timer(v):start(math.random(20, 120) / 10)
 		end
+	end
+end
 
-		local rad = minetest.registered_nodes[node.name].groups.leafdecay
-		-- Assume ignore is a trunk, to make this
-		-- work at the border of a loaded area
-		if minetest.find_node_near(pos, rad, {"ignore", "group:tree"}) then
-			return
-		end
-		-- Drop stuff
-		local itemstacks = minetest.get_node_drops(node.name)
-		for _, itemname in ipairs(itemstacks) do
-			if itemname ~= node.name or
-					minetest.get_item_group(node.name, "leafdecay_drop") ~= 0 then
-				local p_drop = {
-					x = pos.x - 0.5 + math.random(),
-					y = pos.y - 0.5 + math.random(),
-					z = pos.z - 0.5 + math.random(),
-				}
-				minetest.add_item(p_drop, itemname)
+local function leafdecay_on_timer(pos, def)
+	if minetest.find_node_near(pos, def.radius, def.trunks) then
+		return false
+	end
+
+	local node = minetest.get_node(pos)
+	local drops = minetest.get_node_drops(node.name)
+	for _, item in ipairs(drops) do
+		local is_leaf
+		for _, v in pairs(def.leaves) do
+			if v == item then
+				is_leaf = true
 			end
 		end
-		-- Remove node
-		minetest.remove_node(pos)
-		minetest.check_for_falling(pos)
+		if minetest.get_item_group(item, "leafdecay_drop") ~= 0 or
+				not is_leaf then
+			minetest.add_item({
+				x = pos.x - 0.5 + math.random(),
+				y = pos.y - 0.5 + math.random(),
+				z = pos.z - 0.5 + math.random(),
+			}, item)
+		end
 	end
-})
 
+	minetest.remove_node(pos)
+	minetest.check_for_falling(pos)
+end
+
+function default.register_leafdecay(def)
+	assert(def.leaves)
+	assert(def.trunks)
+	assert(def.radius)
+	for _, v in pairs(def.trunks) do
+		minetest.override_item(v, {
+			after_destruct = function(pos, oldnode)
+				leafdecay_after_destruct(pos, oldnode, def)
+			end,
+		})
+	end
+	for _, v in pairs(def.leaves) do
+		minetest.override_item(v, {
+			on_timer = function(pos)
+				leafdecay_on_timer(pos, def)
+			end,
+		})
+	end
+end
 
 --
 -- Convert dirt to something that fits the environment
--- a/mods/default/nodes.lua
+++ b/mods/default/nodes.lua
@@ -2104,3 +2104,58 @@ minetest.register_node("default:cloud", {
 	sounds = default.node_sound_defaults(),
 	groups = {not_in_creative_inventory = 1},
 })
+
+--
+-- register trees for leafdecay
+--
+
+if minetest.get_mapgen_setting("mg_name") == "v6" then
+	default.register_leafdecay({
+		trunks = {"default:tree"},
+		leaves = {"default:apple", "default:leaves"},
+		radius = 2,
+	})
+
+	default.register_leafdecay({
+		trunks = {"default:jungletree"},
+		leaves = {"default:jungleleaves"},
+		radius = 3,
+	})
+
+	default.register_leafdecay({
+		trunks = {"default:pine_tree"},
+		leaves = {"default:pine_needles"},
+		radius = 3,
+	})
+else
+	default.register_leafdecay({
+		trunks = {"default:tree"},
+		leaves = {"default:apple", "default:leaves"},
+		radius = 3,
+	})
+
+	default.register_leafdecay({
+		trunks = {"default:jungletree"},
+		leaves = {"default:jungleleaves"},
+		radius = 2,
+	})
+
+	default.register_leafdecay({
+		trunks = {"default:pine_tree"},
+		leaves = {"default:pine_needles"},
+		radius = 2,
+	})
+end
+
+default.register_leafdecay({
+	trunks = {"default:acacia_tree"},
+	leaves = {"default:acacia_leaves"},
+	radius = 2,
+})
+
+default.register_leafdecay({
+	trunks = {"default:aspen_tree"},
+	leaves = {"default:aspen_leaves"},
+	radius = 3,
+})
+


sofar:
As opposed to #1552.

sofar:
Attribution to @MarkuBu since he came up with the nodetimer idea, and I really only objected against his API (and that's what I fixed). The flow and execution is mostly his code.

thePalindrome:
This does seem a bit cleaner compared to the other one, and I do prefer the functions that are exposed here.

I do prefer examining a chain of nodes to find the tree, but in terms of direct comparison to #1548, this one is much better imo

sofar:
@thepalindrome

So, we can actually do this, and nonrecursively. The code sits in `default` and is part of `nodeupdate`, and a next iteration of this code should probably use it to path.

Secondary, digging leaves should potentially also trigger leafdecay as well, which this also does not do yet.

MarkuBu:
:-1: for the box. This changes the decay behavior. You have to remove ANY trunk node, even randomly spawned logs. There are too many leaves left. Most worse are jungle trees. 

You expect that decay starts, but nothing happen because there is somewhere a hidden log

And you need to know the tree size. Might be difficult if you just have a mts file without any further information

Last but not least you have to loop over a few dozen nodes, nearly 200 at a jungle tree. And I want to add much bigger trees to my game where you will have over 1000, maybe much more, nodes. I don't want to chop down such a tree with this box thing. It may work for the default trees, but not for what I want to do

screenshot_20170208_074352


screenshot_20170208_072054

screenshot_20170208_072350

MarkuBu:
And finally it doesn't work with v6 trees

screenshot_20170208_082540

screenshot_20170208_082729

MarkuBu:
Maybe you didn't understand my intention with the radius and lradius, but it works better than everything else

You don't need two different values. Just enter the expected leave decay radius (e.g. 2 for all v7 trees) and add one for the search box around the trunk. This works better than the old ABM and much better than your tree bounding box

And I'm still not happy with your register_leavedecay function. The MTG code is a mess and this just add's a little bit more

sofar:
@markbu nothing unfixable, the `if mg_v6` that I omitted in the initial version won't look as bad as your patch with 5x the mapgen check, and we just need to tune the box size.

I can incorporate `radius` and `lradius` in my patch without exposing it like your patch does, another good reason for a more opaque API design that doesn't force the mod writer to account for them and pass them around in function calls themselves.

As discussed in the thread above (if you didn't see it) we can optimize the algorithm further based on code that was done by me for falling nodes (a factor 100 or so improvement) and that code actually can detect connections to logs properly and eliminate floating islands entirely, but optimizations like that should be secondary, and there's nothing wrong with doing optimizations later.

This isn't personal, it's technical. I have just as much "right" to propose code as you. I encouraged you to submit the PR in the first place because your idea was good, we just disagree on the details, but guess what, those details are important to people.

I even credited you. Do you want me to remove that credit?

MarkuBu:
@sofar of course it is nothing personal. And of course you have the right to propose code. But I spend hours to write and test the code. The code might be not perfect, but it works perfect within the limits of Minetest and with the spirit of voxel games in mind, with minimal resources and the best result you can expect in a voxel game

So there was no need to replace the code with something bloated that doesn't work as well as my code. It just needs some polish. That's why I'm upset.

I do not talk about the API. I have some other concerns about that. But it fit's into the MTG code structure

sofar:
I'm going to ignore the non-technical complaints by @markbu and focus on the technical aspects.

> Maybe you didn't understand my intention with the radius and lradius, but it works better than everything else

It indeed works fairly well, and the values you used seem to have been good.

So I have changed this PR to reflect the exact same algorithm and values as you proposed, and updated the API to account for it. I've also accounted for mg_v6 exactly in the same matter as you have, tuning the radii where needed.

tested and pushed. If this needs further tuning please chime in.

mods/default/functions.lua
```diff
@@ -325,47 +325,69 @@ default.after_place_leaves = function(pos, placer, itemstack, pointed_thing)
...
-			return
-		end
+-- Leafdecay
+default.leafdecay_speed = 15
+local function leafdecay_after_destruct(pos, oldnode, def)
+	local speed_min = (default.leafdecay_speed * 0.75) * 10
+	local speed_max = (default.leafdecay_speed / 0.75) * 10
+	for _, v in pairs(minetest.find_nodes_in_area(vector.subtract(pos, def.leaves_radius),
+			vector.add(pos, def.leaves_radius), def.leaves)) do
+		minetest.get_node_timer(v):start(math.random(speed_min) / 10,
```
paramat: set(timeout,elapsed)

MarkuBu: Why set() and not start()?

paramat:
I oppose a setting for decay speed as this is so trivial, settings should only be added when really needed.

The default max time of 20s seems far too long, the current leafdecay delay is only due to the necessary slowness of the ABM method, it's not for any good reason, it actually makes more sense that leaves come down instantly as you remove braches, but the timer delays are useful for spreading load.

I feel we don't need randomised min and max times. Having some leaves decay immediatly on trunk removal is fine as explained above. Also it's better to vary the decay time from 0 to MAX because that spreads out the trunk searches and lighting updates over a larger timespan.

I feel the check for param2 should be in `leafdecay_after_destruct()` so that for non-zero param2 we don't even start the timer.

I also didn't like the `box` method, but you've just changed that.

As discussed in the ither PR the radius values can be combined into one and must be set to the max distance of a leaf from a trunk for each tree type. See the other PR for the necessary radius values (i still need to check these values). Trunk radius = 2 is too small for some trees.

I don't like the overriding of nodes, overriding is messy and should be avoided if possible. I see no good reason for this apart from saving a few lines of code. The simpler method is clearer because you see it in the nodedef itself. Instead of using the API modders can just copy-paste the lines in nodedef, which is no more difficult than registering leafdecay.

Overall i think the other PR is much closer to acceptable so i am supporting that, sorry :-1:

paramat:
> I do prefer examining a chain of nodes to find the tree

Might not be possible due to tree overlap.

sofar:
Fixed mapgen version check.

mods/default/functions.lua
sofar: Restarting the nodetimer like this is probably more efficient. We _can_ fetch an existing node timer, but that's just extra work, and the second `start()` is guaranteed anyway.

sofar:
> I feel the check for param2 should be in leafdecay_after_destruct() so that for non-zero param2 we don't even start the timer.

Yes, absolutely, good point.

sofar:
> I oppose a setting for decay speed as this is so trivial, settings should only be added when really needed.

It's not a setting. It is a subgame tunable variable. A setting is retained between different worlds, this is not.

I'd made it a local but I think there's value in overriding it in subgames. But it isn't worth overriding in the settings menu, that I agree with you.

sofar:
>I feel we don't need randomised min and max times. Having some leaves decay immediatly on trunk removal is fine as explained above. Also it's better to vary the decay time from 0 to MAX because that spreads out the trunk searches and lighting updates over a larger timespan.

Contradiction. You say "we don't need randomised min and max times" but then you say "better to vary the decay time from 0 to MAX". You just replaced '0' with 'min' :)

So the discussion is whether to range the random value from (0 - max) instead of (min - max).

I strongly prefer not starting at 0 (it's too immediate), and 20 max is indeed a bit long, but what we can do as a concession is to change the min/max so they're in the middle of what we both put.

old: 15-20
you want: 0 - 10 (assuming here, you didn't state what you prefer for 10)

my counter offer: 5.4 - 15.0 (using '9' and '0.6' as factors in the math)

sofar:
Pushed:
- param2 check before starting timers.
- tune min/max times to vary between 5.4s - 15s.

sofar:
> Overall i think the other PR is much closer to acceptable so i am supporting that, sorry

I'm really disappointed to hear this, considering I've been more than clear that the other API is unacceptable to me.

paramat:
Well i need to rethink your API points which are likely correct, but there are other issues with your PR, so perhaps we need a mix of both.

sofar:
> As discussed in the ither PR the radius values can be combined into one and must be set to the max distance of a leaf from a trunk for each tree type. See the other PR for the necessary radius values (i still need to check these values). Trunk radius = 2 is too small for some trees.

yes, I've noticed that the second radius effectively is the other `radius - 1`. I'll definitely consider changing that, and it does clean up the API even further.

thePalindrome:
For leaf decay speeds, I'd recommend 4/20s - 11/20s, which follows the mod "Fast Leaf Decay," which I find near universal among modpacks for the other voxel game :P

mods/default/functions.lua
paramat: Nevermind my comment, i could not find docs of a 2-argument `start` so assumed this was an error and was meant to be `set`.

paramat:
Ok i'm liking this again, more input coming.

paramat:
Leafdecay radii:

Appletree: mgv6 2, others 3 (due to randomised branches)
Jungletree: mgv6 3 others 2
Pine: mgv6 3 others 2
Acacia: others 2
Aspen: others 3 (as the trunk stops 3 nodes below top of tree, this could perhaps be changed in the schematic to 2 below)

Please can we use one radius for both searches, as this minimises the search radii while still removing all leaves. The necessary radii values are above.

paramat:
For decay time, i watched markbu's video which had a time of 0-10s and seemed about right.
I prefer that the minimum is 0 as this spreads out the decays the most, spreading the load of trunk searches and lightng updates more.
If i assume that timers work if within the active block radius of 3 * 16 = 48 nodes, walking at a speed of 4n/s a player can be out of range in 12s, i think all leaves should decay before a player leaves the active radius. A player may already be a few nodes away when digging so 10s seems a rough maximum for decay time.

paramat:
> Pushed:
param2 check before starting timers.

This is not showing up for me in Github yet.

sofar:
> This is not showing up for me in Github yet.

I Must have missed something. I pushed again and it shows it updated now.

sofar:
Let's consider this WIP for now.

sofar:
Fixed up nodetimer:start() call, which was entirely nonsense (but, due to random, worked ok).

I took out the math and config option, it's now hardcoded between 2.0 and 12.0 seconds.

sofar:
Reduced the API to only involve `radius` once. I still use the `-1` compensation for the second search step. It seems to work, so I don't see a reason to increase it, and I don't want to reduce the radius of the initial search. This is effectively not a functional change, just an API simplification.

MarkuBu:
@sofar sorry about yesterday. I had a bad day.

I suggest that you enter the leaves decay radius to  the def and add 1 to the search box around the trunk like I did. There is no technical reason, it is just a bit of security. That's why a leave decay radius of 2 is enough for v7 apple trees

MarkuBu:
@sofar I don't wanna bother you, but what if a node has already an on_timer callback, e.g. a fruit that should ripen over time? The API will overwrite this without a warning

Node timers become more popular and the chance rises, that on_timer is already used

sofar:
@markbu yes, that indeed happens, but we can consider this:

1) it's documented in the API explicitly, so it's not "without a warning" at all.
2) after calling register_leafdecay, the mod can overwrite it again with code that handles growth and leafdecay.

Having a plant growth nodetimer server a dual function is going to be fairly complex, and any such code most likely should implement a custom version of leafdecay to maintain e.g. a growth time counter, otherwise you could make fruit grow one stage by chopping down a tree.

There is a third solution: We could actually pass on a timer call back to the old handler if there was no leafdecay. We use this mechanism in several locations, The callee will still have to have a more than simple on_timer function to account for accidental timer ticks calling it.

paramat:
Yes i recommend using `radius` for the trunk search and `radius + 1` for the leaf search, and registering `radius` 1 smaller. This way the registered value has meaning because it is set to the maximum distance of a leaf node from a trunk node = the former leafdecay radius.

However it seems pointless to increase the leaf search radius by 1 because the extra leaves are unaffected by the trunk you just removed. It doesn't clean up missed leaves because the code ensures all leaves are removed. So this is a big increase in search volume for no reason.

Some of the radii values are wrong, see my comment above, appletree needs to be inside the mgv6 check.

Shall we make the aspen trunk extend 1 node further up the tree to enable a smaller radius to be used? Currently it is:
```
		_, _, T, _, _,
		_, _, T, _, _,
		_, _, T, _, _,
		_, _, T, _, _,
		_, _, T, _, _,
		_, _, T, _, _,
		_, L, T, L, _,
		L, L, T, L, L,
		_, L, T, L, _,
		L, L, T, L, L,
		_, L, T, L, _,
		L, L, L, L, L,
		_, L, L, L, _,
		_, _, L, _, _,
```
I'm happy to make this update if you agree.

Overall though i'm fairly happy with this PR.

paramat:
Maybe even better i can alter the non-mgv6 schematics subtly so that all of them only need radius 2.

paramat:
I'm a little unhappy with new appletrees anyway. My thought is to extend the centre trunk up into the canopy. For jungletree if i make just one of the top branches full probablility that will be enough, but maybe i'll make  2 opposite branches full probability as 1 branch will look weird.

sofar:
@paramat even though you say that we should make this as efficient as possible and the search area as small as wel can, I think that you're underestimating the improvement that this patch already gives.

This is going to save us millions of clock cycles in the long run. 

This is going to save us millions of clock cycles in the long run, even if the performance of the algorithm sucks.

Please, understand, that we're getting rid of a perpetual algorithm that functions when nothing changes in the world, and hits us for every, every, every damn loaded leaf on the map.

The new algorithm will not use a single clock cycle *ever*, until the heat death of the universe, if nobody cuts down a tree.

So, just because of that, I think we can safely say that searching 150 extra nodes when a tree is cut down is not, I repeat, NOT, a deal breaker.

So heightening the trunk in a schematic isn't that much gain. The map around the player is entirely hot already. A mesh update is already going to happen.

So I would politely suggest we just tune the radius values for each tree so that it works for the existing trees, and then later think about how to refine schematics. :)

So, with that said, let's work on those radii?!

sofar:
So I made one more adjustment:

radius is used without increment/decrement everywhere now.

This exposed an issue with Aspen and Apple tree, as paramat noted. So I increased the radius for those trees by 1.

MarkuBu:
> This is going to save us millions of clock cycles in the long run.

How about lava cooling ABM? This should be more heavy because it runs evey second

paramat:
Sofar yes i understand all that and agree with you.

> So I would politely suggest we just tune the radius values for each tree so that it works for the existing trees, and then later think about how to refine schematics. :)

Yes this is fine.

My only serious request was that radius is not increased by 1 for the leaf search, which you have now done.
Appletree can be radius 2 in mgv6 (i checked mgv6 treegen) so you can move that inside the conditional.

sofar:
ahh apple tree for v6. Yes, I'll put that one to `2`. (I thought you meant v7, but we need 3 there otherwise it leaves crowns).

Pushed.

paramat:
:+1:

MarkuBu:
I'm still not happy with that implementation. It might work for default trees, but now, that the l-system light bug is fixed, much more large trees will come to the game.  And currently I build large trees by myself

screenshot_20170213_092523

It is possible to build those trees now because they don't want to decay by default, but with the current implementation they don't decay too. The top leaves are up to 6 or 7 nodes away and I can't make the center trunk higher because this will look odd.

Even with my approach it won't work well. We can leave it like it is and let the player decide what he wants to do with the non decaying leaves, or rethink the algorithm

sofar:
Yes, but the old leafdecay would work just as bad, and be much more inefficient. With the new efficient method, it would be far less problematic tuning the radii up to 6 or 7. So it's not that the new code falls short, it's that it doesn't solve the problem on a more fundamental level.

MarkuBu:
The old leafdecay wouldn't work because a lot of the leaves would decay immediately

But increasing the search radius may cause some unwanted decay on neighbour trees or the leaves on top of the tree decays if you start digging the trunk from the bottom.

I will try that if I can place those trees on a map

paramat:
Hm? I don't understand the problem. Do you mean the lack of leafdecay if large l-trees are added that use normal nodes? That problem exists now also, mod nodes should be used with customised radii, and probably already are.

mods/default/functions.lua
```diff
@@ -325,47 +325,64 @@ default.after_place_leaves = function(pos, placer, itemstack, pointed_thing)
...
-	action = function(pos, node, _, _)
-		-- Check if leaf is placed
-		if node.param2 ~= 0 then
-			return
+-- Leafdecay
+local function leafdecay_after_destruct(pos, oldnode, def)
+	for _, v in pairs(minetest.find_nodes_in_area(vector.subtract(pos, def.radius),
+			vector.add(pos, def.radius), def.leaves)) do
+		local node = minetest.get_node(v)
+		if node.param2 ~= 1 then
```
SmallJoker: Why was it changed to `~= 1`? Before: `~= 0`

mods/default/functions.lua
```diff
@@ -325,47 +325,64 @@ default.after_place_leaves = function(pos, placer, itemstack, pointed_thing)
...
-			if itemname ~= node.name or
-					minetest.get_item_group(node.name, "leafdecay_drop") ~= 0 then
-				local p_drop = {
-					x = pos.x - 0.5 + math.random(),
-					y = pos.y - 0.5 + math.random(),
-					z = pos.z - 0.5 + math.random(),
-				}
-				minetest.add_item(p_drop, itemname)
+local function leafdecay_on_timer(pos, def)
+	local list = minetest.find_nodes_in_area(vector.subtract(pos, def.radius),
```
rubenwardy: find_node_near would be better, as

1. the old code used it, so same behaviour
2. it early exits
3. the def says "radius"

also: what if they're on the border of ignore?

mods/default/functions.lua
paramat: Should be '== 0'?

mods/default/functions.lua
```diff
@@ -325,47 +325,64 @@ default.after_place_leaves = function(pos, placer, itemstack, pointed_thing)
...
+			vector.add(pos, def.radius), def.trunks)
+	if #list > 0 then
+		return false
+	end
+	local node = minetest.get_node(pos)
```
rubenwardy: Would be nice to have a space here above here

mods/default/functions.lua
```diff
@@ -325,47 +325,64 @@ default.after_place_leaves = function(pos, placer, itemstack, pointed_thing)
...
+				not is_leaf then
+			minetest.add_item({
+				x = pos.x - 0.5 + math.random(),
+				y = pos.y - 0.5 + math.random(),
+				z = pos.z - 0.5 + math.random(),
+			}, item)
+		end
 	end
-})
+	minetest.remove_node(pos)
```
rubenwardy: would be nice to have a space above here

mods/default/functions.lua
```diff
@@ -325,47 +325,64 @@ default.after_place_leaves = function(pos, placer, itemstack, pointed_thing)
...
-	chance = 10,
-	catch_up = false,
-
-	action = function(pos, node, _, _)
-		-- Check if leaf is placed
-		if node.param2 ~= 0 then
-			return
+-- Leafdecay
+local function leafdecay_after_destruct(pos, oldnode, def)
+	for _, v in pairs(minetest.find_nodes_in_area(vector.subtract(pos, def.radius),
```
rubenwardy: see note about `find_node_near`

mods/default/functions.lua
```diff
@@ -325,47 +325,64 @@ default.after_place_leaves = function(pos, placer, itemstack, pointed_thing)
...
+	minetest.check_for_falling(pos)
+end
 
+function default.register_leafdecay(def)
+	assert(def.leaves)
+	assert(def.trunks)
+	assert(def.radius)
+	for _, v in pairs(def.trunks) do
+		minetest.override_item(v, {
+			after_destruct = function(pos, oldnode)
```
rubenwardy: what about old after_destruct callbacks?

game_api.txt
```diff
@@ -658,20 +658,35 @@ default.player_get_animation(player)
...
+`leafdecaydef` is a table, with following members:
+{
+	trunks = {"default:tree"}, -- nodes considered trunks
+	leaves = {"default:leaves", "default:apple"},
+		-- nodes considered for removal
+	radius = 3, -- radius to consider for searching
+}
+
+Note: all the listed nodes in `trunks` have their `on_after_destruct`
+callback overridden. All the nodes listed in `leaves have their
```
rubenwardy: missing `

game_api.txt
```diff
@@ -658,20 +658,35 @@ default.player_get_animation(player)
...
+the node (should not be used for nodes that use param2 otherwise
+(e.g. facedir)).
+
+If the node is in the `leafdecay_drop` group then it will always be
+dropped as an item.
+
+`default.register_leafdecay(leafdecaydef)`
+
+`leafdecaydef` is a table, with following members:
+{
```
rubenwardy: indent the code one level in, and place blank lines either side

```
`leafdecaydef` is a table, with following members:

	{
		trunks = {"default:tree"}, -- nodes considered trunks
		leaves = {"default:leaves", "default:apple"},
			-- nodes considered for removal
		radius = 3, -- radius to consider for searching
	}

Note: all the listed nodes in `trunks` have their `on_after_destruct`
```

mods/default/functions.lua
paramat: But we need the list of positions here.

mods/default/functions.lua
paramat: This can be `find node near`

paramat:
+1 when my line comments are addressed.

mods/default/functions.lua
sofar: correct, fixed.

mods/default/functions.lua
sofar: right, can't use it here.

on ignore, we'll start a node timer for an ignore node, and nothing happens.

mods/default/functions.lua
sofar: We can potentially store and call it, but the API clearly states it will be overwritten, nor do I expect there to be a use for it.

game_api.txt
sofar: ok, sure

game_api.txt
sofar: ty

paramat:
Might be worth doing register_leafdecay automatically with these

-- all_legacy_nodes is an array of node names
default.register_leafdecay({
	trunks = all_legacy_nodes,
	leaves = {"group:tree"},
	radius = 3,
})

Seems messy, i'd prefer disruption and asking mods to update.

mods/default/functions.lua
sofar: actually, manually placed leaves have param2 =1 (through place_param2) which stops leaf decay. Placing with `shift` causes them to decay normally, since that code puts param2 at `0`.

Here we start the leafdecay timer for nodes that need to decay, so `== 0` is good, and so is `~= 1`.

I'll change it to `== 0` though, since that's a bit simpler to understand.

mods/default/functions.lua
paramat: Fine for me as is.

paramat:
game_api.txt needs updating, otherwise +1.

sofar:
Oh, I though I had, but I had forgotten to `git add` that part :).

pushed

paramat:
@sfan5 @Ekdohibs @ShadowNinja

paramat:
Tested and excellent :+1:

paramat:
60e5b299af564f250af8304b5146f79693c314ff

PR #1564, opened 2017-02-12
register_rail: Fix the table overwrite behaviour

SmallJoker:
The function used the default table to overwrite the new definitions, which is exactly the wrong direction.

Already merged in boost_cart: https://github.com/SmallJoker/boost_cart/commit/cba9699


sfan5:
:+1:

sofar:
Yeah, that is the wrong way around :)

PR #1569, opened 2017-02-14
Generate dyes based on html color strings using a base and overlay texture.

treytomes:
I've separated out the dye changes from the other changes in my last pull request, and made the suggested code changes.  I was also able to remove the helper functions by making a change to the dye_overlay.png image.

////////////
EDIT by paramat:
Was part of related #1566 

1. All dyes are generated dynamically from the dye.dyes table, which I added html color values to.  More dyes can be created simply by adding to this table.
2. All colors are now located in a single "color_"..<color name> group.  It may simply be my own ignorance, but having 3 different color sets seemed to be needlessly complicating color usage.
3. Replaced the texture set with 2 textures: dye_base and dye_overlay.  These are combined into a single texture using the html colors and some rgb<-->hsv helper functions that I added.


treytomes:
Realized upon further testing that I was failing to add dyes to the "dye" group, so crafting colored wool wasn't working.  The 2nd changeset corrects this issue.

paramat:
This will break texture packs, maybe use the existing multiple textures?
Textures are very small files and we only have 14 of them so there is little advantage to using colourisation.

treytomes:
The advantage to this method is that it is easy to dynamically generate more dye colors based solely on the code.

treytomes:
I think that I understand your concern, as this would basically reduce texture customization in favor of easier code customization.  I guess it comes down to a preference on what direction this game is going to take.  I like  doing as much as possible from the code, but that doesn't necessarily make it the best way.

paramat:
I actually like the use of colourisation here, no-one is that bothered with being able to fine-tune the appearence of individual dye textures, it's just texture pack breakage that is the issue, in the long term it may be worth it though.

treytomes:
When I first attempted dynamically colorizing these textures, I tried applying a color with the alpha set to full and to half.  In both cases, the alpha value was ignored and the color overwrote every non-transparent pixel in the destination texture.

For example, in an expression such as this:
`inventory_image = "(dye_white.png^[colorize:"..html..")"`
...where html="#FF000080", or html="#FF0000FF", the image is completed converted to bright red.  Perhaps there is an aspect to the colorize keyword that I don't understand?  I would be happy to convert this changeset to base the dynamically generated textures off the dye_white.png texture, if only I could solve this problem.

sfan5:
Did you consider using the recently added `[multiply:` [modifier](https://github.com/minetest/minetest/blob/master/doc/lua_api.txt#L422)?

treytomes:
I did not.  I learned the syntax for texture expressions from http://dev.minetest.net/texture, which has no reference to multiply.  I will give that a try.

treytomes:
Is the `[multiply:` modifier available in v0.4.15?

paramat:
The wiki is often out of date, best look at lua_api.txt also.

sfan5:
@treytomes no but minetest_game git is supposed to be used with minetest git so it's okay to use it.

treytomes:
Thanks; I'll need to get a C++ environment setup then.  If I can get `[multiply:` to do what I think it should do, I'll be able to submit another changeset in the next couple of days.

VanessaE:
Shouldn't hardware colorization be preferred over texture compositing?

HybridDog:
VanessaE, as far as l know, hardware colorization doesn't work on inventory images.

VanessaE:
Apparently, it does, https://github.com/minetest/minetest/blob/master/doc/lua_api.txt#L3830-L3833

paramat:
But not for param2 colouring.

tacotexmex:
As texture pack maintainer I'd like to state that colored series of nodes might not always be fully dyed in a single color, depending on style. It might be that dither or other methods of alternating pixel colors are utilized. Therefore I find this method quite imposing, limiting the choices for when drawing new textures and making certain already drawn styles impossible to maintain.

paramat:
^ Yes that's my concern with wool. as stated in #1566

numberZero:
A nice idea, but...
Before:
dyes-old
After:
dyes-new

The difference is obvious. I certainly don’t like it.

Note: colorization may be softer when necessary, as here for bronze, copper:
colorization-example-1
gold and iron:
colorization-example-2

paramat:
I agree, :-1:

tacotexmex:
As much as I like the idea in general, I'd vote for leaving this headache to the texture makers.

paramat:
PR is quiet, any more opinons?

paramat:
This will break texture packs, although the fix is easy, and reduces the flexibility and visual appeal of the dye item textures, although that's not a big issue.
It becomes easy to add new dye colours, but this would have to then lead to reworking wool and beds mods to cope with the possible new dye colours, this would break wool and bed textures in texture packs too, and means that wool and bed textures would have to use colourisation, which reduces texture flexibility (many current wool textures have a variety of hues).

I don't feel strongly but i'm more -1 than +1.

numberZero:
Until someone finds a way to generate better dye inventory images, such change should not be merged. I know several things that may help, like using several layers, adjusting colors before using for colorization, etc. but really, experimenting with that requires way more time that I’d like to spend on this. PR author doesn’t seem to be interested in this PR anymore, either.

Also, it would be nice would there be a way to use manually drawn textures, if present, and fall back to generated ones. But generated textures should look acceptable anyway, while in this PR they don’t.

paramat:
Actually the athor has not updated this as they said they would (which is fine) and it's been months, so closing for now.

PR #1588, opened 2017-02-26
Make inventory crafting arrow customizeable

MarkuBu:
This PR adds a own arrow for the crafting field if a texture pack wants to have different images

Related issue https://github.com/minetest/minetest_game/issues/1538


stujones11:
Using the the furnace arrow was cheap IMO so I agree to this in principle. However, if you are going to give it its own image then you should rotate it correctly to avoid the need for modifiers.

MarkuBu:
Hope this was right now. I'm a little bit confused with all the git branches and subgames. I'm still not good with git

stujones11:
I would suggest perhaps a less generic name, like `gui_arrow_craft_bg` to make it more clear what it's used for but maybe I am being picky, I'd see what the devs think before changing it.

sofar:
The arrow is generic, the furnace one can just refer to the generic icon?

It wouldn't be so bad to have a bunch of these standard gui elements around anyway.

stujones11:
> The arrow is generic, the furnace one can just refer to the generic icon?

I think the point was that texture packs could differentiate them, though I have a tendency to agree.

paramat:
> The arrow is generic, the furnace one can just refer to the generic icon?

I agree, i doubt we need 2 different arrows. I would prefer removing the 'furnace arrow' and have both use the new generic one.

stujones11:
If there is to be only one arrow then it would be more conventional for it to point upwards IMO, you may as well just rename the file and leave the rest unchanged.

paramat:
Up would not be used, better for the texture to be either right or down, whichever seems more commonly used, probably right for crafting.

stujones11:
Unless this actually fixes something then it's a rather pointless change IMO as it would not address the related issue.

octacian:
IDK, I don't think this is worth doing. I like the concept of renaming to the generic `gui_arrow_bg.png`, but again, this would break mod and TPs. On the other hand, I don't feel as though we should be letting backwards compatibility hold us back so much.

There's three solutions IMO:
* Merge as is (just deal with breakage)
* Keep old image as deprecated until next release
* or, ofc, don't merge at all (even if this means this is just put on hold until a later release)

MarkuBu:
The TP argument doesn't count I think. They can add support for this, just like mods have to add support for the new leaf decay. And even if they don't, it still works

Fixer-007:
You mean this arrow?
default

octacian:
@Fixer-007 yes, that arrow.

Wait, doesn't the crafting output use this arrow too? If yes, why haven't you updated the code for that?

DS-Minetest:
> The TP argument doesn't count I think.

True. https://github.com/minetest/minetest_game/pull/793#issuecomment-285392313

paramat:
:-1: I don't think anyone will need different arrow textures for furnace and inventory, better to have one as now, and possibly rename it to just 'gui_arrow_bg' if we decide texture pack breakage is worth it.

sfan5:
@paramat 
no, a texture pack author might want to add a text `smelt` or `craft` above the arrow.
just because you think we don't need different ones doesn't mean we shouldn't allow texture packs to have different ones.

SmallJoker:
In order to allow texture packs to be more flexible, I agree with this idea.
I think that breaking the texture packs that already customized this texture is not a significant issue. Even if the TP author is not active anymore, this can be fixed easily by hand.

tobyplowy:
@sfan5 agreed

paramat:
> a texture pack author might e.g. want to add a text smelt or craft above the arrow.

Good point.
Should this be called 'crafting arrow' then?

paramat:
I'm neutral, but if merged i think it should be called 'crafting arrow'.

paramat:
Needs to be called 'gui_furnace_arrow_bg.png'.

paramat:
@MarkuBu ^ i might +1 if texture name is changed.

MarkuBu:
I'm sorry, I deleted the repo. Will close this and maybe make a new PR later

PR #1596, opened 2017-02-28
Add desert/silver sandstone-related blocks.

kaeza:
This PR adds a few new blocks:

* Desert sandstone, sandstone brick, and sandstone block.
* Silver sandstone, sandstone brick, and sandstone block.
* Stairs and slabs for every node above.

In total, 18 nodes are added (6 blocks, 6 stairs, 6 slabs).

screenshot

EDIT by paramat:
Silver sandstone textures have been updated to this:

screenshot_20170305_032426


paramat:
These look good. Silver sand is sort of meant to be a breakdown of stone so unsure about that one, although it looks good. +1 for an orange sandstone.

Fixer-007:
That will be great for builders, I already liking it.

GreenXenith:
:D So why is it that there was so much debate earlier about this being added on #1477 but here it is instant yes? Or was there a discussion about this that I missed?

kaeza:
>So why is it that there was so much debate earlier about this being added on #1477 but here it is instant yes?

Magic.

GreenXenith:
Well I thank you none the less! :D

paramat:
How about adding a 'red sand' so we can have a full set in red, orange and yellow? (even though desert stone is not a sandstone).
I'm not keen on adding the silver sandstone, i feel if we add a pale stone it would be best to have some other kind of stone, not a sandstone.

paramat:
Please can we use the non-combined form for 'sandstonebrick', so 'sandstone_brick'?
There's a note in nodes.lua about using non-combined words for all new nodes even if this is inconsistent with previous nodes. This would need a change to your registering.

GreenXenith:
@paramat so you are saying that we will have 4 sand types now? Silver (silversand), Yellow (beachsand), Orange (desertsand), and Red (?)? Or did I misunderstand?

paramat:
Yes 4 types. Sorry for my previous resistence. There's a high chance i will use some of these new nodes in biomes.

GreenXenith:
What will this new type be called?

paramat:
'red_sand' hehe.

kaeza:
More discussion on the IRC logs [here](http://irc.minetest.net/minetest-dev/2017-02-28#i_4822340).

paramat:
Ok 'sandstonebrick' is fine for all.

paramat:
The silver sandstone seems to have just a little too much contrast in it's texture-noise, the others are more subtle.

GreenXenith:
@paramat I agree.

DS-Minetest:
> 'sandstonebrick'

This brought me to the idea that red sand could be some sort of pulverized brick (the one made out of clay).

mods/default/crafting.lua
```diff
@@ -494,37 +494,50 @@ minetest.register_craft({
...
-minetest.register_craft({
-	output = 'default:sandstone_block 9',
-	recipe = {
-		{'default:sandstone', 'default:sandstone', 'default:sandstone'},
-		{'default:sandstone', 'default:sandstone', 'default:sandstone'},
-		{'default:sandstone', 'default:sandstone', 'default:sandstone'},
-	}
-})
+local function register_sandstone_recipes(prefix)
+
```
rubenwardy: We don't have spaces at beginning of blocks

paramat:
Desert stone is a non-sandstone due to it's texture.

DS-Minetest:
> Desert stone is a non-sandstone due to it's texture.

It also has other properties of stone and not sandstone, eg. the crackyness but not crumblyness.

paramat:
@kaeza thought this might be useful information.
The silver_sand texture was created from default_sand.png.
In Gimp:
colours->hue-saturation->(click master) saturation negative 75
So it has a very slight yellow hue, making it a warm light grey.

paramat:
Testing your new texture in Gimp, it has a slight blue hue, better that it has a slight hue of default_sand.png.

paramat:
This is fine, new textures are close enough for now. As discussed on IRC i will tweak the silver sandstone textures later (as requested by author).

mods/default/crafting.lua
```diff
@@ -503,26 +503,90 @@ minetest.register_craft({
...
+		{"default:sandstone", "default:sandstone", "default:sandstone"},
+		{"default:sandstone", "default:sandstone", "default:sandstone"},
+	}
+})
+
+minetest.register_craft({
+	output = "default:desert_sandstone",
+	recipe = {
+		{"default:sand", "default:sand"},
+		{"default:sand", "default:sand"},
```
paramat: desert_sand

mods/default/crafting.lua
```diff
@@ -503,26 +503,90 @@ minetest.register_craft({
...
+		{"default:desert_sandstone", "default:desert_sandstone", "default:desert_sandstone"},
+		{"default:desert_sandstone", "default:desert_sandstone", "default:desert_sandstone"},
+	}
+})
+
+minetest.register_craft({
+	output = "default:silver_sandstone",
+	recipe = {
+		{"default:sand", "default:sand"},
+		{"default:sand", "default:sand"},
```
paramat: silver_sand

paramat:
2 line comments, otherwise +1.

kaeza:
Just in case, silver sandstone, sandstone block and brick textures are from [GreenXenith/sandplus](https://github.com/GreenXenith/sandplus).

License says CC-0.

paramat:
Just needs texture credits in README and if needed texture authors added to license.txt under the corresponding licenses. For CC-0 we re-license these to CC BY-SA 3.0 for use in MT.

kaeza:
Is that OK?

paramat:
:+1:

sfan5:
:+1:

rubenwardy:
I'd like to have @sofar's opinion on this before merging

Fixer-007:
About silver_sandstone_brick texture, I would modify one tiny detail here:
default
That one white pixel makes it confusing, I suggest to darken it a bit to make it coherent.

paramat:
I'm going to update the silver sandstone textures, on request of the PR author. Apart from the issue you point out here, the slight hue is wrong, it is blue instead of yellow (see comments above).

paramat:
The dark pixel you mention is because there are diagonal bevels at the corners and the light is coming from top left, so the lower right corner bevel is darkest.

PR #1637, opened 2017-03-14
Creative: Cache creative mode setting

zaoqi-unsafe:


SmallJoker:
Why `Don't merge`? Commit looks good to me.

octacian:
Why do you keep opening seemingly identical PRs?

PR #1642, opened 2017-03-16
Fix crash on keys without meta.

sofar:
If you /giveme a default:key, it will crash on use on a door.

We can harden it a little bit by parsing the old metadata string
and if that errors (which will show in singleplayer/local server)
we just error, but not crash the client.

### DIFF
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -562,7 +562,11 @@ function default.can_interact_with_node(player, pos)
 		local key_meta = item:get_meta()
 
 		if key_meta:get_string("secret") == "" then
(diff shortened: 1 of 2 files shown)


sofar:
@octacian 

fixes #1640 
alt fix  #1641

octacian:
Latest commit in my version prevents the additional possibility of a crash as does yours, however, in a way that fewer error messages will be sent to the chat. Also, I have just resolved the conflict with the recently merged PR changing doors to `default.can_interact_with_node`.

I'm going to be away until Sunday though, so if there's any issues with mine, it's probably best to just work off this one @sofar.

PR #1654, opened 2017-03-19
xpanes: support use_texture_alpha option

ForbiddenJ:
My [stained_glass](https://github.com/ForbiddenJ/stained_glass) mod uses a soup-bowl combination of textures to make it transparent while being dynamically updated to whatever texture. However, xpanes doesn't support use_texture_alpha. That's why I need this option supported.

screenshot from 2017-03-18 21-48-08
Before

screenshot from 2017-03-18 21-47-34
After

### DIFF
--- a/mods/xpanes/init.lua
+++ b/mods/xpanes/init.lua
@@ -94,6 +94,7 @@ function xpanes.register_pane(name, def)
 	minetest.register_node(":xpanes:" .. name .. "_flat", {
 		description = def.description,
 		drawtype = "nodebox",
+		use_texture_alpha = def.use_texture_alpha,
 		paramtype = "light",
 		is_ground_content = false,
 		sunlight_propagates = true,
@@ -120,6 +121,7 @@ function xpanes.register_pane(name, def)
 	groups.not_in_creative_inventory = 1
 	minetest.register_node(":xpanes:" .. name, {
 		drawtype = "nodebox",
+		use_texture_alpha = def.use_texture_alpha,
 		paramtype = "light",
 		is_ground_content = false,
 		sunlight_propagates = true,


sofar:
at issue is that transparency doesn't work. Try and place the panes you made near water and look at water from different angles through them.

ForbiddenJ:
@sofar I see what you're saying, though the issue on my computer is certain things not drawing behind other transparent objects. Like this:

screenshot from 2017-03-19 00-46-04

sofar:
Correct, and it's a known issue, and it's the reason that we don't allow alpha for any nodes in minetest_game. It needs to get fixed first.

sfan5:
Modders know what they are signing up for when they make use of `use_texture_alpha`, this is fine as long as we don't officially add this kind of broken nodes.
Just add a note to the docs warning about this.

paramat:
:-1: I agree with sofar, i don't feel it's a good idea to actively encourage modders to use a broken z-sorting.

ForbiddenJ:
Okay I guess. The ocean wouldn't look nice if there's no visible water through the window. I do like what juhdanad said over at https://github.com/minetest/minetest/issues/5083, however it means dividing the mapblocks and sorting the nodes of said mapblocks for whichever corner is closest every time the player moves.

Fixer-007:
We need z-sorting for sure, it is not just missing water, it is annoying epileptic blinking.

ForbiddenJ:
https://github.com/minetest/minetest/blob/master/src/content_mapblock.cpp#L550 seems to be relevant to these transparent nodes. Reason is, I reversed z at this line, and it affected the glitch.
`for(s16 z = MAP_BLOCKSIZE - 1; z >= 0; z--)`

screenshot 001
Same water tank made from stained glass. I'm facing x+ in these pictures.

Any way this can be used to fix part of the issue?

paramat:
That change just changes the order of rendering, the bug would likely then occur when seen in the other direction. Which translucent textures you see or not apparently depends on order of render which is why it is inconsistent. The issue is apparently a lot of work to fix, hmmmm went into detail on IRC years ago.

ForbiddenJ:
Yeah. That's all it does pretty much. But I still don't yet get is why does the water either show in it's entirety or not at all? In the last picture I posted, that glass tank was brim full of water.

sofar:
until something like https://github.com/minetest/minetest/issues/5083 gets merged, we should not consider merging alpha stuff like this.

PR #1658, opened 2017-03-22
PB&J Pup.

sofar:
Replace possibly trademarked artwork with royalty free and
OK for commercial use artwork.

The Pup appears by default but does not alias the Nyan Cat nodes
away. For that you will have to enable the setting. A settingtypes.txt
is provided to make that really easy. This allows people to get
an out-of-tree mod and not have their Nyan nodes disappear.

The mod can also be disabled entirely this way for those who don't
want it.

There's some eatser eggs in here as well. Instead of making the
nodes burnable, they are edible, and make you bark and howl. And
perhaps something else too.


sofar:
https://www.youtube.com/watch?v=NM_JaLxKBSk

sofar:
Fixes #1647

mods/pbj_pup/init.lua
```diff
@@ -0,0 +1,131 @@
...
+	},
+	paramtype = "light",
+	light_source = default.LIGHT_MAX,
+	paramtype2 = "facedir",
+	groups = {cracky = 2},
+	is_ground_content = false,
+	legacy_facedir_simple = true,
+	sounds = default.node_sound_defaults(),
+	stack_max = 1,
+	on_use = function(itemstack, user, pointed_thing)
```
sfan5: why not `on_punch` too?

mods/pbj_pup/init.lua
```diff
@@ -0,0 +1,131 @@
...
+		end
+	end
+
+	minetest.register_on_generated(generate)
+end
+--
+-- compat
+--
+
+if minetest.setting_getbool("pbj_pup_alias_nyan_cat_blocks") then
```
sfan5: i feel the name is way too long.

what about `pbj_pup_compat_aliases`?

sfan5:
Did you run the textures through `optipng`? (better to do this before)

overall sad to see nyancats go but :+1:

tenplus1:
keep the rainbow tail, players use it in builds.

ExeterDad:
I mostly like it, and feel children will really enjoy it. It certainly is cute.
But the tail looks nothing like rainbow. Existing builds will be awful looking.

Thomas--S:
Please don't change the rainbows. They will break existing buildings.

paramat:
If you want to keep your rainbows, use the settings to keep your nyan cats, and use the separate nyan cat mod.

This looks good, once approved we need a little time to warn people of the change, and add the nyan cat mod to minetest-mods or elsewhere.

mods/pbj_pup/init.lua
paramat: Descriptive is good but maybe 'pbj_pup_alias_nyancat' since nyancat nodes are combined words?

sofar:
For those worried: In the default settings, this mod won't replace nyan cat blocks. This will protect old builds and will allow you to install a nyan cat mod yourself without losing any builds.

Thomas--S:
@paramat It would be quite silly to keep the nyan cat and risk legal problems. I have no problem with the PB&J pup, but please keep the rainbows. I think it doesn't matter that much if there are candy block behind it or rainbow blocks. But the rainbow will keep backwards compatibility, and it also looks somewhat noble, whereas the candy looks cartoon-ish.

paramat:
See what i wrote above :]

mods/pbj_pup/init.lua
sofar: It's not needed - on_use is called before on_punch

mods/pbj_pup/init.lua
sofar: went with that - pbj_pup_alias_nyancat

sofar:
The mod is setup to encourage people to install the old nyan cat mod, and keep their rainbow blocks that way. I don't want to make a mod that just adds the same rainbow blocks.

mods/pbj_pup/init.lua
paramat: It's the player making noises from eating these, eating by punching doesn't make sense.

paramat:
Heh this is silly.
Only problem is they don't generate unless you go into advanced settings and switch generation off then on, .conf will be empty until then. Generation should be enabled when the setting is missing?
Otherwise +1.

paramat:
The howling seems to go on a little too long? (5 mins).

paramat:
If the tail passes over a mapblock border the candy animation doesn't sync, it would be possible to define their position dependent on tail length (1-15 nodes long) such that tails never pass a border.

tenplus1:
By default the change to PB&J Pup should keep the rainbows by default without having players mess around with mod additions or settings, please stop breaking compatibility.

VanessaE:
I should point out that gloopblocks has a rainbow block also.  It needs a new texture, but I'll fix that soon-ish.

sofar:
@sfan5 yes these are optipng'd

sofar:
> they don't generate unless you go into advanced settings and switch generation off then on

that's not supposed to happen, I'll check it.

sofar:
> By default the change to PB&J Pup should keep the rainbows by default without having players mess around with mod additions or settings, please stop breaking compatibility.

@tenplus1 I'm making it so that everyone can get what they want, and in order to do that some players are going to have to go into the settings. Any other solution will break something for someone. Going into the settings once to turn on an option if you see unknown nodes is a much better solution than people unable to get the nyan cat blocks back that they want.

BBmine:
Now I know why I called the nyan cats "Nyan puppies". And I didn't even know about this! I've been calling it that for a few months. XD

sofar:
did @tenplus1 really thumbs up his own post? lol

sofar:
@paramat changed it to _enable and fixed the generation code, tested, now works when settings are missing.

tenplus1:
@sofar - I'm more concerned that the rainbow will be replaced by default which is used as a building block by players and that it should not be touched since it has no issues like the nyancat...  Crafting rainbow into candies would be a good addition though.

Also, yes... I do thumb my own posts if I stand behind them.

paramat:
:+1: If you're happy with tails passing over mapblock borders (it breaks the movement of candies from node to node).

rubenwardy:
I don't like the textures or the name "PB&J pup", however I'm neutral overall

> Also, yes... I do thumb my own posts if I stand behind them.

That's weird :P

DS-Minetest:
I'm still not sure how I should think about this change. :confused: 

_I'm wondering why you all recognized that late that 11 (@tenplus1) adds :+1:s to his comments. Am I the only one who looks that drastically on reactions? But well, it's like someone would laugh about his own joke._

stujones11:
I am happy enough with PB&J pup as a replacement for Nyan cat, however, I still think that more effort could be made to not break existing builds, the rainbow block should somehow be retained even if not used for the 'puppies' themselves.

Fun fact: I've be playing the game for over 4 years now and still never found one of these in the 'wild'. Looks like I never will.

sofar:
Offtopic, but I look at reactions to gauge how others feel. Mostly because unpopular opinions are usually never expressed several times, but people tend to leave support for them with reactions. So, it usually gives a slightly more nuanced view.

While voting for yourself is allowed in many elections, this isn't a voting system. It's a discussion system.

stujones11:
Offtopic, I've done this a few times myself accidentally when using my tablet, if you happen to touch the emoticon you automatically increment it (very annoying)

stujones11:
Suggestion: Alias existing nyan cat nodes to `default:tart` (or something else) and alias existing rainbows to `default:rainbow` (or whatever) but remove the nyancat mod so that no more can generate.

Then add pjb_pup as a new mod in its own right with entirely different content ids (no aliases)
This way nyan-cats and puppies could even co-exist if so desired. I see no good reason why this should not be possible,

paramat:
They can already co-exist if someone installs the nyancat mod.
The nyancat nodes can therefore not be aliased to anything else as that breaks builds.

paramat:
Tested again by creating a world in master and then using this new mtgame version.
Nyan cats becaome unknown nodes alongside pbj pups, then when the nyancat mod is added as a separate mod the unknown nodes become nyan cats again.
So fine.

paramat:
9b45e7b3e92b12791c6d6bc88c28d42fc9bd0858

HybridDog:
̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅ ̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅ ̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅ ̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅ ̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅ ̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅ ̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅ ̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅ ̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅ ̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅ ̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅ ̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅̅

PR #1665, opened 2017-03-25
Adding Unicode arrows to player inventory

MrRar:
screenshot_20170325_122654


mods/creative/inventory.lua
```diff
@@ -96,9 +96,9 @@ function creative.register_tab(name, title, items)
 					list[current_player;main;0,5.85;8,3;8]
 					list[detached:creative_trash;main;4,3.3;1,1;]
 					listring[]
-					button[5.4,3.2;0.8,0.9;creative_prev;<]
-					button[7.25,3.2;0.8,0.9;creative_next;>]
-					button[2.1,3.4;0.8,0.5;creative_search;?]
++					button[5.4,3.2;0.8,0.9;creative_prev;◄]
+ +					button[7.25,3.2;0.8,0.9;creative_next;►]
+ +					button[2.1,3.4;0.8,0.5;creative_search;→]
```
sfan5: what is the `+` doing here?

SmallJoker:
thx_

Thanks custom font, for full Unicode support \*cough*.

mods/creative/inventory.lua
SmallJoker: Diff copy fail :man_facepalming:

MrRar: I don't know. I've removed them.

MrRar:
@SmallJoker are you testing other arrows or my arrows?

SmallJoker:
@MrRar Your arrows, what else. Apparently the fallback font is only used for other languages but not rarely used characters that aren't supported by some fonts.

DS-Minetest:
I don't like the arrow for searching.
Also there would be others like , ,  or . (Not to see on github.)
:mag: ←emoticon What about a loup? letter→ 🔍
```
UTF-8: 0xF0 0x9F 0x94 0x8D
UTF-16: 0xD83D 0xDD0D

C-Oktal terminiertes UTF-8: \360\237\224\215
Dezimale XML-Entität: &#128269;
```

kaeza:
I tried doing this in the past and it's a mess. Some people like to use custom fonts and you can't be sure the font will have those glyphs.

IMHO, it should be done with image buttons and plain white icons for better portability.

kaeza:
The [Material Icons Library](https://material.io/icons/) contains some simple and recognizable icons, but I'm not entirely sure if the license ([Apache 2.0](http://www.apache.org/licenses/LICENSE-2.0.txt)) is compatible.

  

Note: these examples are black to be able to see them here; you can get white versions from that page.

kaeza:
This is how it looks with icons and a bit of tweaking:

paramat:
:-1: For above reasons.

SmallJoker:
@DS-Minetest PNG compression tool's fault.

I agree with kaeza and paramat here: If changed, images should be used. :-1: from my side

paramat:
See #1666

paramat:
Closing due to font issue, also the 'search' icom makes no sense, no one will know what it means.

bigfoot547:
> , ,  or 

Lol! This is what I see with my font! *Replacement characters!*
screenshot from 2017-05-12 18 42 12

DS-Minetest:
@bigfoot547 I don't remember. But using unicode chars make no sense anyway.

PR #1666, opened 2017-03-25
Creative: Use icons for buttons.

kaeza:
Also adds missing tooltips for page buttons. I can remove this if it's not wanted.

Note: This comment shows the current state; comments below are for an older version.


sfan5:
IMO those icons are a little too big, i'd prefer if they looked more like in the unicode PR

Fixer-007:
Too big for me too, also, they look poorly anti-aliased.

paramat:
Yes, now seen in a screenshot, too big and weird anti-aliasing.

rubenwardy:
These icons should be added to https://github.com/minetest/minetest imo, and used in the main menu as well. Useful to share them

tobyplowy:
@kaeza +1 but only if the icones can be changed with texture packs

DS-Minetest:
@tobyplowy I think, they can, as well as every other texture in a mod.

kaeza:
New version:

paramat:
The arrows are fine, but the cross and search need to be pixelated in the same way, they still have weird edges that will probably be fixed by pixelation. They will be blocky but will look better,

paramat:
The dustbin is only 16x16, maybe lower resolution will help?

paramat:
I like it.

paramat:
Please can you convert the textures to indexed colour for smaller files.
Also, i was thinking that the cross may look better if you add a single (3x3) pixel to the tip of each of the 4 bars to square-off the end. Perhaps you could try this and see if you like it?

paramat:
Unfortunately new screenshots are showing ugly distortion on the icons, it's possible image icons will never look good. The dustbin is distorted too but is not so much of a problem as is dark grey.

dgasc48

rubenwardy:
What if does it look like if you double the resolution?

Ferk:
I think there are simply scaling issues. Did you actually made the pixel art image or just resized a bigger one?

I quickly drew a magnifier icon myself in piskel, it's 16px, you can use it.

http://www.piskelapp.com/p/agxzfnBpc2tlbC1hcHByEwsSBlBpc2tlbBiAgIC3sorbCAw/edit

paramat:
The textures in this PR are 48x48, the dustbin is 16x16, both are distotred, a higher resolution might help.

paramat:
:-1: due to distortion issues when the inventory is resized. I think the current font symbols look fine.

sfan5:
The distortions are not too bad IMO

paramat:
We could try increasing the resolution a bit, for example double it.

paramat:
I tried doubling the size, no improvement of distortion, i tried 16x16 and the distortion was worse.
I'm neutral on this now though, removed my -1.

paramat:
@kaeza the icons need converting to indexed colour.

sfan5:
That can be done trivially on merge: `optipng -o5 -strip all mods/creative/textures/*`

paramat:
@rubenwardy same opinion on this? We have one +1 from sfan5, SmallJoker is neutral http://irc.minetest.net/minetest-dev/2017-12-20#i_5173584 and i'm -1.

rubenwardy:
I think it would look better if the icons were higher res

paramat:
@sofar @Ekdohibs any comment? This has been hanging around too long.

Ekdohibs:
I'm neutral on whether to use hi-res or low-res icons, but having the icons indeed seems like a good change to me, so :+1:.

sfan5:
IMO this should be merged now, high-res icons can be added later.

sofar:
I kinda like these icons. If someone wants to fine-tune them later that's OK, so, feel free to merge.

👍

paramat:
Good, whoever merges remember to convert the textures to indexed colour on merge.

sfan5:
dfe74948e298f6f9014e075512532c5fdd195a5e 5b57815af164325c6e6cc5cd37c68dcc9add23e8

PR #1670, opened 2017-03-28
doors: add groups to the door craftitem

cx384:
This change makes it possible to add groups for the door craftitems.

Reason:
I am making a mod which should add new doors which aren’t in the creative inventory, but at the moment this is not possible, because the `doors.register(name, def)` function doesn’t allow to set the not_in_creative_inventory group of the craftitem to 1.


SmallJoker:
LGTM.

PR #1689, opened 2017-04-06
Replace aspen leaves texture with BlockMen's

LNJ2:
https://github.com/BlockMen/minetest_next/commit/5b029b3f28b7bd85a473b45951e504fcb1dc7202


C1ffisme:
I like the look of them as individual textures, but I'm not sure whether or not or how much they clash with other textures.

sofar:
Taking just a few is awkward. I'm not a fan of the textures (some of these are good, but they're a significant departure of the existing style) all together.

Maybe just make a texture pack first out of minetest_next?

LNJ2:
1. screenshot_20170407_144547
2. screenshot_20170407_144504
You can't tell me that 1. looks better than 2.

Ferk:
Are you sure real trees are that bright? To me the current one looks more realistic and less cartoony, with proper use of contrast. Usually the leaves color depends a lot on how the light is hitting each leaf and the contrast in the original leaves kinda reflects that effect.

Plus the Minetest Next trunk looks like it has a lot of random pixel noise. Being brighter also makes it blend too much with the color of dirt, imho.

LNJ2:
> Are you sure real trees are always that bright?

https://duckduckgo.com/?q=apple+trees&iar=images
I'd say yes, in bright sunlight in summer, yes.
I personally think games should better look good and make fun to play, realism comes second (it's also in Minetest so: you can even carry up to 8*4*99 m^3 gold, that are 61,085,376 kg, and that is without the crafting grid :D).
Ok, ok I also don't want to have Minetest looking as a comic (see the sphax texture pack), but some more colours would be nice! I don't have a problem with the pine, acacia or aspen leaves/needles, they look good (and are realistic). But the apple tree leaves texture just looks too dark! And jungles are also a bit brighter (at least at the top, where the sunlight shines).

> The Minetest Next trunk looks like it has a lot of random pixel noise.

There I have to agree with you. The sides really look a bit random.

My personal opinion hasn't changed, but if everyone is against this, the textures can't fit here. (-> this is why we have texture packs :smile: )

paramat:
The trunk textures are horrible :-1: 
They do the ridiculous thing of giving a cubic trunk circular rings.
The barks are less linear and are a noisy mess. Out current jungletree bark texture with its vines is a long-time classic texture.
The internal rings are simplistic darkened pixels instead of the rich detail we have curretnly.

The 2 changed leaf textures are not worse but no better.

paramat:
The aspen leaf might be ok as it makes the colours richer.
The jungletree leaf seems to be a messy and undefined, not sure it's an improvement.

C1ffisme:
> Out current jungletree bark texture with its vines is a long-time classic texture.

Considering that we're still nowhere near 1.0.0, we shouldn't be this nostalgic about a texture on a block.

Besides, [Minecraft isn't object to changing it's textures](http://minecraft.gamepedia.com/File:Cobblestone_textures.png).

paramat:
1.0.0 is irrelevant, MT wil never be 'finished' and in a way was also 'finished' years ago.
It's more a case of the current texture being excellent while the proposed one is bad.

LNJ2:
I've reverted everything unless the aspen leaves texture.
I can understand you, paramat. The contrast between grass and leaves is much better with the old ones, but I still think that the new ones look better and are more colourful.

paramat:
default_aspen_leaves128

^ Current

default_aspen_leaves_lnj2_128

^ PR

paramat:
screenshot_20170411_032445
   
screenshot_20170411_032506

^ Current

screenshot_20170411_032612

screenshot_20170411_032549

^ PR

@sofar what do you think? I think it's an improvement.

GreenXenith:
> I think it's an improvement.

I think it looks nicer as well.

paramat:
:+1: and approval from sofar http://irc.minetest.net/minetest-dev/2017-04-11#i_4868092

paramat:
fe021281e62504674f57a196e6dfba83bdd52847

PR #1699, opened 2017-04-12
nodebox -> node_box

Thomas--S:
Seems like this was a typo


SmallJoker:
Since this is an invisible node that's required to make the player collide with it, the while `node_box` definition could be omitted.

C1ffisme:
Wait, do both of these work? I would think that the previous code was broken.

sofar:
Nah, this is cosmetic since the texture is blank, and the collision box is what this is about. This is a good fix though :)

Thomas--S:
Thanks for merging!

PR #1720, opened 2017-05-02
Making details more visible in logo

MrRar:
I added an overlay gradient and raised the contrast of the logo.

minetest_game
corresponding PR: https://github.com/minetest/minetest/pull/5695


paramat:
I've been looking at the current header and it could be a little more 'shiny' and defined, i wanted it to be like smooth-polished stone. A subtle gradient might help with that too.
I'm not sure about how this increases contrast for the whole texture, better that just the bevels have more contrast but the stone texture itself keeps the same contrast.
Ideally i should redo the header and make the bevels lighter and darker, then after that add a very subtle gradient.

I'll try this one though.

ShadowNinja:
I prefer the current image, this looks a bit washed out.

MrRar:
@paramat  I think the only way to express that the logo is polished stone is to put a smooth reflection on it.
Here is a demo:
header_polish
screenshot:
polished

DS-Minetest:
@MrRar That doesn't look like stone anymore. The current header looks like stone block.

SmallJoker:
The current texture looks good to me, no need to change anything in it.
:-1: from my side.

paramat:
3 -1s so closing.

PR #1732, opened 2017-05-08
redo new open chests to use all single textures

VanessaE:
Keep the filenames and their usage the same as they used to be (as a texture author or modder referencing the textures would normally see it, anyway), with two minor differences:
* The image on the right side of the chest is deliberately X-mirrored from the left side instead of being identical.
* A separate two-tile image for top- and bottom-inside of the "open" chest was added, taken from @sofar's all-in-one image.  By extension, the order of the textures in the tiles={} field for the open chest is not an exact match to the closed chest (can't be, and doesn't need to be anyway).


VanessaE:
Also, one could delete the back and bottom images and change four lines of code to use the "side" and "top" image in their places, if so desired (making the texturing essentially identical to the original chest).

sofar:
I won't throw my approval or disapproval on it, but I will repeat that I'm a fan of making models with full UV mapping that don't do "tricks|hacks|workarounds|clever things" like this proposal. At one point someone is going to want a chest with gold coins and show gold coins only on the bottom surface of the open chest, and not the top. At some point someone is going to want the left and right side of the chest to not be mirrored, etc.

Any change like this seems to be clinging to old customs for the sake of "backwards compatibility" or "too lazy to redo texture packs". Neither of those are arguments that make me feel happy and comfortable.

Note that for those too lazy to redo their texture packs (which takes 2 minutes in gimp at most) they don't even need to - the new chests will work just fine without doing so and not show up broken in the game, and this was done on purpose.

VanessaE:
The way chest textures exist now (that is, with the fully-UV-mapped image that this PR seeks to undo), every texture pack in existence would require an update.  Pipeworks and technic mods are affected also (since they redefine chests).

Making the open chest use separate tiles for the top- and bottom-inside is trivial, by the way -- done.

VanessaE:
...and to be honest, I don't think it's particularly likely for someone to find a good reason for the left and right sides to be anything but mirror images of one another.

tenplus1:
@VanessaE - silly question but is there any reason why the closed chest shouldn't be a normal nodebox ?

VanessaE:
Off the top of my head, there's no real reason for it to actually be a mesh, but it would add a few more lines of code to do it.  I just didn't feel like messing with that.

VanessaE:
That seems to work.  For some reason, closing an open, locked chest induces a brief "flicker", not sure why or if it's my change that did it.

SmallJoker:
Code is good but the texture documentation needs to be updated. Two removed textures and one new.

paramat:
> The way chest textures exist now (that is, with the fully-UV-mapped image that this PR seeks to undo), every texture pack in existence would require an update.

So with this PR texture packs would provide 'top' 'side' 'front' 'lock' and default would provide the new 'inside' textures? The inside textures have a 1px edge which would not match, so texture packs need to update anyway.
Your PR does make updating texture packs easier because only 'inside' and 'bottom' textures need adding and these are 16x16 as standard, opening and composing layers in the new large format is less easy for people.

Also note we now have the old chest textures present in MTG.

Any other advantages? I'm feeling neutral currently.

paramat:
> Note that for those too lazy to redo their texture packs (which takes 2 minutes in gimp at most) they don't even need to - the new chests will work just fine without doing so and not show up broken in the game,

Good point, the missing texture gets supplied by default.
So i'm currently :-1:

VanessaE:
Actually with the last commit I made, only the inside texture would need added (I made the back and bottom the same as the left and top, same as the old chests), and probably most texture packs can ignore it since it's just black with a brown edge.  Even HDX doesn't look bad next to it.

DS-Minetest:
It's not about texturepacks IMO,
- the crack is very ugly and
- the closed chest needs an extra cubic mesh.

VanessaE:
The closed chest is a normal node, in this PR.  Why does it need to be a mesh?  And if cracks are ugly, fix them.  They're ugly on lots of non-cubic nodes.

DS-Minetest:
I mean without the PR. With this PR it's ok.

DS-Minetest:
:+1: Tested. Works fine. And diffs look good.

paramat:
Removed my -1, i'm fairly neutral.

Ekdohibs:
So as I said earlier on IRC, I feel that backwards compatibility is important, but the extra flexibility that was added is good as well, so +0.5

tenplus1:
Backwards compatibility was the whole reason this pull was made, the current .obj chests break too many mods and texture packs which would be a lot more work to fix, so pushing this would return things to a working state and still have animated chests.

paramat:
Texture packs wil have to update anyway to add a suitable 'inside' texture, but it's easier to update.

DS-Minetest:
>  a suitable 'inside' texture

I don't think, the inside texture is so important to be made to suit to the rest.
>The mesh chest remains for open chest, so update is still needed there too.

Do you mean the problem that could be now for mods, that there's now also an open chest?
If yes, it doesn't fit to this PR. => Probably the answer would be no. So, what do you mean? >_<

Don't forget, this PR is not only about backwards compatibility. Just think about the ugly crack.

Anyway, what would speak against this to be merged?

<details><summary>Offtopic</summary>

> Wow someone thumbed-up your comment.

Now guess, who this someone is:
2017-05-16-150720_1280x1024_scrot
</details>

paramat:
The inside texture has an edge that will need colour-matching to a texture pack textures.

What do you mean by 'ugly crack'? on current or PR version? Do you mean 'digging crack animation'? That is ugly on many nodes anyway.

I mean, some are claiming this PR will solve all mods updating issues, it will not:

>  the current .obj chests break too many mods 

The open chest is still an .obj.
Mods will need updating whether this PR goes in or not.

tenplus1:
@paramat - it's not the .obj file that's breaking the mods but the current texture file in use that doesnt allow pipeworks to overlay it's textures properly, as well as barter shops, protected chests, more_chests etc. etc.  The above .obj model for open chests uses the previous textures and works just fine and the closed chest is a simple nodebox which doesnt have the ugly crack when dug... Also, opening a chest and having the default inlay is fine on most texture packs and can be easily replaced without much work.

paramat:
Ok i see, yes it will help with mod texture issues.

paramat:
:+1:

numberZero:
Er, texture packs that aren’t being updated are mostly unusable anyway, there are so many new nodes that usually don’t fit the modified look (esp. with non-16px TPs)

> Pipeworks and technic mods are affected also (since they redefine chests).

That’s the only problem I see. But there should be a better way. Also, `technic` should be updated to use the models as well (that may need some API changes, though).

Note that any mod tampering with chests needs changing anyway (at least because there are 2 chest nodes now), your PR won’t help compatibility much. And, well, if you don’t want to alter the textures in a mod, you can just combine them on the fly; it is easy to detect is that necessary checking the presence of the open chest node.

By the way, the less textures are used, the faster the rendering is.

So, I’m against this PR. MT needs to move forward, keeping the good and discarding the bad.

P.S. @VanessaE I personally have the opposite problem with your TP: it doesn’t support old MTG version installed on `beha.zapto.org`. (Actually it doesn’t support `ethereal` either, but that’s another story)

paramat:
d104b9b10ffc2ee2dd16bd2f2a6aaa5296913ebc

Ferk:
@DS-Minetest I don't understand the complaint about the "ugly crack". Could you explain?

I mean.. if anything, this PR would actually *cause* ugly cracks, not fix them. Old textures that were not designed to match the open chest crack will be forced into the new mesh, possibly looking weird if the old texture didn't have the crack at exactly the same position.

DS-Minetest:
Before this PR was merged, the crack was like this:
screenshot_20170518_180120

VanessaE:
I think what happens with the cracks when applied to a mesh node, is that the crack image gets scaled up to the size of the image(s) used on the model, then applied over it, rather than being scaled to fit the sides/faces/whatever.  For a node like the previous mesh-based closed chest, I guess each frame gets scaled to 64x32 to match the UV-mapped texture image, and then blitted right over it.

PR #1736, opened 2017-05-16
Add inner and outer stair corners

MarkuBu:
This PR adds simple stairs for corners. It uses nodeboxes. 

If someone can create mesh nodes I will add them

### DIFF
--- a/mods/stairs/init.lua
+++ b/mods/stairs/init.lua
@@ -311,7 +311,7 @@ function stairs.register_stair_inner(subname, recipeitem, groups, images, descri
 			minetest.register_craft({
 				type = "fuel",
 				recipe = 'stairs:stair_inner_' .. subname,
-				burntime = math.floor(baseburntime * 0.75),
+				burntime = math.floor(baseburntime * 0.625),
 			})
 		end
 	end
@@ -372,7 +372,7 @@ function stairs.register_stair_outer(subname, recipeitem, groups, images, descri
 			minetest.register_craft({
 				type = "fuel",
 				recipe = 'stairs:stair_outer_' .. subname,
-				burntime = math.floor(baseburntime * 0.75),
+				burntime = math.floor(baseburntime * 0.875),
 			})
 		end
 	end
--- a/mods/stairs/init.lua
+++ b/mods/stairs/init.lua
@@ -274,6 +274,13 @@ function stairs.register_stair_inner(subname, recipeitem, groups, images, descri
 				{-0.5, 0, -0.5, 0, 0.5, 0},
 			},
 		},
+		on_place = function(itemstack, placer, pointed_thing)
+			if pointed_thing.type ~= "node" then
+				return itemstack
+			end
+
+			return rotate_and_place(itemstack, placer, pointed_thing)
+		end,
 	})
 
 		if recipeitem then
@@ -328,6 +335,13 @@ function stairs.register_stair_outer(subname, recipeitem, groups, images, descri
 				{-0.5, 0, 0, 0, 0.5, 0.5},
 			},
 		},
+		on_place = function(itemstack, placer, pointed_thing)
+			if pointed_thing.type ~= "node" then
+				return itemstack
+			end
+
+			return rotate_and_place(itemstack, placer, pointed_thing)
+		end,
 	})
 
 	if recipeitem then
--- a/mods/stairs/init.lua
+++ b/mods/stairs/init.lua
@@ -275,6 +275,39 @@ function stairs.register_stair_inner(subname, recipeitem, groups, images, descri
 			},
 		},
 	})
+
+		if recipeitem then
+			minetest.register_craft({
+				output = 'stairs:stair_inner_' .. subname .. ' 6',
+				recipe = {
+					{recipeitem, recipeitem, ""},
+					{recipeitem, recipeitem, recipeitem},
+				},
+			})
+
+			-- Flipped recipe for the silly minecrafters
+			minetest.register_craft({
+				output = 'stairs:stair_inner_' .. subname .. ' 6',
+				recipe = {
+					{"", recipeitem, recipeitem},
+					{recipeitem, recipeitem, recipeitem},
+				},
+			})
+
+		-- Fuel
+		local baseburntime = minetest.get_craft_result({
+			method = "fuel",
+			width = 1,
+			items = {recipeitem}
+		}).time
+		if baseburntime > 0 then
+			minetest.register_craft({
+				type = "fuel",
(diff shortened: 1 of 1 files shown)


MarkuBu:
https://github.com/minetest/minetest_game/issues/1443

paramat:
These will have a lot of hidden faces so meshnodes are needed for these. Good feature though.

MarkuBu:
I know that nodeboxes are suboptimal. Will change that as soon as I have mesh nodes. But I can't make them myself because I have absolutely no Blender skills.

Recipes are added, but not uploaded

But I don't know what you mean with _the cut-away part is diagonally facing the player_

paramat:
68747470733a2f2f7331372e706f7374696d672e6f72672f67386f68736f6c6e6a2f636f726e65725f7374616972732e706e67

See the orientations on the right, that is how the node should be placed relative to the player facing diagonally.

paramat:
Then maybe also use the existing 'pointed thing to face pos' function to turn the corner stair upside-down depending on what part of a wall's tile is pointed at.

paramat:
The meshnodes can be swapped in much later when they're done, maybe even in another PR, nodeboxes are fine for now while this is developed.

MarkuBu:
I add rotate and place. I think it works intuitive.

mods/stairs/init.lua
```diff
@@ -255,13 +255,137 @@ if replace then
...
+		local baseburntime = minetest.get_craft_result({
+			method = "fuel",
+			width = 1,
+			items = {recipeitem}
+		}).time
+		if baseburntime > 0 then
+			minetest.register_craft({
+				type = "fuel",
+				recipe = 'stairs:stair_inner_' .. subname,
+				burntime = math.floor(baseburntime * 0.75),
```
paramat: 5 small cubes out of 8:
`baseburntime * 0.625`

mods/stairs/init.lua
```diff
@@ -255,13 +255,137 @@ if replace then
...
+		local baseburntime = minetest.get_craft_result({
+			method = "fuel",
+			width = 1,
+			items = {recipeitem}
+		}).time
+		if baseburntime > 0 then
+			minetest.register_craft({
+				type = "fuel",
+				recipe = 'stairs:stair_outer_' .. subname,
+				burntime = math.floor(baseburntime * 0.75),
```
paramat: 7 small cubes out of 8:
`baseburntime * 0.875`

paramat:
I'll test sometime.

mods/stairs/init.lua
MarkuBu: Ah, now I understand that. Will change tomorrow

mods/stairs/init.lua
```diff
@@ -255,13 +255,137 @@ if replace then
...
+		on_place = function(itemstack, placer, pointed_thing)
+			if pointed_thing.type ~= "node" then
+				return itemstack
+			end
+
+			return rotate_and_place(itemstack, placer, pointed_thing)
+		end,
+	})
+
+		if recipeitem then
```
SmallJoker: Please repair these indents. The following lines should have the same level as `local baseburntime`.

MarkuBu: you're right. No idea why there was an indention. Fixed in the next push

GreenXenith:
Is this PR still needing meshes? If the meshes are done before June 4th will this be merged for 0.4.16 release?

GreenXenith:
@MarkuBu here's some meshes:
[corner_stairs.zip](https://github.com/minetest/minetest_game/files/1033167/corner_stairs.zip)

TumeniNodes:
he was looking for .b3d, not sure why it matters. .obj should work fine but if you are able to export .b3d to add, I think he would be happy

sorry to chime in,

GreenXenith:
Don't know why he would use .b3d if the normal stair is .obj...

TumeniNodes:
no idea. .obj is pretty much the standard for MT. .b3d offers better animation but def not needed in this instance. I think at this point he will be happy with these. esp if it will mean being merged now : )

MarkuBu:
I'm sorry. Cause I work with mobs I had b3d in mind. obj is fine too.

But I don't have Internet on my PC I can't make any changes right now

GreenXenith:
Good thing you can't do it right now. The models need to be fixed anyway :/
Fixing now...

paramat:
Should be .obj like the stair.
This will not be in release because 'feature freeze', bugfixes only.

GreenXenith:
Problem. The wooden outer stair has the wrong wood grain on the top part on the right side, but if I fix that, the bronze outer stair (or any metal block for that matter) will have the wrong outline orientation.
screenshot_20170527_112225

MarkuBu:
I'm temporally back in business, but if this won't be add to 0.4.16 there is no hurry

paramat:
It should be possible to fix both, normally a texture is projected onto the side so the top half of the wood stair should not be vertical anyway.

GreenXenith:
@MarkuBu here are working meshes. Note, you will have to define new hit/selection boxes.
[corner_stairs.zip](https://github.com/minetest/minetest_game/files/1033765/corner_stairs.zip)

TumeniNodes:
If it could be considered that stairs are missing the outer / inner corners, is indeed a "bug", this becomes a bugfix.
(could be a sleazy, but workable way of working this into release)? Knowing this addition will not add any actual bugs. And would be a very welcome addition to end users.

MarkuBu:
@GreenXenith thanks, I try it tomorrow. They have to behave like the nodebox version without changing the code to rotate them

paramat:
About nodeboxes that get converted to meshes, they get generated with the internal faces of all boxes, which is why .obj meshes are better, no internals, faster render.
This is why slabs are not .obj, no internal faces so no advantage to using a .obj.

paramat:
MarkuBu, these will need a dedicated placement function because these are diagonally-orientated nodes, so their rotation should switch at player yaw 0, 90, 280, 270 degrees instead of 45, 135, 225, 315 degrees. The function can use the same 'pointed thing to face pos' code for deciding upright or upside-down.

paramat:
Also .obj meshes are needed.

paramat:
@MarkuBu are you able to finish this? A new PR has appeared #1814 but it has no meshnodes.

MarkuBu:
@paramat as I mentioned in the forum I reduce my Minetest activity because I want to start my own project  and I don't have time for both. 

https://forum.minetest.net/viewtopic.php?f=3&t=17879

paramat:
Ok. You've been a good contributor and will be missed.

New PR #1820

PR #1764, opened 2017-06-07
Adding backface culling to stairs

MrRar:
I used this Node.js script:

var fs = require("fs")
var file = fs.readFileSync("init.lua").toString()
var newFile = file
var matches = []
var cursor = 0;
while(true) {
	var pngPtr = file.indexOf(".png", cursor)
	if(pngPtr == -1) break
	matches.push(pngPtr)
	cursor = pngPtr + 4
}
for(var i in matches) {
	var end = matches[i]
	var start = file.lastIndexOf("\"", end)
	++start
	//console.log(start + ":" + end)
	var imgName = file.slice(start, end)
	var oldTexture = '"' + imgName + '.png"'
	var newTexture = "{name = \"" + imgName + ".png\", backface_culling = true}"
	newFile = newFile.replace(oldTexture, newTexture)
}
fs.writeFileSync("init.lua",newFile)

### DIFF
--- a/mods/stairs/init.lua
+++ b/mods/stairs/init.lua
@@ -272,7 +272,7 @@ stairs.register_stair_and_slab(
 	"wood",
 	"default:wood",
 	{choppy = 2, oddly_breakable_by_hand = 2, flammable = 2},
-	{"default_wood.png"},
+	{{name = "default_wood.png", backface_culling = true}},
 	"Wooden Stair",
 	"Wooden Slab",
 	default.node_sound_wood_defaults()
@@ -282,7 +282,7 @@ stairs.register_stair_and_slab(
 	"junglewood",
 	"default:junglewood",
 	{choppy = 2, oddly_breakable_by_hand = 2, flammable = 2},
-	{"default_junglewood.png"},
+	{{name = "default_junglewood.png", backface_culling = true}},
 	"Jungle Wood Stair",
 	"Jungle Wood Slab",
 	default.node_sound_wood_defaults()
@@ -292,7 +292,7 @@ stairs.register_stair_and_slab(
 	"pine_wood",
 	"default:pine_wood",
 	{choppy = 3, oddly_breakable_by_hand = 2, flammable = 3},
-	{"default_pine_wood.png"},
+	{{name = "default_pine_wood.png", backface_culling = true}},
 	"Pine Wood Stair",
 	"Pine Wood Slab",
 	default.node_sound_wood_defaults()
@@ -302,7 +302,7 @@ stairs.register_stair_and_slab(
 	"acacia_wood",
 	"default:acacia_wood",
 	{choppy = 2, oddly_breakable_by_hand = 2, flammable = 2},
-	{"default_acacia_wood.png"},
+	{{name = "default_acacia_wood.png", backface_culling = true}},
 	"Acacia Wood Stair",
 	"Acacia Wood Slab",
 	default.node_sound_wood_defaults()
@@ -312,7 +312,7 @@ stairs.register_stair_and_slab(
 	"aspen_wood",
 	"default:aspen_wood",
 	{choppy = 3, oddly_breakable_by_hand = 2, flammable = 3},
-	{"default_aspen_wood.png"},
+	{{name = "default_aspen_wood.png", backface_culling = true}},
 	"Aspen Wood Stair",
 	"Aspen Wood Slab",
 	default.node_sound_wood_defaults()
@@ -322,7 +322,7 @@ stairs.register_stair_and_slab(
 	"stone",
 	"default:stone",
 	{cracky = 3},
-	{"default_stone.png"},
+	{{name = "default_stone.png", backface_culling = true}},
 	"Stone Stair",
 	"Stone Slab",
 	default.node_sound_stone_defaults()
@@ -332,7 +332,7 @@ stairs.register_stair_and_slab(
 	"cobble",
 	"default:cobble",
 	{cracky = 3},
-	{"default_cobble.png"},
+	{{name = "default_cobble.png", backface_culling = true}},
 	"Cobblestone Stair",
 	"Cobblestone Slab",
 	default.node_sound_stone_defaults()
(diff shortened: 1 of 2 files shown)


tenplus1:
What advantages does it give?

paramat:
Possibly faster render since it does not render the faces facing away from the camera?

paramat:
screenshot_20170608_115153

^ Inside a stair node.

Works.
Interestingly a slab already had backface culling somehow, maybe because it is a nodeboxes converted to a meshnode, instead of an .obj meshnodes like the stair.

paramat:
:+1:

sfan5:
couldn't `stairs.register_stair_and_slab` just set `backface_culling` in the nodedef before registering?

paramat:
Good point, once at line 54.

MrRar:
@sfan5 That assumes that no stair will require backface rendering. Is that a good assumption to make? 
Should another parameter be added to register_stair_and_slab for whether to have backface culling for the stair?

paramat:
> Is that a good assumption to make?

Erm, some mods may be registering glass stairs, do they need backfaces rendered?

paramat:
> That assumes that no stair will require backface rendering. Is that a good assumption to make?

This is fine, because slabs (that arendeboxes converted to meshnodes) automatically do not have backface culling (and never have had), so stairs just become consistent with slabs. Glass slabs and stairs do not need backface rendering.

paramat:
Thanks will test.

paramat:
How about
```
	for index, image in pairs(images) do
		images[index] = {
			name = image,
			backface_culling = true,
		}
	end
 	groups.stair = 1
 		description = description,
 		drawtype = "mesh",
 		mesh = "stairs_stair.obj",
		tiles = images,
```

paramat:
DS-minetest also warns that 'images' can now only contain a .png file, no other fields.

DS-Minetest:
Actually I would rather suggest something like:
```
for i = 1, #images do
	if type(images[i]) == "string" then
		images[i] = {
			name = images[i],
			backface_culling = true,
		}
	else
		images[i].backface_culling = images[i].backface_culling == nil and true or images[i].backface_culling
	end
end
```
But I'm not sure.

sfan5:
That code will modify the passed table, use something like this:
```lua
	if type(image) == "string" then
		images[i] = {
			name = image,
			backface_culling = true,
		}
	else
		if image.backface_culling == nil then -- override using any other value
			images[i] = table.copy(image)
			images[i].backface_culling = true
		end
	end
end
```

MrRar:
@sfan5 It does not work because you're still modifying the original images table.

paramat:
Latest version seems potentially good.

sfan5:
oh yeah, well just throw a `images = table.copy(images)` in there

mods/stairs/init.lua
```diff
@@ -46,12 +46,24 @@ end
...
 function stairs.register_stair(subname, recipeitem, groups, images, description, sounds)
+	local stair_images = {}
+	for i, image in ipairs(images) do
+		if type(image) == "string" then
+			stair_images[i] = {
+				name = image,
+				backface_culling = true,
+			}
+		elseif image.backface_culling == nil then
+			stair_images[i] = table.copy(image)
```
paramat: `stair_images = table.copy(images)`?

paramat: Is this correct? or
`stair_images = table.copy(images)`?
or
`stair_images[i] = images[i]`?

mods/stairs/init.lua
```diff
@@ -46,12 +46,24 @@ end
...
 
 function stairs.register_stair(subname, recipeitem, groups, images, description, sounds)
+	local stair_images = {}
+	for i, image in ipairs(images) do
+		if type(image) == "string" then
+			stair_images[i] = {
+				name = image,
+				backface_culling = true,
+			}
+		elseif image.backface_culling == nil then
```
sfan5: you should keep the comment about overriding here

mods/stairs/init.lua
sfan5: Neither of those are correct, it already does what it's supposed to.

paramat: Ok, but your comment suggests there is an error.
 > oh yeah, well just throw a `images = table.copy(images)` in there

sfan5: Yes, the original images table should either be copied or (like here) constructed from scratch.

paramat: Ok i see now, seems good as it is.

paramat:
Code seems good now.

paramat:
I'm now thinking it's better to just add 'backface_culling = true' to the tiles as the PR originally did, because that teaches modders to add that by being visible in the registrations. It's better that modders are aware of the advantage instead of hding it and doing it automatically for them.

sfan5:
Adding a note to the docs should suffice IMO.

paramat:
+1 if the author has tested.

mods/stairs/init.lua
```diff
@@ -46,12 +46,24 @@ end
...
 
 function stairs.register_stair(subname, recipeitem, groups, images, description, sounds)
+	local stair_images = {}
+	for i, image in ipairs(images) do
+		if type(image) == "string" then
+			stair_images[i] = {
+				name = image,
+				backface_culling = true,
+			}
+		elseif image.backface_culling == nil then -- override using any other value
```
paramat: `elseif image.backface_culling != true then`
To override someone setting it to 'false'?

sfan5: (in that case the if condition would be useless anyway)

why?
maybe a modder wants to set it to false for some reason, that should be supported

paramat: Ok, i don't mind, but the comment suggests otherwise:
` -- override using any other value`
FIne as is then.

sfan5: the comment is a little ambiguous, it's supposed to mean that "the code that sets it to true can be overriden (disabled) by using (setting) any other value (than `nil`) in the definition table"

paramat: I see, fine.

paramat:
:+1:

DS-Minetest:
This needs documentation.

paramat:
I don't think docs in game_api.txt are needed.

sfan5:
Tested and works. :+1:

paramat:
Applying: Adding backface culling to stairs
error: patch failed: mods/stairs/init.lua:342
error: mods/stairs/init.lua: patch does not apply

sfan5:
You need to use `git am -3`

paramat:
Ok 3 way merge.

paramat:
7b46d3b6afa7c14a6ed26368820334b1adfc5e4d

PR #1767, opened 2017-06-08
Meta-set nodedef

Thomas--S:
To be merged together with minetest/minetest#5919


Thomas--S:
> I won't be able to finish this before June 2018. I'm also thinking of a different approach.

PR #1794, opened 2017-06-23
[WIP] make beds colored

DS-Minetest:
This shall solve the problem of finding red dye.
Todo:
- [x] also make "fancy" bed colored
- [x] make textures good
- [x] choose the right colors (only 8 are allowed)
- [ ] make colored bed item good
- [x] document
- ~~[ ] use `"color"` `"..wallmounted"` instead of `"..facedir"` for more colors?~~
<details><summary>screenshots</summary>

How it could be:
screenshot_20170625_202926
How it is:
screenshot_20170626_152820
</details>


paramat:
I guess this will continue to support any beds that mods register (including the 2 registered in the mod itself) but add 8 that are hardware coloured.
Is there any point in adding 8 coloured fancy beds, maybe just the simple ones are enough?

Ezhh:
I'd much prefer to see coloured fancy beds than coloured simple beds. If you are going for anything decorative and care about appearance, you are much more likely to want fancy beds after all.

mods/beds/beds.lua
```diff
@@ -58,28 +58,47 @@ beds.register_bed("beds:bed", {
...
+	},
+	palette = "beds_palette.png",
 	nodebox = {
 		bottom = {-0.5, -0.5, -0.5, 0.5, 0.06, 0.5},
 		top = {-0.5, -0.5, -0.5, 0.5, 0.06, 0.5},
 	},
 	selectionbox = {-0.5, -0.5, -0.5, 0.5, 0.06, 1.5},
 	recipe = {
-		{"wool:red", "wool:red", "wool:white"},
+		{"group:wool", "group:wool", "wool:white"},
```
paramat: Remember the white bed should not need any dye.

paramat:
Yes i assume they will be popular, so that was probably a silly question. I don't object to coloured fancy beds.

DS-Minetest:
Just to make it clear, there aren't 8 beds added but the existing nodes get "colorfacedir" as paramtype2, on craft the bed item is colored and on place the node is colored with param2 depending on the "palette_index" item meta field. But not all registered beds will be changed automatically, only if they have a palette.

cx384:
This is similar to https://github.com/minetest/minetest_game/pull/1566 ?

DS-Minetest:
No.

DS-Minetest:
For everything else on my todo list I think I need help in form of information.

paramat:
I recommend staying with facedir for simplicity, 8 colours is enough.

White (of course, no dye needed)
Red, Orange, Yellow, Green, Blue, Violet (too match flowers and basic dyes).
Leaves 1 spare, ideas?

DS-Minetest:
colors
Ok, I currently have dark red (on first place and dark for backwards compatibility), blue, green, yellow, magenta, cyan, orange and a bright black. I'll take instead of magenta and cyan white and violet.

cx384:
I think a dark green bed looks better than a light (normal) green bed.

DS-Minetest:
Ok, white bed look bad, magenta is better than violet imo.
I will look for darker green.

DS-Minetest:
<details><summary>some screenshots</summary>

Current:
screenshot_20170625_103559
Paramat's suggestion:
screenshot_20170625_103535
Some colors a bit darker:
screenshot_20170625_103457
</details>

Ezhh:
I prefer paramat's  suggestion,, but you are right about the white bed not looking good, so maybe switch that one to cyan or magenta, unless someone can think of a way to make white look better?

DS-Minetest:
I think in 3 the grayish white looks ok.
I would like to take 3 but instead of violet magenta.

Ezhh:
I really dislike how dark the orange is in option 3. The darker colours (apart from the green) look really drab and uninteresting to me. I also much prefer to have violet and match the flowers as paramat said.

Grey looks okay, but it just seems a pity to lose a colour for it. Losing a colour for white would be fine if it looked nice, but I think losing one for grey isn't worth it.

DS-Minetest:
@Ezhh Don't forget, since only 8 colors are easily possible, bedcolors have more than one possible crafting recipe. Eg. gray bed can be crafted with white or grey wool, green bed with green or dark green dye, orange bed with orange or brown dye…

Ezhh:
I didn't comment on the crafting, only on which colours I like... If you are using the orange in the option 3 screenshot, it is really, really ugly in my opinion, but the choice isn't mine, so I'll leave it to you and the devs.

DS-Minetest:
The colors can be changed with a texturepack, so brightness shouldn't be a problem.
I'll take magenta instead of violet because magenta has the same values of blue and red, violet is to bluish.

The real problem is with the item. At the moment it looks very bad.

paramat:
Please can we keep violet? it does not look like blue at all. Then we will have colours matching the 8 basic dyes.
Black, white, red, orange, yellow, green , blue, violet.

Note a white or creamy-natural wool colour is essential because the point of changing beds is that we have a white bed that does not need dyes to craft. How about a cream-white instead of a grey-white (which is ugly i agree).

paramat:
The thing about pink being essential for girls is a bit of a stereotype. In MTG the basic colours are the basic colours of the spectrum (rainbow) so beds should follow that, pink items can and are added by mods.

DS-Minetest:
Personally I don't like the creamy white, but there are texturepacks.
<details><summary>screenshot</summary>

screenshot_20170626_152820
</details>

Actually that girls like pink isn't just stereotype, I've once seen something in TV where they said that very small kids rather like blueish colors if it is a boy or pinkish (or redish) colors if it is a girl.

paramat:
Hmm maybe the normal white would look better, i'm fine with the bright white one.
Thanks.

mods/beds/beds.lua
```diff
@@ -102,3 +140,28 @@ minetest.register_craft({
 	recipe = "beds:bed_bottom",
 	burntime = 12,
 })
+
+-- colored Crafting
+
+minetest.register_on_craft(function(itemstack, player, old_craft_grid, craft_inv)
```
juhdanad: You don't need a server-side function to do the craft colorization if you only support the default dyes: you can register a craft recipe for each dye then. In lua_api.txt you can find an example how to make colored craft recipes in the Hardware coloring section.

DS-Minetest:
Now the crafting is better. It also gets now a description that includes the color. The item color can be made better when minetest/minetest#6064 is addressed.

DS-Minetest:
Hm, maybe I should not include the color into the description. Or is there an easy way to change the description of the item when the player gets it?

paramat:
Sorry to suggest this now, i think we should consider using conventional texture colourisation instead of hardware colourisation.
I think #1566 does this for beds.
Hardware colourisation is more complex and for just a few beds there is no advantage to using it.
Normal texture colourisation allows more colours than 8, we could then have a bed for each wool colour.

octacian:
We can't have a bed for each wool colour as it is? If we can't, then I for one would prefer conventional texture colourization. Otherwise, I for one think that the advantages to hardware colorization and the way it is typically implemented make it much nicer.

paramat:
We're limited to 8 colours because the bed is 'colorfacedir', it has to store 24 rotations plus colour in param2.
There are no advantages to hardware colourisation if you only have a few nodes, it comes in useful for very large amounts of nodes only (we will likely only have as many beds as wool colours). However there are disadvantages due to complexity and issues it causes, the feature is not finished yet, for example any beds dropped as items don't have the suitable colourisation.
We devs will have to maintain this so i'd like to avoid hardware colourisation unless there is an advantage to it.

paramat:
Sorry but :-1: 
For a few nodes it's better to use normal texture colourisation and avoid the extra complexity of hardware colourisation.
I would like a bed for each wool colour.

DS-Minetest:
>a few nodes

Actually if you would add more nodes instead, it won't only be a few nodes. There are 2 nodes per bed and 2 beds and there are 16 colours => 2*2*16 = 64 => many nodes imo.

paramat:
I mean counting a bed as one item, 32 items is still not a lot of items to register.

Also for this task, the way it interacts with the existing registering of beds needs careful consideration. I'd like @sofar 's input if possible.

paramat:
Perhaps i'll need to add a simple white bed after all.
Either that or we add 14 colours of bed using static hardware colouring.
PR is dead.

DS-Minetest:
>PR is dead.

It's not dead, it's WorkInPause.

paramat:
Ok, more dev opinions are needed.

I'm requesting static hardware colouring and all wool colours, still :-1: for param2 colouring and 8 colours.

DS-Minetest:
Hm, having 16 colors maybe wouldn't be too hard using the bed api to register more beds with the same defs apart from the palette.

Ezhh:
Strongly in favour of all colours, and the colours matching the wool. For example, orange that looks like orange instead of brownish orange as in the screenshots. (Using a texture pack just to get nice colours isn't an answer.)

C1ffisme:
Would it be impossible to make a second set of bed items for the last few dye colors we have? Minecraft already does this with `minecraft:wood` (For Oak, Birch, Jungle and Spruce) and `minecraft:wood2` (Which is meant for Dark Oak and Acacia).

DS-Minetest:
@C1ffisme That's what I meant.

paramat:
Wut? Are you suggesting 8 colours using param2 colouring, then add the remaining colours using static colouring? That's really hacky, they should all be one or the other.

Note we have a rule that we close PRs that have no response from the author in a month, so this could have been closed by now. Please let us know if you are happy to use static hardware colouring for all 14 colours.

> Actually if you would add more nodes instead, it won't only be a few nodes. There are 2 nodes per bed and 2 beds and there are 16 colours => 2*2*16 = 64 => many nodes imo.

What i mean is that 64 nodes is not enough for param2 hardware colouring to be of any significant advantage over static hardware colouring, all that happens is you save 64 nodedefs being sent to the client.

How these new beds relate to the existing registeration code needs careful consideration, the registration code probably needs addtions. Or maybe these new beds should be completely separate?
Also, is the existing red bed replaced by the new red bed or stays separate? You also need to consider the beds that mods and subgames have registered and the effect on those.

DS-Minetest:
>Are you suggesting 8 colours using param2 colouring, then add the remaining colours using static colouring?

No. The remaining colours simply use another bed which has another palette.
>How these new beds relate to the existing registeration code needs careful consideration, the registration code probably needs changing. Or maybe these new beds should be completely separate?
Also, is the existing red bed replaced by the new red bed or does it stay separate?
You also need to consider the beds that mods and subgames have registered and the effect on those, they need to be supported.

```
oberlay_tiles = { -- optional
	bottom = {'Tile definition'}, -- the overlay_tiles of the bottom part of the bed.
	top = {Tile definition} -- the overlay_tiles of the top part of the bed.
},
color = "white", -- optional
palette = "beds_palette.png", -- optional. hardware coloring with param2 is possible if specified
```
The new things in the api are all optional, the subgames can support themselves if they want to. In palette red is the first color, so old beds won't be actively replaced.

paramat:
> No. The remaining colours simply use another bed which has another palette.

I see, that's still very messy and there's no advantage to param2 colouring.

Fixer-007:
Can you explain how exactly it works from usability standpoint? There will be tons of beds in creative inventory or they will be hidden with a hint "coloured" in classic ones descriptions?

DS-Minetest:
@Fixer-007 In creative are only items that are in `minetest.registered_items` (or `…_nodes`, `…_craftitems` or `…_tools`, depending on the selected tab). Ergo in creative will only be red beds.

paramat:
Multiple red beds that have the varying colours described in the tooltip? Or just 1 red bed? If 1 how would creative players select colured beds?
I thought item stacks could display colours now?

DS-Minetest:
Two red beds, one "simple" and one "fancy".

paramat:
How would a player use creative to place coloured beds then?

DS-Minetest:
Filling the creative inventory with beds would be meh.
Maybe an `on_rightclick` or `on_punch` could be added that colors the bed depending on the color of the dye, the player is holding, if player is in creative and holds a dye.

DS-Minetest:
Oops. >_<

Fixer-007:
And how coloring works in game? You hit bed with the dye?

paramat:
> Filling the creative inventory with beds would be meh.

Only 14. The whole idea of creative is that it contains every node, all colours of wool, all colours of dye, why would beds be different?
In creative players want to find and place a colour of bed instantly, they do not want to have to colour each one individually.

This PR is a mess with many issues of bad implementation :-1:

jastevenson303:
There's 15 dye colors including white.  We will have white beds, too, I hope?  Otherwise colored beds would support all dye colors, except one.

paramat:
Oops 15 ok, yes white too.

DS-Minetest:
@Fixer-007 You get a bed of the same color as the one the wool has that was used at crafting.

@paramat Showing all colored beds in creative would only be possible by changing something in creative api (which could also be good for other mods but should be done by another PR) or by adding the beds to `minetest.registered_items` and co..

paramat:
.. or by using static hardware colouring as i request, or even normal texture colorisation which is fine also. For only 30 nodes hardware colourisation really isn't essential.

paramat:
Sorry but this PR is unacceptable and neglected, and you seem unwilling to respond to requests for improvement, but instead suggest other unacceptable ideas. This needs to be done by someone else. Closing.

PR #1796, opened 2017-06-24
new crack anylength

tobyplowy:
a new crack anylength i made
the current one is too saturated and doesent blend with the original color of the node underneath

here you can see what it looks like 
clay 
8fd5e8e2-47b2-11e7-9307-1a9f06cadea4
stone
8fd2ed2c-47b2-11e7-9e2a-39382cfd0eb2
obsidian
8fd8be96-47b2-11e7-8af5-a9fa2382b452


paramat:
An improvement over your last submission, more visible on stone and less grey pixels on obsidian. 8 frames is good too. Will test.

theraven262:
Thats some really nice work, though I would advise making a texture that has more frames, like the crack in Carbone subgame.
Not that I don't like this one, it is much better than the old one, but if the texture is already being changed it would be wise to take the chance.
Others might not agree with this, since 8 frames is already an improvement, and making more frames would require some additional work for questionable success.

tobyplowy:
@theraven262 adding More frames is really hard I already had to add 2 extra frames to get to 8 and that was a nightmare

Adding even more is just not going to happen the animation was originally made for 6 frames

paramat:
8 frames is fine.

GreenXenith:
Tested. Works :+1:

paramat:
I prefer it, but it could perhaps have more bright pixels, as it is quite subtle on several nodes, especially coalblock.

tobyplowy:
> It could perhaps have more bright pixels

That will make it look bad on obsidian

paramat:
Lets see what other devs think.

tobyplowy:
@paramat do you whant more bright pixels know that it will make obsidian look odd
How will the other devs know about this issue?

paramat:
The crack texture looks odd on many nodes already, i think it being visible is more imporatant, which is why the current crack has so much contrast (probably a little too much).

tobyplowy:
@paramat I'm confused with the above comment

> The crack texture looks odd

So my texture is not good?

> i think it being visible is more imporatant

So I need to add more contrast?

paramat:
> more bright pixels know that it will make obsidian look odd

I mean don't be too concerned with obsidian looking odd because many nodes look odd when they have a crack added, like sand, leaves etc. So 'looking odd' isn't much of a problem, visibility is more important.

Yes i think more contrast, but less than the current MT crack.

octacian:
Isn't it possible to have transparency in the crack texture? I did it in one of my subgames and found that it nearly entirely dealt with the issue of some nodes looking weird when the crack texture was applied.

tobyplowy:
@octacian that's already the case

@paramat ok will try

tobyplowy:
@paramat I don't know if it updated but I added more contrast

paramat:
Ok will test.

tobyplowy:
@paramat also sorry for the long wait :/

paramat:
Better but i'm still unsure due to visibility on dark nodes, coalblock and obsidian:

screenshot_20170815_144244

screenshot_20170815_144327

screenshot_20170815_144502

screenshot_20170815_144420

It seems to have too many greyish pixels and not enough whitish pixels. But let's see what others think.

tobyplowy:
@paramat i Cant make it look good on light and dark textures it's one or the other :/
Unless theres 2 crack animations (but that is residues)

paramat:
The current texture is clear on both light and dark nodes, because it has many light and dark pixels, and fewer grey ones. Your crack pattern is good but the pixel colouring not as clear.

paramat:
No response in a month so closing, but if you come up with an improved texture comment here and we can reopen.

PR #1801, opened 2017-06-27
New Desert Cobblestone Texture

vitali7r:
I've added a new textures file.
default_desert_cobble


DS-Minetest:
:-1: 
This is just the normal cobble texture colored red.
That would give less variation of textures and so it would look worse.

paramat:
This needs texture credits changing in README.txt and `brunob.santos` and the `Attribution-ShareAlike 4.0 International (CC BY-SA 4.0)` section removing from license.txt.
Could you add a screenshot showing this directly next to desert stone?
The existing desert cobble is interesting but has a little too much pink in it for my taste, but i'm not sure if yours is better.

vitali7r:
screenshot_20170628_122055

vitali7r:
screenshot_20170628_125323

tenplus1:
I prefer the current texture since it looks like desert stone slate instead of a cobble type.

paramat:
Yes it's good to have a different structure for desert cobble.
I think the ideal would be using the pixel colours of this submitted texture but with more linear cobbles as in the current.

C1ffisme:
The current desert cobble texture seems to have white cement or something sticking the cobble together. This is not the case with regular cobblestone.

GreenXenith:
This texture looks nice, but IMO it's a bit 'shiny'. Needs to look a bit more dull.

azekillDIABLO:
actually, the default texture is very nice but the desert stone is what should be actually a bit different from normal stone!

TumeniNodes:
Maybe simply reducing the brightness of the mortar of the current desert cobble could be a workable compromise?
orig and rework deseert cobble

this is the reworked tile if anyone wishes to try it
default_desert_cobble

paramat:
This has potential, thanks.

TumeniNodes:
here is a xcf if someone wants to experiment, [https://drive.google.com/open?id=0B4iQfHPKm189WDUwbHpWRjFvaEk](url)
and the tile needs to be run through pngcrush

paramat:
:-1: for the PR, i think desert cobble should remain a special and different texture, darkening the mortar is a good idea.

TumeniNodes:
Do you want a new PR done for the darker mortar, or would this be on hold for now?

paramat:
A PR for darker mortar is welcome, i would like to make the texture less bright pink.

paramat:
See #1859 for an alternative.

paramat:
sfan5 disapproves http://irc.minetest.net/minetest-dev/2017-08-03#i_5032482

PR #1803, opened 2017-06-27
add uses for tin

ThomasMonroe314:
I think that tin needs to be more useful as it is so common, so i have switched the steel_bottle from vessels to tin_jug.

PS: the texture did not need to be changed due to the nature of tarnished tin
PS2: more commits may be coming, so don't merge too quickly.


DS-Minetest:
Don't forget to alias the items or there will be unknown items.

ThomasMonroe314:
gah , thanks @DS-Minetest 
I'll do that

ThomasMonroe314:
i need to undo what i did and create a branch on my side

PR #1847, opened 2017-07-24
Reexport the player model wiht split edges

TeTpaAka:
I hope this fixes it. I did what was said in https://github.com/minetest/minetest_game/issues/1846. It didn't help, so I also added a triangulate modifier and applied it.

I tested the model with the advanced lighting patch and it seemed to work. Someone else should test it, too, though. Maybe someone with more experience with exporting b3d files can look into it if it is still wrong.


DS-Minetest:
>Someone else should test it, too, though.

Tested, it works.

>I did what was said in #1846. It didn't help

Hm, as I did the things described in #1848, it helped. Maybe you made a mistake.

paramat:
Will test.

paramat:
screenshot_20170725_221508

Fixes the issue.

paramat:
#1848 also changes the .blend file, while this PR doesn't include that, is that the problem?

paramat:
I prefer #1848 which works.

paramat:
Player goes dark below y = 0.

TeTpaAka:
If it works, I'm fine with https://github.com/minetest/minetest_game/pull/1848, so I close this PR. He probably knew better than I what needed to be done.

PR #1848, opened 2017-07-25
reexport player model

### LINKED ISSUE #1846: Player model needs improvement

DS-Minetest:
Fixes #1846.
Edit: Oops, there is already a PR for this, sorry. >\_<
Edit2: But this one was chosen. ¯\\\_(ツ)_/¯


paramat:
screenshot_20170725_222235

This also works, and includes both files and less editing, so this may be best.

JurajVajda:
`build: 201707250702-0~31b84ce~ubuntu17.04.1`
Is it just me or am i missing something ? ..when you are below y = 0 the character becomes very dark, tried with these models and also with current one in dev. If its a known issue then you can disregard this comment.

screenshot_20170726_005113-min
screenshot_20170726_005122-min

paramat:
Will test, not a known bug.

paramat:
Confirmed, thanks.

paramat:
#1852

JurajVajda:
@paramat no problem,thanks for creating an issue for it.

PR #1905, opened 2017-09-29
Do not give creative priv to admin.

lhofhansl:
Related to minetest [#6460](https://github.com/minetest/minetest/pull/6460)

Admin does not get creative priv by default (which changes gameplay in annoying ways), but can still grant it to self.

### DIFF
--- a/mods/creative/init.lua
+++ b/mods/creative/init.lua
@@ -2,7 +2,8 @@ creative = {}
 
 minetest.register_privilege("creative", {
 	description = "Allow player to use creative inventory",
-	give_to_singleplayer = false
+	give_to_singleplayer = false,
+	give_to_admin = false
 })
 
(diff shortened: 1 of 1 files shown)


lhofhansl:
Turns out that this has to be done in every mod that define a priv that changes game behavior.

SmallJoker:
Obsolete PR. `give_to_admin` defaults to `give_to_singleplayer`.

lhofhansl:
Not sure this should be obsolete. The defaulting is done so that mods to not *have* to change anything. Admin and singleplayer should still be called out independently, IMHO.

SmallJoker:
I agree that extending it by explicit parameters is easier to understand how it works.
This is a trivial change without any effect so why not? *shrug*
Will merge in the next minutes.

PR #1921, opened 2017-10-21
Add loot to dungeons

sfan5:
values are currently tuned so that <i>most</i> qualified dungeons have at least one chest
qualified = not consisting solely of <i>small</i> rooms

findable items: 
* buckets (empty or with lava/water)
* stick [common]
* flint, glass fragments, rail
* string, wheat, apples
* cotton seed/cactus
* coal [common]
* gold- and steel ingots
* mese crystal [rare]
* wooden sword [common]
* stone pick
* diamond axe [very rare]
* dirt, desert sand/normal sand [common, above/near surface only]
* obsidian, mese [rare, deep underground only]


cx384:
This should really be a part of the minetest game but for now I'm using https://github.com/BlockMen/dungeon_loot or https://forum.minetest.net/viewtopic.php?t=17005.

Fixer-007:
>glass fragments

Why? Is there any reason for its existence in minetest, I've never heard anyone using it and never used it myself.

sfan5:
They are part of vessels to allow recycling glass bottles (you can cook them into glass).
I haven't heard of those before either but they are not 100% useless so why not.

rubenwardy:
This should have an API to add items

paramat:
Will review and test.

mods/dungeon_loot/loot.lua
```diff
@@ -0,0 +1,56 @@
...
+
+	-- natural materials
+	{name="default:desert_sand", chance=0.8, count={4, 32}, y={-64, 32768}, types={"desert"}},
+	{name="default:sand", chance=0.8, count={4, 32}, y={-64, 32768}, types={"normal"}},
+	{name="default:dirt", chance=0.6, count={2, 16}, y={-64, 32768}},
+	{name="default:obsidian", chance=0.25, count={1, 3}, y={-32768, -512}},
+	{name="default:mese", chance=0.15, y={-32768, -512}},
+}
+
+dungeon_loot.register = function(t)
```
rubenwardy:
```lua
function dungeon_loot.register(t)
```

mods/dungeon_loot/mapgen.lua
```diff
@@ -0,0 +1,159 @@
...
+	for n = 1, count do
+		local idx = rand:next(1, #list)
+		table.insert(ret, list[idx])
+		table.remove(list, idx)
+	end
+	return ret
+end
+
+local function find_walls(cpos)
+	local wall = ItemStack("mapgen_cobble"):get_name()
```
rubenwardy: does this have any benefits over the following? Seems a bit hacky to me.

```lua
minetest.registered_aliases["mapgen_cobble"]
```

sfan5: any benefits? no
i wouldn't call this hacky tho, just <i>clever</i>

sfan5:
@paramat 
Started a new world, flew around quite a bit while profiling.
Here's the result (LuaJIT): http://sprunge.us/Ecdh

Fixer-007:
I know for a fact people love using stone_with_ore kinds of blocks on servers for decoration - can they be used as rare loot, or instead there will be maybe different approach like special pickaxe that picks ups original blocks? Anyone likes idea? like/dislike

paramat:
1.4 ms max, fine.

mods/dungeon_loot/mapgen.lua
```diff
@@ -0,0 +1,159 @@
...
+				-- abort even if it wasn't a wall cause something is in the way
+				break
+			end
+		end
+	end
+
+	return {
+		walls = ret,
+		size = {x=mindist.x*2, z=mindist.z*2},
+		type = wallnode == wall_sand and "desert" or "normal"
```
paramat: Type "desert" should be for classic desert not sandstone desert.
We actually have 4 types of dungeon now:
Desert stone walls in biome "desert" (biome stone default:desert_stone).
Sandstone brick walls in biome "sandstone desert" (biome stone default:sandstone).
Cobble / mossycobble walls in many biomes (biome stone default:stone).
Then there is a recently added dynamic dungeon wall code, if the above 3 biome stone types are not found the dungeon wall node is set to the biome stone, so the 4th dungeon type is:
Ice walls in "icesheet" biome (biome stone default:ice).

mods/dungeon_loot/mapgen.lua
```diff
@@ -0,0 +1,159 @@
...
+		end
+	end
+end
+
+
+minetest.register_on_generated(function(minp, maxp, blockseed)
+	local poslist = minetest.get_mapgen_object("gennotify")["dungeon"]
+	if poslist == nil then return end
+
+	local noise = minetest.get_perlin(10115, 4, 0.5, 1)
```
paramat: Please use 1 octave for noise, the visual quality of the noise is unimportant for chests.

mods/dungeon_loot/mapgen.lua
```diff
@@ -0,0 +1,159 @@
...
+
+	local dirs = { {x=1, z=0}, {x=-1, z=0}, {x=0, z=1}, {x=0, z=-1} }
+	local get_node = minetest.get_node
+
+	local ret = {}
+	local mindist = {x=0, z=0}
+	local min = function(a, b) return a ~= 0 and math.min(a, b) or b end
+	local wallnode
+	for _, dir in ipairs(dirs) do
+		for i = 1, 8 do -- 8 = max room size / 2
```
paramat: Dungeon max room size is 18x18 so this can be 9.
https://github.com/minetest/minetest/blob/master/src/mapgen.cpp#L891

paramat:
One thing to consider is that we now have the ability to have multiple dungeons in 1 mapchunk, and in new core mapgens i am experimentng with creating structures by greatly increasing this number. I feel there should be a limit to the number of dungeon room positions that are processed per mapchunk because the mod runs wall finding code (using 'get node') for every room.

Max number of rooms per dungeon is 16 https://github.com/minetest/minetest/blob/master/src/mapgen.cpp#L859 so please can the code limit to the first 16 rooms? Otherwise mapchunks with multiple dungeons will then have an excessively high number of chests.

mods/dungeon_loot/mapgen.lua
sfan5: Maybe it was my mistake but i couldn't get the room detection code to detect the first type, is it possible that those are missing from gennotify?

I didn't even know that the last type existed, will add support for it.

sfan5:
> I feel there should be a limit to the number of dungeon room positions that are processed per mapchunk because the mod runs wall finding code (using 'get node') for every room.

For performance reasons?

> Otherwise mapchunks with multiple dungeons will then have an excessively high number of chests.

The amount chests per mapchunk is picked in advance, the amount of rooms in a single mapchunk does not change that. The only reason I can see for this limit is performance.

sfan5:
also is this my code being broken or does any part of the mapgen run after on_generated?

Fixer-007:
Flint feels to frequent to me

paramat:
> The amount chests per mapchunk is picked in advance, the amount of rooms in a single mapchunk does not change that. The only reason I can see for this limit is performance.

Ok.
Yes for performance reasons, all those 'get node's.

> do you know if this is my fault or does any part of the mapgen run after on_generated?

What exactly, the strange intersection of rooms? That's 2 dungeons intersecting, they overgenerate into neighbour mapchunks so sometimes intersect, that also causes the floating chests. the chest was properly placed in the 1st, the 2nd removed the floor from under it.

paramat:
You could probably avoid floating chests and chests turned into walls by not placing a chest within 16 nodes of a mapchunk border.

sfan5:
> Does the metadata remaining cause 'chest' to be displayed when pointing at it?

Haven't checked whether inventory meta is still present, but yes this is the case.

Do you think the limit of rooms processed per mapchunk (16) I've added should be lowered?

paramat:
16 is ok, most dungeons have far less than 16 anyway so it will average to 9 per mapchunk.

paramat:
When walls are created, there is only a check for tge 'untouchable' flag https://github.com/minetest/minetest/blob/master/src/dungeongen.cpp#L299
that flag is
```
#define VMANIP_FLAG_DUNGEON_UNTOUCHABLE (\
		VMANIP_FLAG_DUNGEON_INSIDE|VMANIP_FLAG_DUNGEON_PRESERVE)
```
so we can avoid chests beibg replaced by a wall node by marking any node that is not 'ground content' as 'preserve' here https://github.com/minetest/minetest/blob/master/src/dungeongen.cpp#L109

mods/dungeon_loot/loot.lua
```diff
@@ -0,0 +1,62 @@
+dungeon_loot.registered_loot = {
+	-- buckets
+	{name="bucket:bucket_empty", chance=0.55},
```
paramat: Please can we have spaces around binary operators throughout the commit? It really helps readability.

mods/dungeon_loot/mapgen.lua
```diff
@@ -0,0 +1,170 @@
...
+		table.insert(poslist, entry)
+	end
+	if #poslist == 0 then return end
+
+	local noise = minetest.get_perlin(10115, 4, 0.5, 1)
+	local rand = PcgRandom(noise3d_integer(noise, poslist[1]))
+
+	local candidates = {}
+	-- process at most 16 rooms to keep runtime of this predictable
+	local no_process = math.min(#poslist, 16)
```
paramat: 'no' is a confusing abbreviation of 'number', because it reads as 'no', 'num' is better.

mods/dungeon_loot/mapgen.lua
```diff
@@ -0,0 +1,170 @@
...
+	for i = 1, no_process do
+		local room = find_walls(poslist[i])
+		-- skip small rooms and everything that doesn't at least have 3 walls
+		if math.min(room.size.x, room.size.z) >= 4 and #room.walls >= 3 then
+			table.insert(candidates, room)
+		end
+	end
+
+	local no_chests = rand:next(dungeon_loot.CHESTS_MIN, dungeon_loot.CHESTS_MAX)
```
paramat: 'no' is a confusing abbreviation of 'number', because it reads as 'no', 'num' is better.

mods/dungeon_loot/mapgen.lua
```diff
@@ -0,0 +1,170 @@
...
+		-- choose place somewhere in front of any of the walls
+		local wall = room.walls[rand:next(1, #room.walls)]
+		local v, vi -- vector / axis that runs alongside the wall
+		if wall.facing.x ~= 0 then
+			v, vi = {x=0, y=0, z=1}, "z"
+		else
+			v, vi = {x=1, y=0, z=0}, "x"
+		end
+		local chestpos = vector.add(wall.pos, wall.facing)
+		local off = rand:next(-room.size[vi]/2 + 1, room.size[vi]/2 - 1)
```
paramat: Spaces around '/'.

mods/dungeon_loot/mapgen.lua
```diff
@@ -0,0 +1,170 @@
...
+local function find_walls(cpos)
+	local wall = minetest.registered_aliases["mapgen_cobble"]
+	local wall_alt = minetest.registered_aliases["mapgen_mossycobble"]
+	local wall_ss = minetest.registered_aliases["mapgen_sandstonebrick"]
+	local wall_ds = minetest.registered_aliases["mapgen_desert_stone"]
+	local is_wall = function(node)
+		return table.indexof({wall, wall_alt, wall_ss, wall_ds}, node.name) ~= -1
+	end
+
+	local dirs = { {x=1, z=0}, {x=-1, z=0}, {x=0, z=1}, {x=0, z=-1} }
```
paramat: Unnecessary spaces inside outer braces (Linux kernel coding style).

mods/dungeon_loot/README.txt
```diff
@@ -0,0 +1,7 @@
+Minetest Game mod: dungeon_loot
+============================
```
paramat: 3 more '='.

mods/dungeon_loot/README.txt
```diff
@@ -0,0 +1,7 @@
...
+See top-level LICENSE.txt for license information.
```
paramat: The top level file says:
```
License of source code
----------------------
Copyright (C) 2010-2012 celeron55, Perttu Ahola <celeron55@gmail.com>
See README.txt in each mod directory for information about other authors.
```
Each mod should have it's own license.txt for consistency, it is needed to mention you as the author.

mods/dungeon_loot/loot.lua
sfan5: will do except for vectors, writing `{x = 0, y = 0, z = 1}` wastes a lot of space IMO

AntonBoch1244:
Squash commits.

mods/dungeon_loot/mapgen.lua
```diff
@@ -0,0 +1,168 @@
+minetest.set_gen_notify({dungeon = true, temple = true})
+
+local function noise3d_integer(noise, pos)
+	return math.abs(math.floor(noise:get3d(pos) * 2147483647))
```
SmallJoker: `* 0x7FFFFFFF` would be more readable IMO

AntonBoch1244: Accept.

SmallJoker:
Code looks good. The mod depends at least on `default` to have chests to place. The other used mods are a minor issue, as they're all not required during load time.

@AntonBoch1244 A :+1: reaction is already sufficient to express your opinion. PRs are always squashed on merge. Several commits make it easier to track down bugs or changes that have been made.

AntonBoch1244:
@SmallJoker well i learn something today.

mods/dungeon_loot/mapgen.lua
paramat: Yes i suspected this, it's a bad habit you have ;] and no excuse. Anyway i won't insist.

mods/dungeon_loot/loot.lua
paramat: Readability is more important, space is good. But ok.

paramat:
Thanks, all is fine for me, feel free to merge.

paramat:
@AntonBoch1244 please don't use the 'approval' feature, that's more for core devs, it's better to express your approval another way.

sfan5:
It's not that you _can't_ use it, rather it's _not useful_.

Helping with reviewing is fine (and encouraged).
However approvals are not useful to us as the development ruleset requires approval by at least two members of the (in this case) minetest_game team, "outsiders" don't matter in this case. Furthermore we can't be sure that someone else has applied the same standards during review.

AntonBoch1244:
@sfan5, thanks for answer.

MarkuBu:
Great idea, but sometimes chests are rare (I need to search in 10 dungeons to find 1 chest) and the loot could be more valuable than a broken wooden sword and strings.

MarkuBu:
And how about removing items from the loot list? For example if I remove items with unregister_item, but I can't remove it from the default loot list (except I would edit the code), than I get unknown items in the chests

MarkuBu:
Get an error message if I remove an item

```
2017-10-27 19:06:45: ERROR[Main]: ServerError: AsyncErr: Lua: finishGenRuntime error from mod 'default' in callback environment_OnGenerated(): ...etest/games/farlands050/mods/dungeon_loot/mapgen.lua:91: attempt to index local 'itemdef' (a nil value)
2017-10-27 19:06:45: ERROR[Main]: stack traceback:
2017-10-27 19:06:45: ERROR[Main]: 	...etest/games/farlands050/mods/dungeon_loot/mapgen.lua:91: in function 'populate_chest'
2017-10-27 19:06:45: ERROR[Main]: 	...etest/games/farlands050/mods/dungeon_loot/mapgen.lua:165: in function '?'
2017-10-27 19:06:45: ERROR[Main]: 	...markus/dev/minetest/bin/../builtin/game/register.lua:412: in function <...markus/dev/minetest/bin/../builtin/game/register.lua:392>
2017-10-27 19:06:45: ACTION[Server]: singleplayer leaves game. List of players: 
Loaded texture: /home/markus/.minetest/games/farlands050/menu/background.1.png
Loaded texture: /home/markus/.minetest/games/farlands050/menu/header.png
```

sfan5:
> Great idea, but sometimes chests are rare (I need to search in 10 dungeons to find 1 chest) and the loot could be more valuable than a broken wooden sword and strings.

bad luck, that's how randomness works

> And how about removing items from the loot list?

While the docs advise against it, it's fine to modify the `dungeon_loot.registered_loot` table at load time to remove the items you don't want.

MarkuBu:
Well, if the doc says "don't edit" it normally means "don't edit".

And maybe you can catch the error and print a warning to the console

MarkuBu:
Why didn't you use an indexed table? Now you have to loop over the table to find the right name instead of simply write

`dungeon_loot.registered_loot['default:mese_crystal'] = nil`

I recommend to add an "official" way to remove items from the table

sfan5:
> Why didn't you use an indexed table?

because it doesn't work
```lua
	{name = "bucket:bucket_water", chance = 0.45, types = {"sandstone", "desert"}},
	{name = "bucket:bucket_water", chance = 0.45, y = {0, 32768}, types = {"normal"}},
```

MarkuBu:
Ok, good point. But again, I recommend to add a remove function

paramat:
@AntonBoch1244 yes what sfan5 wrote, reviews and suggestions are appreciated.
However, using this particular approval feature is problematic because it looks very official and is used by core devs to mean an official approval, so your use of it looks like a core dev approval and could be misunderstood as one at a quick glance through the thread.

Also, your comment saying 'approved' is not suitable wording because 'approved' is our official teerm for a core dev approval, so it also looks like someone acting like a core dev.
So this is why we ask people not to use it.
Please could you change it to a +1 or thumbs up or whatever?

paramat:
> sometimes chests are rare (I need to search in 10 dungeons to find 1 chest)

Seems ok, dungeons do get very dense and we shouldn't give too much or valuable items too easily as this would spoil gameplay and progression. So erring on the side of undergenerosity is best.

MarkuBu:
> dungeons do get very dense

That's one thing I mentioned in an earlier issue. Sometimes dungeons are dense, sometimes you have a huge area without any dungeons at all.

Imagine how frustrating it is if you are walking and digging for hours and if you finally find dungeons you don't find any loot. And after some more hours you find the first chest and all you get is a broken wooden sword and some strings.

After such an experience, you won't bother searching dungeons anymore.

Wuzzy2:
It would also make sense to add otherwise unobtainable items in chests for v6 mapgens.
E.g. bush saplings, acacia sapling, aspen sapling, silver sand, (maybe) coral skeleton.

I imagine it would be quite interesting to populate your world with rare plants. :-)

MarkuBu:
This is an interesting place for a chest. It blocks a staircase

screenshot_20171029_172501

MarkuBu:
Finally I just clear registered_loot and rewrite it.

But would it be possible to add settingtypes support for some hard coded settings like the noise parameter? Maybe some people want to change some settings and currently they have to do it in the code.

PR #1929, opened 2017-10-27
Carts: Correct visuals of going up/down

### LINKED ISSUE #1927: Minecarts and slopes.
screenshot_20170113_125822

This doesn't just affect the First Person view:

screenshot_20171026_170149

screenshot_20171026_170150

SmallJoker:
Fixes #1927 

screenshot_20171027_125531
screenshot_20171027_125658

Ugly side-effects:
* The player stays rotated like that for a *few seconds* after detaching. Seems to be a core issue.
* The eye offset is changed, but it's not bound to the player head when rotating around. Core issue as well (?)

### DIFF
--- a/mods/carts/cart_entity.lua
+++ b/mods/carts/cart_entity.lua
@@ -228,7 +228,8 @@ local function rail_on_step(self, dtime)
 	if stop_wiggle or vector.equals(dir, {x=0, y=0, z=0}) then
 		vel = {x = 0, y = 0, z = 0}
 		local pos_r = vector.round(pos)
-		if not carts:is_rail(pos_r, self.railtype) then
+		if not carts:is_rail(pos_r, self.railtype)
+				and self.old_pos then
 			pos = self.old_pos
 		elseif not stop_wiggle then
 			pos = pos_r
@@ -292,8 +293,12 @@ local function rail_on_step(self, dtime)
 
 	self.object:set_acceleration(new_acc)
 	self.old_pos = vector.new(pos)
+	local old_y_dir = self.old_dir.y
 	if not vector.equals(dir, {x=0, y=0, z=0}) and not stop_wiggle then
-		self.old_dir = vector.new(dir)
+		self.old_dir = dir
+	else
+		-- Cart stopped, set the animation to 0
+		self.old_dir.y = 0
 	end
 	self.old_switch = switch_keys
 
@@ -319,13 +324,15 @@ local function rail_on_step(self, dtime)
 		rail_on_step_event(railparams.on_step, self, dtime)
 		return
 	end
+	-- Re-use "dir", localize self.old_dir
+	dir = self.old_dir
 
 	local yaw = 0
-	if self.old_dir.x < 0 then
+	if dir.x < 0 then
 		yaw = 0.5
-	elseif self.old_dir.x > 0 then
+	elseif dir.x > 0 then
 		yaw = 1.5
-	elseif self.old_dir.z < 0 then
+	elseif dir.z < 0 then
 		yaw = 1
 	end
 	self.object:set_yaw(yaw * math.pi)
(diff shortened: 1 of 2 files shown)


paramat:
Impressive we can actually tilt the player.
Having 'set attach' and 'set eye offset' every step seems intensive.
Can these be reset only when flat->sloped or sloped->flat is detected? Te code already knows when the rail is flat or sloped so seems possible.

SmallJoker:
@paramat https://github.com/minetest/minetest_game/pull/1929/files#diff-83985b3ea786ae335dc94e7a3bc63b14R318
The trick behind the smoother movement is also skipping position and velocity updates when they're not necessary.

paramat:
Ok it almost does this already, but it will still update every node if the cart is being punched or is accelerating. perhaps a new variable `update.tilt` that is determined by a change in 'cart_dir.y`?

paramat:
> The eye offset is changed, but it's not bound to the player head when rotating around. Core issue as well (?)

My guess is it's to do with the fact that although the visible player model is rotated to face the direction of cart movement, the camera is not.
In testing, by looking in certain directions the camera ends up in very weird places, like outside the cart to one side of a slope, or low down almost level with the rails.

C1ffisme:
Nice! I'm glad this can get fixed.

stujones11:
Re-setting the attachment like this might look okay in singleplayer but it will not look well the effects of server lag. The rotation should be done client-side by inheriting the cart's parent bone rotation.

https://github.com/minetest/minetest_game/issues/1927#issuecomment-340287477

paramat:
Currently the problems with this outweigh the benefits.

DS-Minetest:
I'm not sure whether it looks good how sam sticks on the ground of the cart with his shoes, even on high velocities. Maybe it would be nice if there was a new animation frame for standing in the cart and clutching its borders.

stujones11:
> a new animation frame for standing in the cart and clutching its borders.

That would look pretty cool and should, in theory, be possible by just setting the player bone positions, however, I am not sure how well that behaves with attachments, likely another bug-fest :P

DS-Minetest:
>I am not sure how well that behaves with attachments

I assume, it would work as good as with boats (which use the sitting animation).

stujones11:
> I assume, it would work as good as with boats

Indeed, we know that it works with animations but that would require changes to the player model. What I am suggesting is that the bone rotations are set directly. This should be fine as the normal movement animations are disabled by `default.player_attached` though I have never tried this.

DS-Minetest:
>What I am suggesting is that the bone rotations are set directly.

Afaik once a bone's position or rotation is set, the model can not be animated normally again.

What would speak against changes to the player model? `character.b3d` is from and for minetest_game and the carts mod is from minetest_game, too. Why shouldn't the player_api mod do something especially for the carts mod?

stujones11:
> Afaik once a bone's position or rotation is set, the model can not be animated normally again.

That sounds like a bug to me, though I have a feeling you are correct.

TekhnaeRaav:
If there is a problem of performance or outweighed benefits, I think the best solution would be to not tilt the minecart at all.

C1ffisme:
@TekhnaeRaav That would be a simple fix, although on the other hand, if a simple issue of rotating a player model is a problem... I think it's worth dev-time to modify the API to allow pitching a player model... Maybe...

SmallJoker:
Updated the PR with new changes from boost_cart. The player now gets rotated by 30° (instead of 45° previously). Meanwhile there's also a trivial bugfix for an edge-case crash (additional `and self.old_pos` check).

paramat:
How about the weirdness caused by view change?

SmallJoker:
@paramat That effect was reduced but cannot be solved properly without an engine change.

paramat:
Ok will retest.

paramat:
Tested. The forward view is fixed but all other view directions are still very weird, the way it swings the player around is nauseating too. Dependent on view direction i am outside the cart and low down to one side of the slope, or at track level. When riding in a cart players will often be looking around in many directions and the new weirdness is more of a problem than the original problem.

👎 i think the view weirdness needs to be fixed in the engine first.

screenshot_20180214_062439

screenshot_20180214_062444

screenshot_20180214_062449

paramat:
Reopen on improvement.

PR #2035, opened 2018-01-29
add obsidian_pane

TumeniNodes:
adds an obsidian glass pane to xpanes


SmallJoker:
:+1: sure, why not?

TumeniNodes:
That was quick, thank you.
I literally just realized they didn't exist yet, in the middle of a build

mods/xpanes/init.lua
```diff
@@ -160,6 +160,19 @@ xpanes.register_pane("pane", {
 	}
 })
 
+xpanes.register_pane("obsidian_pane", {
+	description = "Obsidian Glass Pane",
+	textures = {"default_obsidian_glass.png","xpanes_pane_half.png","default_obsidian.png"},
+	inventory_image = "default_obsidian_glass.png",
+	wield_image = "default_obsidian_glass.png",
+	sounds = default.node_sound_glass_defaults(),
+	groups = {snappy=2, cracky=3, oddly_breakable_by_hand=3},
```
Ezhh: Obsidian glass, unlike normal glass, does not have the oddly_breakable_by_hand group, so should not have that here. Full support from me if that's removed.

TumeniNodes: done

paramat:
"xpanes_pane_half.png" is probably wrong for this pane as that is a whiteish texture, will test now.
I support the concept though.

paramat:
Actually it works, what are "xpanes_pane_half.png" and "xpanes_white.png" used for then, they don't seem visible?

TumeniNodes:
xpanes_white is used for the bottom/top edges of default glass pane
this is why I set it for default obsidian for obsidian pane
the "half" covers when they connect and angle to connect

paramat:
Yes just figured out the 3rd texture. Concerned about the 2nd texture being glassy-white though.

TumeniNodes:
which texture is that? I am not sure I understand what you mean

TumeniNodes:
I actually played around with the frame for glass, filling in the alpha bits with the blueish highlight so might be able to replace that edge with that or, I may need it's own dedicated _side.png if that is not a problem

I will check it out in a few moments

TumeniNodes:
a quick link to what I did with the default_glass_pane (pay no attention to the obsidian or glass details)
https://imgur.com/a/fNvPx

paramat:
The imgur screenshot looks worse to me, i think we should keep the transparent pixels as they give more of an impression of glassyness.

paramat:
> which texture is that? I am not sure I understand what you mean

Sorry, i was referring to "xpanes_pane_half.png" which is blue-white, but it is not visible in the obsidian pane.

TumeniNodes:
gotcha, I will work on this right now

TumeniNodes:
that xpanes_half.png I need to look into, if it's obsolete, it's outta here

paramat:
New 'xpanes white' is perfect but just needs converting to indexed colour (if that works).

Bars use the same texture for the first 2 tiles, maybe glass panes can too?

paramat:
screenshot_20180130_173958

Yes indexed does work, much better.

paramat:
screenshot_20180130_174241

Obsidian panes will need a dedicated frame side texture, as you can see that the pixels of 'default obsidian png' are being split down the middle into half-width pixels.
Again best to derive the texture from the frame front and i to may need to be a 2x16 pixel texture with both rows identical, such that it appear as unsplit pixels.

paramat:
Actually, it seems a 1x16 texture is stretched so 1x16 is probably ok.

TumeniNodes:
I mistakenly added the files to the main folder but fixed it now, sorry bout that.

Odd, the xpanes_half.png file, if I rename it so it is unused, it makes no change yet, if I delete it from the code, the top and bottom faces go complete alpha...., even though the xpanes_white, and xpanes_obsidian pngs are used for these faces.

I will have to figure that witchery out later tonight

mods/xpanes/init.lua
```diff
@@ -104,7 +104,6 @@ function xpanes.register_pane(name, def)
 		groups = flatgroups,
 		drop = "xpanes:" .. name .. "_flat",
 		sounds = def.sounds,
-		use_texture_alpha = def.use_texture_alpha or false,
```
paramat: Why removed?

mods/xpanes/init.lua
```diff
@@ -149,9 +147,10 @@ end
 
 xpanes.register_pane("pane", {
 	description = "Glass Pane",
-	textures = {"default_glass.png","xpanes_pane_half.png","xpanes_white.png"},
-	inventory_image = "default_glass.png",
-	wield_image = "default_glass.png",
+	textures = {"default_glass.png^default_glass_detail.png","xpanes_pane_half.png","xpanes_white.png"},
+    use_texture_alpha = true,
```
paramat: Wrong indentation.

mods/xpanes/init.lua
```diff
@@ -149,9 +147,10 @@ end
 
 xpanes.register_pane("pane", {
 	description = "Glass Pane",
-	textures = {"default_glass.png","xpanes_pane_half.png","xpanes_white.png"},
-	inventory_image = "default_glass.png",
-	wield_image = "default_glass.png",
+	textures = {"default_glass.png^default_glass_detail.png","xpanes_pane_half.png","xpanes_white.png"},
```
paramat: Why is 'glass detail' needed?

paramat:
You can't remove the texture because it is the 2nd texture, even if unused if you remove it the 3rd texture then becomes the 2nd and the 3rd texture is then missing.
See the iron bars, you can just use the same texture for textures 1 and 2.

mods/xpanes/init.lua
TumeniNodes: ugh, sorry I will fix both of these. I got pulled away and it threw me off focus

TumeniNodes:
I am so inefficient at this, sorry
I should have double checked all these details ahead of time.
Although not sure what happened with renaming Iron Bar desc to Iron bar, as I had not even touched that.

Anyway, I hope this fixes all those mistakes you pointed out

mods/xpanes/init.lua
```diff
@@ -129,7 +129,6 @@ function xpanes.register_pane(name, def)
 		groups = groups,
 		drop = "xpanes:" .. name .. "_flat",
 		sounds = def.sounds,
-		use_texture_alpha = def.use_texture_alpha or false,
```
paramat: Should be left in.

mods/xpanes/init.lua
```diff
@@ -160,6 +159,19 @@ xpanes.register_pane("pane", {
 	}
 })
 
+xpanes.register_pane("obsidian_pane", {
+	description = "Obsidian Glass Pane",
+	textures = {"default_obsidian_glass.png","xpanes_pane_half.png","xpanes_obsidian.png"},
+	inventory_image = "default_obsidian_glass.png",
+	wield_image = "default_obsidian_glass.png",
+	sounds = default.node_sound_glass_defaults(),
+	groups = {snappy=2, cracky=3, oddly_breakable_by_hand=3},
```
paramat: 'oddly_breakable_by_hand' needs removing.

paramat:
screenshot_20180130_205923

screenshot_20180130_205955

I was looking at the 'bar top' texture and noticing how it has detail in a cross formation.
In testing, when creating an 'X' pane you can see that your current textures are missing the detail pixels in one axis.
So 'xpanes white' and 'xpanes' obsidian need to be in the form of 'bar top', as in having the detail drawn 2-pixels wide and in a cross formation.
For a smaller file size don't use alpha in the texture, as with 'bar top'.

These textures need better names too, 'white' is now inaccurate, how about 'xpanes_edge' and 'xpanes_edge_obsidian'?

mods/xpanes/init.lua
TumeniNodes: this is so weird, I actually had not even removed this in my original commit. I am using the current pull for this branch, but it seems to have differences from my 0.4.16 install

is it possible some changes were made since 0.4.16 release, that I am now reverting by "fixing" all of these lines?

mods/xpanes/init.lua
TumeniNodes: this is weird too, I had already removed that in a prior commit.

Ezhh: You added it again in a later one.

mods/xpanes/init.lua
paramat: These lines were the lines changed in the previous commit to xpanes, you may have somehow reverted the last commit.

TumeniNodes:
doing the edge texture does look much better @paramat

paramat:
All seems good now, will now test.

paramat:
Will test.
You just need to remove 'xpanes white' and 'xpanes obsidian'.

TumeniNodes:
apologies, I used drag n drop being lazy, and forgot it would not remove them due to being renamed. done

paramat:
:+1:

TumeniNodes:
Awesome. Thank you everyone

PR #2036, opened 2018-01-29
Stairs Unbind table groups.

Jat15:
Unbind table groups for base block, stairs, slabs.


SmallJoker:
Travis already points out the name conflict. However, `groups = table.copy(groups)` can be used, as modifying the table variable directly discards the reference.

Jat15:
I create a variable new_groups, if it disturbs travis.

Jat15:
Rebase with var "new_groups"

PR #2041, opened 2018-02-06
Adjust kelp placement and add selection box

Ezhh:
This does three things:
- Adds a selection box to the kelp itself.
- When digging kelp you now leave the sand it was in behind, meaning no more ugly holes getting left after collecting it.
- When planting kelp, it no longer places a new sand node; it just adds the kelp to whichever sand node you targeted.

### DIFF
--- a/mods/default/nodes.lua
+++ b/mods/default/nodes.lua
@@ -1523,6 +1523,13 @@ minetest.register_node("default:sand_with_kelp", {
 	inventory_image = "default_kelp.png",
 	paramtype2 = "leveled",
 	groups = {snappy = 3},
+	selection_box = {
+		type = "fixed",
+		fixed = {
+				{-0.5, -0.5, -0.5, 0.5, 0.5, 0.5},
+				{-0.1, 0.5, -0.1, 0.1, 3.5, 0.1},
+		},
+	},
 	node_placement_prediction = "",
 
 	on_place = function(itemstack, placer, pointed_thing)
@@ -1537,7 +1544,11 @@ minetest.register_node("default:sand_with_kelp", {
 			end
 		end
 
-		local pos = pointed_thing.above
+		local pos = pointed_thing.under
+		if minetest.get_node(pos).name ~= "default:sand" then
+			return itemstack
+		end
+
 		local height = math.random(4, 6)
 		local pos_top = {x = pos.x, y = pos.y + height, z = pos.z}
 		local node_top = minetest.get_node(pos_top)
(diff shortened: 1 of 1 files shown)


mods/default/nodes.lua
```diff
@@ -1523,6 +1523,13 @@ minetest.register_node("default:sand_with_kelp", {
 	inventory_image = "default_kelp.png",
 	paramtype2 = "leveled",
 	groups = {snappy = 3},
+	selection_box = {
+		type = "fixed",
+		fixed = {
+				{-0.5, -0.5, -0.5, 0.5, 0.5, 0.5},
+				{-0.1, 0.5, -0.1, 0.1, 3.5, 0.1},
```
paramat: Could this be pixel-aligned for consistency with other plant selectionboxes? So -2 / 16 to 2 / 16 horizontally.

paramat:
Will test.
The node description may need changing, currently it considers the node as kelp plus sand, maybe it could be 'Sand Kelp' or whatever, since the sand no longer comes with it.

Ezhh:
I was actually just thinking it should simply be "Kelp" now.

paramat:
That's ok too.

mods/default/nodes.lua
paramat: Even better:
{-0.125, 0.5, -0.125, 0.125, 3.5, 0.125},

Ezhh: Done. I don't like it being this large, but I doubt anyone should mind too much.

paramat:
:+1: Tested, big improvement.

Ezhh:
@paramat As a last thought, should it have waving = 1?

Looks like this:
kelp_waving

mods/default/nodes.lua
```diff
@@ -1561,6 +1573,10 @@ minetest.register_node("default:sand_with_kelp", {
 		end
 
 		return itemstack
+	end,
+
+	after_destruct  = function(pos, oldnode)
+		minetest.set_node(pos, {name = "default:sand"})
```
sfan5: might be a good idea to add a [node_dig_prediction](https://github.com/minetest/minetest/blob/master/doc/lua_api.txt#L4624) entry for this

Ezhh: Done, thanks.

paramat:
1e9c61f241f6495d944d7e8e21fa5e150e7f8315

Fixer-007:
Forgot to ask, does it spread?

paramat:
I wouldn't mind the box being -1/16 to 1/16 if you prefer.
And / or the box could only extend 1 node up the kelp?
Just random ideas.

Ezhh:
I tried -1/16 to 1/16 and it doesn't match the pixels properly (remember textures are on a diagonal for plantlike), and I thought this was the point of changing it to -2/16 to 2/16.

I think the height of the selection box is fine. I wouldn't want to reduce it.

paramat:
Ah of course, it's fine.

PR #2158, opened 2018-06-28
add fence rails

TumeniNodes:
It's a long shot but, figured, why not?
Adds fence rails, to add the option of longer rails between posts (and possibly more uses)

adds a new node registration for "fence rails"
all wood types added

screenshot_20180628_012823


TumeniNodes:
I apologize. I made a mistake during the last commit.
I then attempted to use the git stash command, to learn it, and learned I do not know it well.
I will re-attempt fixing the indent at line 353 some time tomorrow.

I only have 4hrs left to get some sleep

TumeniNodes:
I did fix line 353 of functions.lua but did not seem to take in this commit?
Perhaps this is due to the last commit still being active with the same file?
But I am not sure how to fix that, if someone could give me some info related to that please.

Added group:wall to connection list. (this looks really nice)
fence rails sample

paramat:
These look good. Since we already have fences longer rails are not high priority so i'm unsure about adding them to MTG, feel sort of neutral at the moment.
I don't like how they join to walls, looks weird and ugly.

benrob0329:
I for one, support fence rails connecting to walls. I'd imagine that they are embedded slightly into the wall, and would make more sense than fences themselves connecting to walls.

paramat:
Sofar wrote the fence and wall code. Looking at the 'connects to' parameters of both it was a concious decision to not connect fences to walls. If we do this then fences need to connect too, and the 'connects to' of fences will need updating too. However for now at least i am -1 for connecting walls and fences. That can be a separate PR.

TumeniNodes:
Honestly, I do not see that fences would need to connect-to walls, just because the fence rails do.
fence rails connect-to / fence, fence_rails, and walls

I did this for a reason.

I am not against fences connecting to walls (I just would not use the combination in any builds, myself. But others might like it, i guess)
I mean, I guess if using in a method of "wall | fence_rail | fence | fence_rail | wall" that could look rather nice.

SmallJoker:
I like those rails connected to walls. Mixed fences could be built this way as shown in this comment: https://github.com/minetest/minetest_game/pull/2158#issuecomment-401261580 -- which looks very nice.

TumeniNodes:
Thank you Krock.
Forgive my ignorance, related to the actual "playing" aspects...
when you mention the rails are "cheap" are you referring to what it takes to craft them? Or what they drop when punched?

paramat:
Ok, i'm not opposed to connection to walls :) Not opposed to adding rails to MTG either, neutral.

'Cheap' was mentioned in context of the crafting recipe.

TumeniNodes:
I made the crafts as this now, with this commit https://github.com/minetest/minetest_game/pull/2158/commits/cb263d22c944f93f21cf09bd379504a8a7c47b23
screenshot_20180630_114655

TumeniNodes:
No idea what Travis' problem is.
Possibly something with my build, idk but, I have not had the issue with any other work I have done anywhere else, under the same build

paramat:
For a build fail, click on 'details' link to see the code issue (long lines).

paramat:
The recipe is unintuitive, the wood nodes make no sense.
Fences are:
```f)
	minetest.register_craft({
		output = name .. " 4",
		recipe = {
			{ def.material, 'group:stick', def.material },
			{ def.material, 'group:stick', def.material },
		}
```
So fence rails should just be sticks, maybe:
```
SSS
---
SSS
```
One stick per rail means the recipe should return 3 rails.

TumeniNodes:
there has to be a reference to the def material, or you don't get the type of wood you want.

I feel it does make sense, (actually, I forgot to change the output number)

But here is my thought.
1 wooden nod has an output of 4 sticks.
using 3 of the def material, the output would be 12 rails
the row of sticks across the top and bottom represent the rails

TumeniNodes:
Just using sticks does not function properly, and fence uses one 'def material' in the center, so would like to keep it as 3 across center, as anything else will just look odd, and could be confused with fence craft

paramat:
> there has to be a reference to the def material, or you don't get the type of wood you want.

Oops i see.
Ok how about representing the rails with wood nodes, and greatly raising the output number:
```
WW
--
WW
```

'default_rail_overlay.png' can't be used as a texture name because it confuses with cart rails, it has to be '..fence_rail..'. Changing the name is a bad way to try to reduce long lines, and the lines are still too long (max is 80 columns with tab size 4). Krock showed in IRC how to correctly split the lines, please do it that way :)

TumeniNodes:
ok, I will make those changes then.

paramat:
These texture lines in normal fence definitions should be split too.

TumeniNodes:
do you want those taken care of here or in a sep PR?

I also see tons of areas where there is no consistency in the order of node names.
i.e. in one section there is acacia, aspen, junglewood, pine, wood / then another section the order is different.
This can be seen all over different areas of the code, and could prob use cleaning up over time. Which I am more than glad to help with.

But, maybe best thing to do is to fix areas needed here, and do other things in later PRs

TumeniNodes:
ok, those lines are fixed now. (not committed yet)
I want to be sure we have the recipe agreed on before I do that, and finish this up.

tenplus1 had mentioned using <wood stick stick / wood stick stick>
and that could have an output of 8 rails

it was mentioned that <wood wood wood / wood wood wood> might be used in another recipe already?

paramat:
> tenplus1 had mentioned using <wood stick stick / wood stick stick>

For fences the 'wood' represents the fence posts and the sticks the rails. For rails there are no posts so 2 different ingredients doesn't make sense to me and that recipe doesn't represent the structure of rails. But we must have wood as you mention, so i suggest only wood ingredients.
```
WW
--
WW
```
Isn't used in MTG as far as i know, any mods using it will just have to change.

> do you want those taken care of here or in a sep PR?

This PR, since that is a closely related change, nevermind any other errors in nodes.lua.

TumeniNodes:
ok got it.

paramat:
Since a wood node creates 4 sticks, with 1 stick per rail, the return would be 16 rails from my recipe.

mods/default/functions.lua
```diff
@@ -327,6 +327,57 @@ function default.register_fence(name, def)
...
+--
+
+function default.register_fence_rail(name, def)
+	minetest.register_craft({
+		output = name .. " 16",
+		recipe = {
+			{  def.material,  def.material,  def.material },
+			{  def.material,  def.material,  def.material },
+		}
+	})
```
paramat: I think vertical separation is clearer, your recipe looks like a horizontal door and is more likely to be in use somewhere else. My suggestion is:
```
+		recipe = {
+			{def.material,  def.material},
+			{"", ""},
+			{def.material,  def.material},
+		}
+	})
```
Also note, you have lots of random spaces inside the brackets, please only add 1 space after the commas.

paramat:
Apart from that all looks good so far.

mods/default/functions.lua
TumeniNodes: ah, ok.

paramat: Your recipe is shown at the top of this comment, it's 3 across. What i show is my suggestion, as i have written: "My suggestion is:", i can't be any clearer ;)

TumeniNodes:
I am not sure if I over did it with the tabs, with the line breaks for the texture defs?
They visually seem to sit right when the code is viewed in Geany and even in text editor but, when viewed here on github, they appear tabbed too far to the right, for the second line of each break.

I used 4 tabs in for each second line. Was that too much?

paramat:
Github uses 8 spaces per tab so makes code look weird, don't worry about that.

Something to consider: Should rails be integrated into 'register fence' to be automatically created when a fence is registered? I think probably yes.

TumeniNodes:
I'm unsure about that.
I did create their own reg though. I see your point, I just don't know how to do that so, that I will leave to you.
It does make sense

paramat:
Should be quite easy if it is done, just move the rail code into 'register fence'.
There might be reasons why this cannot be done (can't think of any at the moment), so more input is needed, but if it should be done it should be done in this PR not afterwards.
Let's see what others think. I don't agree to merging until we consider this.

paramat:
Fences and rails could connect to the same types of nodes, fences could be made to connect to walls, they should if rails do.

TumeniNodes:
This is the only issue I see with adding walls to the 'connect_to' function for fence
screenshot_20180701_150154

paramat:
Looks ok to me, it's like the fencepost is embedded in the wall for support, no worse than fence rails being embedded in a wall :) Also, if a player doesn't like how it looks they don't have to build it that way.

TumeniNodes:
touche'

Fixer-007:
Do they connect to simple cobblestones etc?

TumeniNodes:
> Do they connect to simple cobblestones etc?

If you mean just a basic cobblestone node, then yes.
The width of the rails, is the width of a full node so, they can be placed aside any node to appear connected.
They just needed to be added to connect to fence and wall nodes, so the action will happen from those types.

Ezhh:
The fences try to connect to the rails even when the rails are rotated into positions when they can't connect. I doubt there is any good solution for this, and don't feel it should stop this being merge, but maybe worth thinking about?

railrotate

TumeniNodes:
I'm not really sure why someone would do that in the first place. Unless the fence can be rotated too.
Using paramtype2 = facedir, leaves it vulnerable to screwdriver abuse.
And paramtype2 = facedir is senseless to add to fence, as it will only reorient the texture, not the node itself.

I do not see any way to resolve it.

paramat:
Just noticed, the rails are normal 'fixed' nodeboxes but have 'connects to' fields, surely these should be 'connected nodeboxes' for auto-connection to act like fences?

TumeniNodes:
they do not function properly as connected nodeboxes... trust me, I racked my brain until I realized fixed was the solution

TumeniNodes:
Well, that requires a whole new model or nodebox

paramat:
I might be overlooking something and might be wrong. But the rails will not turn into a fence.
As encouragement i will +1 this PR if done 'correctly' (that is, if i'm correct), and SmallJoker supports, so you will be guaranteed a merge. Currently i'm -1 because rails should be created when a fence is registered.
Let me test first to see if i'm wrong then we'll go from there.

TumeniNodes:
OK..., good luck

paramat:
The registration does have to be separate due to how the register function is, you got that right.

A connected nodebox for fence rails works and auto-connects, also creating some new possibilities like crossed rails.
Note: Below i haven't bothered getting the texture correct, i used the fence texture and overlay.

screenshot_20180702_180814

screenshot_20180702_180848

screenshot_20180702_181007

screenshot_20180702_181131

screenshot_20180702_181447

screenshot_20180702_181634

Is there any desired behaviour missing?

Code is (but texture / overlay needs doing correctly):
```
function default.register_fence_rail(name, def)
	minetest.register_craft({
		output = name .. " 16",
		recipe = {
			{ def.material, def.material },
			{ "", ""},
			{ def.material, def.material },
		}
	})

	local fence_rail_texture = "default_fence_overlay.png^" .. def.texture ..
			"^default_fence_overlay.png^[makealpha:255,126,126"
	-- Allow almost everything to be overridden
	local default_fields = {
		paramtype = "light",
		drawtype = "nodebox",
		node_box = {
			type = "connected",
			fixed = {
				{-1/16,  3/16, -1/16, 1/16,  5/16, 1/16},
				{-1/16, -3/16, -1/16, 1/16, -5/16, 1/16}
			},
			-- connect_top =
			-- connect_bottom =
			connect_front = {
				{-1/16,  3/16, -1/2, 1/16,  5/16, -1/16},
				{-1/16, -5/16, -1/2, 1/16, -3/16, -1/16}},
			connect_left = {
				{-1/2,  3/16, -1/16, -1/16,  5/16, 1/16},
				{-1/2, -5/16, -1/16, -1/16, -3/16, 1/16}},
			connect_back = {
				{-1/16,  3/16, 1/16, 1/16,  5/16, 1/2},
				{-1/16, -5/16, 1/16, 1/16, -3/16, 1/2}},
			connect_right = {
				{1/16,  3/16, -1/16, 1/2,  5/16, 1/16},
				{1/16, -5/16, -1/16, 1/2, -3/16, 1/16}},
		},
		connects_to = {"group:fence", "group:wood", "group:tree", "group:wall"},
		inventory_image = fence_rail_texture,
		wield_image = fence_rail_texture,
		tiles = {def.texture},
		sunlight_propagates = true,
		is_ground_content = false,
		groups = {},
	}
	for k, v in pairs(default_fields) do
		if def[k] == nil then
			def[k] = v
		end
	end

	-- Always add to the fence group, even if no group provided
	def.groups.fence = 1

	def.texture = nil
	def.material = nil

	minetest.register_node(name, def)
end
```
+1 if made a connected nodebox, otherwise -1.
> the action of these, when in use, can be run in a way fences cannot,

They should act like a fence because they are parts of a fence. If we added a new type of rail it would need to be 'raillike' and auto-connect.
> So now, if you place them in spots with anything they connect to, nearby, the rails will turn into a fence

No they stay as rails.

TumeniNodes:
If you build a fence using rails/and fence... and a rail happens to be going alongside a tree..., what happens?
I mean, I can see what happens in 4th shot down but..., I _guess_ it's not all _that_ bad.
I can accept this compromise. : P
Let me see what I can do with the texture overlay then

TumeniNodes:
I only have one more question to ask.
Do the rails still connect to other rails? (in a straight line) to offer the option of doing rails longer than just one node between two fences?

paramat:
Rails will connect to a tree yes, i'm not too keen on that so you could remove 'group:tree' from 'connects to' if you want. I think it was only done in fences for connection to logs used in builds, it's not important.

TumeniNodes:
I'm just working on the textures now.
tough one, only due to there not being a post running vert. through the center.
trying to just break even between the tops and sides of the rails looking as nice as poss.

If fences were set as connect_to for logs, I think we best leave it be, before we start another revolution.
I am ok with the way this turned out.

paramat:
> If fences were set as connect_to for logs, I think we best leave it be

Yes i agree.

TumeniNodes:
are world aligned textures flexible, or a fixed feature?

paramat:
Do you mean 'stable'? If so yes.

TumeniNodes:
sorry, I mean, can it be turned off per node, or is it  fixed to every texture globally

paramat:
Surely the textures needed are just the standard wood plank textures plus an overlay?

TumeniNodes:
also, the 'overlay' is for the wield and inv items, not the texture of the rail, itself.

paramat:
screenshot_20180703_000457

It's fine using the standard plank texture, the texture is even continuous across the node border, just as it is when you have 2 planks side by side. The vertical discontinuity you see along the rail isn't actually where the node border is.
It's no worse than having 2 planks side by side.
Please use the plank textures for rails for simplicity and avoiding 5 more textures.

TumeniNodes:
ok. I have no problem doing that , saves me some time :)
I was unsure the look was acceptable or not.
I do not like it much, due to the fact, when placing them while facing North or South, you will get this
screenshot_20180702_203124

When placing them, while facing East or West, they look fine.

So, if you are ok with this, I am fine with it.
If later you, or anyone else change their mind, I will gladly work on the textures again.

TumeniNodes:
idk, I'm pretty happy with the way it turned out.
I had a different idea at the start and was skeptical of adding the same/similar connect_to functions as the fence nodes but, after playing around with it, it's not so bad.

I'm not sure what you mean by the last part. Do you mean 'end to end'? Because even using 2 rails between fence posts looks pretty good

Napiophelios:
just seems kinda unrealistic if you place too many 
without an actual fence post between them.
I mean Paramat's screen grabs show potential for interesting builds,
they don't look 'structurally sound' so to speak.

TumeniNodes:
I myself, would only use 1 between 2 fence posts but, who know what other user's preferences will be.
But during experimenting, I found there are situations where 2 between posts doesn't look too horrific.
My goal was to add an option to simply having fence post after fence post, and I never liked trying to place a fence between 2 stone columns, etc.

Like many other aspects of the base game, over time, someone else will come along and make more (and hopefully better) adjustments.
Or possibly find another method which might have been overlooked here

The on_rotate function, honestly would seem to be overkill, just because all you have to do is place a rail from another direction.

Napiophelios:
>The on_rotate function, honestly would seem to be overkill, just because all you have to do is place a rail from another direction.

It was meant as a possible solution to the 'problem' Ezhh mentioned.

TumeniNodes:
> It was meant as a possible solution to the 'problem' Ezhh mentioned.
ahhhh, sorry, I did'nt realize that. : )

mods/default/functions.lua
```diff
@@ -327,6 +327,71 @@ function default.register_fence(name, def)
...
+		groups = {},
+	}
+	for k, v in pairs(default_fields) do
+		if def[k] == nil then
+			def[k] = v
+		end
+	end
+
+	-- Always add to the fence group, even if no group provided
+	def.groups.fence = 1
```
paramat: Oops this was my mistakre, these 2 lines could be the 'fence_rail' group. However, i doubt we need a 'fence_rail' group, rails can just be in the 'fence' group since they are parts of a fence.

mods/default/functions.lua
```diff
@@ -303,7 +303,7 @@ function default.register_fence(name, def)
 			connect_right = {{1/8,3/16,-1/16,1/2,5/16,1/16},
 				{1/8,-5/16,-1/16,1/2,-3/16,1/16}},
 		},
-		connects_to = {"group:fence", "group:wood", "group:tree"},
+		connects_to = {"group:fence", "group:fence_rail", "group:wood", "group:tree"},
```
paramat: No need for a 'fence rail' group (i think).

mods/walls/init.lua
```diff
@@ -14,7 +14,7 @@ walls.register = function(wall_name, wall_desc, wall_texture, wall_mat, wall_sou
 			connect_back = {{-3/16, -1/2,  1/4,  3/16, 3/8,  1/2}},
 			connect_right = {{ 1/4, -1/2, -3/16,  1/2, 3/8,  3/16}},
 		},
-		connects_to = { "group:wall", "group:stone" },
+		connects_to = { "group:wall", "group:stone", "group:fence_rail" },
```
paramat: group:fence

mods/default/nodes.lua
```diff
@@ -2630,13 +2638,75 @@ default.register_fence("default:fence_pine_wood", {
...
+	wield_image = "default_fence_rail_overlay.png^default_pine_wood.png^" ..
+				"default_fence_rail_overlay.png^[makealpha:255,126,126",
+	material = "default:pine_wood",
+	groups = {choppy = 3, oddly_breakable_by_hand = 2, flammable = 3},
+	sounds = default.node_sound_wood_defaults()
+})
+
+default.register_fence_rail("default:fence_rail_aspen_wood", {
+	description = "Aspen Wood Fence Rail",
+	texture = "default_fence_rail_aspen_wood.png",
```
paramat: default aspen wood

mods/default/crafting.lua
```diff
@@ -1015,6 +1015,37 @@ minetest.register_craft({
...
+minetest.register_craft({
+	type = "fuel",
+	recipe = "default:fence_rail_aspen_wood",
+	burntime = 3,
+})
+
+minetest.register_craft({
+	type = "fuel",
+	recipe = "default:fence_rail_pine_wood",
+	burntime = 3,
```
paramat: 4 for a nice progression

paramat:
> I do not like it much, due to the fact, when placing them while facing North or South, you will get this

I see. That appearence is ok for me.

A few minor issues but PR is near-perfect.

TumeniNodes:
All points addressed. Thank you

paramat:
Will test then probably +1.

paramat:
In testing noticed walls do not connect to fences, you need to add "group:wall" to the 'connects to' in the 'fence registration helper'. +1 with that.

TumeniNodes:
done

paramat:
:+1:

paramat:
Ok i now see what you mean about the top textures changing from linear to lateral along a rail, sorry i misunderstood. So dedicated fence_rail textures would be good, here's what i suggest:

Copy the central horizontal rectangle (2px high 16 px wide), rotate it 90 degrees and paste to be a central vertical rectangle.
Where this vertical bar meets the 2 horizontal rectangles that texture the sides of the rails, 'blend' the junctions so that both look fairly continuous.
I won't remove my +1 though if you prefer to do this later.

TumeniNodes:
yes, I can definitely work on this.
It's actually going to be a slightly tricky texture to do.
I had it close at one point but, I think I managed to not save it. I'll poke at it a bit today to see what I can come up with.

I would love for this to get other reviews as well, when people have time of course. I realize everyone is busy with other things.

Ezhh:
I really do like this, but have to agree the top of that texture needs fixing.

Ezhh:
Hmm.. this bit looks odd. Too much of the same/similar colours grouped together?
fency

TumeniNodes:
There is an old saying for installers to customers, in the hard wood flooring industry.... "It's _wood_ , it does what wood does.

TumeniNodes:
better?

screenshot_20180706_170123

screenshot_20180706_170333

TumeniNodes:
this is with fence gate using the new _default_fence_rail_wood_ texture
screenshot_20180706_170821

Ezhh:
Yes, though the colour is still quite different from the gates (all light, and missing the dark colouring).

TumeniNodes:
It is those 4 very dark strips of color, for the "wood" texture of each wood type, which creates an oddity in the nodes.
That is the only color I have removed during any of this. It is intended to separate the wooden planks from one another in the wood textures.
With that in mind, a fence, fence gate, and fence rails..., would not be made from wood strips meant for flooring, or siding

TumeniNodes:
ok...even better. And I think all will like this.
This is with _all_ the fence nodes using the new _fence_rail_ texture.
screenshot_20180706_171931

and the new fence_rail texture, could simply be renamed to "fence" to replace the old fence texture, and now no need to add a bunch of new textures
... which makes it easier (or same as it was before) for TP creators

paramat:
Looks very wrong, sorry, very washed out and lacking in contrast. We like how fences look because they look like the wood plank textures, and they should. The dark pixels are what make the wood textures good looking.
Also, in one direction the top of the rails now has detail running linearly across it instead of all directions having lateral detail as the existing fences do. So the actual issue is not fixed at all.

I explained how this could be done above, please could you follow that precisely? It''s simple and straightforward. In fact the 'blending' stage i mention isn't even necessary.
I'm very tempted to do this myself, i'll do it for wood to show what i mean and show screenshots.

TumeniNodes:
No matter how you do it, when you add that dark line..., it will show in the other direction as a constant dark pixel (stripped) down along the rail

TumeniNodes:
is this better for you?
screenshot_20180706_184112

paramat:
Actually my instructions were wrong, sorry, because the dark line that runs along the fence rails in one direction has to be preserved (so that's not better, sorry), so the initial copied rectangle has to be offset not central.
Those dark spots, if unavoidable, look much better than removing all dark pixels, they can be considered knots in the wood.

Let's merge this because i can see you won't be happy doing what i request :)

TumeniNodes:
I did do what you requested. But it looked bad
This is using the two central linear stretches of pixels, and rotating 90* to the ummm, right?, and then slight blending pixel adjustments

TumeniNodes:
screenshot_20180706_192204

paramat:
From the other side, nowhere are there 2 dark pixels next to each other:

screenshot_20180707_003023

TumeniNodes:
default_fence_rail_wood
sorry, not sure how to do an enlargement of the texture

paramat:
No problem, looks like you had a similar approach. But the rail tops are not consistent with the existing fence rail tops. Also looks like you have added some new shades of pixels instead of keeping within the existing palette, so it looks weird. Sorry to be critical.

Please can i fix up these textures later and we merge this PR as it is now?

TumeniNodes:
There is no need to be sorry, uniformity is understandable and crucial. And being critical is good, when it's not done in a hostile manner so, I am always open to constructive criticisms

> Please can i fix up these textures later and we merge this PR as it is now?

YES PLEASE! haha

PR #2173, opened 2018-07-09
add evergreen bush

TumeniNodes:
adds an evergreen bush to the game,
fills in decoration in certain biomes a bit more.
If there are 'bush' and 'acacia bush', this only makes sense.

not so sure about the seed I chose for mapgen though


paramat:
Screenshot please.

paramat:
A little while ago i almost replaced the generic bush with a mugo pine bush #1841
I prefer specific species to a generic 'evergreen' so perhaps this can be called a 'mugo pine bush' or even just 'pine bush'?

TumeniNodes:
screenshot_20180709_145621
screenshot_20180709_150053

The seed needs to be adjusted, they are quite rare at '4'

Funny, I was set on calling them 'pine_bush", but figured someone would want a different name

TumeniNodes:
ahhh, now I remember the idea.
I did not follow it into #1841 though : /
kinda wish I had now

paramat:
The noise seed doesn't affect average density.

Taiga currently only has 1 decoration so this might be a good addition for that biome. Which then makes me think perhaps the bushes in 'snowy grassland' should be pine bushes instead of the generic bush? Something to consider.

In taiga it might be good for bushes to be densest where trees are thinnest, so perhaps use these noise params?:
```
		noise_params = {
			offset = -0.004,
			scale = -0.01,
			spread = {x = 250, y = 250, z = 250},
			seed = 2,
			octaves = 3,
			persist = 0.66
		},
```
That is, use the same spread, seed, octaves, persist but use a negative scale.

Note with offset -0.004 and scale +/-0.01 bushes will cover 33% of land area. With offset = 0 that bcomes 50% coverage.

TumeniNodes:
sounds good.

I actually think, pine_bush and even scarce amounts of dead bush would mix well together in snowy grassland?

TumeniNodes:
I should have a brief list of changes to make

paramat:
The schematic is the same structure as the 'bush'? That seems ok.

If you google image search 'wild mugo pine bush' it seems these like alpine / tundra conditions so would be suitable for snowy grassland. Mugo pine is also known as 'mountain pine'.

paramat:
I'm not keen on also adding a dead bush, let's keep it simple.

TumeniNodes:
well, pick a name haha
pine_bush
mugo_pine_bush (seems a bit much but, longer names do exist in the code I guess)
mountain_pine_bush (that's even longer but, sounds goof but, will not only be on mountains)

TumeniNodes:
I think we should just go with pine_bush.
There is already just 'bush" if a name change for both to make them specific, another PR can be done for that... 'name-hang' tends to curse development round here ; )

paramat:
Let's get some more input, but my initial thought is use the short and simple 'pine bush' for node name and either 'pine bush' or 'mugo pine bush' in description.

TumeniNodes:
ok. Ill add the noise param changes now.
I also swapped out bush, with the current pine bush for snowy grassland so we can see how it looks

TumeniNodes:
they still seem a bit scarce

paramat:
Adding a pine bush does indeed make sense, then we have a bush type for each temperature: acacia for hot (savanna), bush for temperate (deciduous forest and grassland) and pine for cold (taiga and snowy grassland).

So i propose we use the new pine bush in snowy grassland instead of the generic bush.
Because of that, the noise params of pine bush will actually have to match that of current 'bush' because the distribution of 'bush' in snowy grassland is carefully tuned. Sorry to request another noise param change, you had it right the first time.

And i'm sure now we should name the nodes 'default:pine_bush_...' to keep the names short. We can decide on the description later.

One thing i'm unsure about is pine bushes in temperate coniferous forest, we already have 3 decoration types there. However i'll try it and see.

TumeniNodes:
ok.
let me know. that way I can make the changes all at once.

TumeniNodes:
the 'bush' in grasslands, could possibly be spread out a bit more (as in size), would make them a little more dramatic and effective in such large, open spaces?

paramat:
Maybe, but a separate PR.

rubenwardy:
Have we made sure to purchase rights from @4Evergreen4?

TumeniNodes:
heh.
ok, renaming to pine_bush, and will pull them out of coniferous and floatland coniferous
probably ready in about an hour or so

There is no @pine_bush, is there? (checking all member lists)

TumeniNodes:
one quick question.
Should I just use default:pine_needles
or default:pine_bush_needles?

It uses the pine needles from pine trees but, it is possible a different texture will be desired in the future. So I feel sticking with pine_bush_needles is good, and continue using the current pine_needle.png for now

TumeniNodes:
nvm, I'm just gonna continue using pine_bush_needles.
If it has to change it can be done later

paramat:
It has to be a separate needles node to drop the correct sapling.

TumeniNodes:
already are. I kept it as 'pine_bush_needles'

paramat:
Ok, but usually the person credited is the last person to alter a texture, plus a 'derived from' comment. But whatever.

mods/default/mapgen.lua
```diff
@@ -1986,6 +1986,28 @@ function default.register_decorations()
 		flags = "place_center_x, place_center_z",
 	})
 
+	-- Pine bush
+
+	minetest.register_decoration({
+		name = "default:pine_bush",
+		deco_type = "schematic",
+		place_on = {"default:dirt_with_snow"},
+        sidelen = 16,
```
paramat: Weird indent

TumeniNodes: sorry, that is weird.. it was an indent of spaces rather than tabs.
fixed now

TumeniNodes:
just rechecked everything, found one tiny bit I missed.
fixed in last commit

paramat:
At first look code seems fine, will test.

mods/default/crafting.lua
```diff
@@ -49,6 +49,13 @@ minetest.register_craft({
 	}
 })
 
+minetest.register_craft({
+	output = 'default:pine_wood',
+	recipe = {
+		{'default:pine_bush_stem'},
+	}
+})
```
paramat: Please use " instead of ' in this registration, we're slowly converting them.

mods/default/mapgen.lua
```diff
@@ -1986,6 +1986,28 @@ function default.register_decorations()
...
+	minetest.register_decoration({
+		name = "default:pine_bush",
+		deco_type = "schematic",
+		place_on = {"default:dirt_with_snow"},
+		sidelen = 16,
+		noise_params = {
+			offset = -0.004,
+			scale = 0.01,
+			spread = {x = 100, y = 100, z = 100},
+			seed = 157,
```
paramat: Seed needs to match that of 'bush' to not change bush distribution in snowy grassland in existing worlds.

paramat:
Pine bush stem is RGBA, needs to be 'indexed alpha'.

Could you post the schematic lua table here?

paramat:
The schematic has the pine tree needles nodes, othewise fine.

TumeniNodes:
dang, I thought I did changes those, please hold

TumeniNodes:
I updated the previous schematic lua table that I posted

paramat:
screenshot_20180711_212141

Discovered this issue that is not your error. When bush saplings grow, some needles will not appear as they are not 'force-place' so will not replace snow. So in the schematic please could you make all needles nodes force-place? All else seems fine.

mods/default/mapgen.lua
```diff
@@ -1986,6 +1986,28 @@ function default.register_decorations()
...
+			offset = -0.004,
+			scale = 0.01,
+			spread = {x = 100, y = 100, z = 100},
+			seed = 137,
+			octaves = 3,
+			persist = 0.7,
+		},
+		biomes = {"taiga", "snowy_grassland"},
+		y_max = 31000,
+		y_min = 1,
```
paramat: 4

paramat:
Docs need updating at https://github.com/minetest/minetest_game/blob/d4b0b73ae004d83dd1ef8643efbedce532d08b05/game_api.txt#L916

paramat:
Tested. All fine now :+1:

TumeniNodes:
Thank you

PR #2180, opened 2018-07-21
Add blueberry bushes

random-geek:
unknown 3

In response to #2043 and also [https://forum.minetest.net/viewtopic.php?f=5&t=20439&p=326067#p326067](url).

This adds huckleberry bushes which spawn in forest biomes (yes, screenshot is in the wrong biome). Digging the leaves drops huckleberries which can be eaten with a 0.1 chance of poisoning, or crafted into 2 violet dye.
Some of the spawn/drop values may need tweaking, but I think I'm pretty happy with the textures and schematic.


tenplus1:
Hyckleberries are a good idea, but why are the fruits themselves placable ? and why 0.1 chance of poisoning player when they aren't poisonois ?

random-geek:
There is no reason for them to be place-able, other than for use as a decoration (which I find appealing). If that is a problem, of course I can change it back to a craftitem. The occasional poisoning was suggested by BlueTangsRock, but upon further consideration, the concept may not work if we specify any one variety of plant. The code is probably pretty messy anyways.

I also debated over whether to name nodes "huckleberry_bush_leaves", etc. or simply "huckleberry_leaves". If anyone has an opinion, please tell me.

Ezhh:
I'll do a proper review later, but I'm also not liking the poison chance. Can it just be kept as a food?

random-geek:
I removed the poisoning, and also removed some irrelevant code I forgot about after copying the node registration from apples. haha

random-geek:
Thanks for the input. I will try to make some changes soon.

random-geek:
I'm considering renaming `default:huckleberry_bush_leaves` and `default:huckleberry_bush_sapling` to `default:huckleberry_leaves` and `default:huckleberry_sapling`. This would eliminate the need for such a long name as `default:huckleberry_bush_leaves_plain`, which is inconvenient. Would this work? @Ezhh

Ezhh:
I think I favour consistency over shortening the length of the name, but let's see if any of the others have an opinion on this as well.

mods/default/README.txt
```diff
@@ -230,6 +230,11 @@ CloudyProton (CC BY-SA 3.0):
 Mossmanikin (CC BY-SA 3.0):
   default_fern_*.png
 
+random-geek (CC BY-SA 3.0):
+  default_huckleberries.png
+  default_huckleberry_bush_stem.png, derived from default_bush_stem (by paramat)
+  default_huckleberry_bush_leaves.png, derived from default_bush_leaves (by paramat)
+  default_huckleberry_bush_sapling.png, derived from default_bush_sapling (by paramat)
```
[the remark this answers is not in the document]
paramat: Yes different please, in fact all saplings should look different: something to work on.

mods/default/mapgen.lua
```diff
@@ -1964,6 +1964,28 @@ function default.register_decorations()
...
+	minetest.register_decoration({
+		name = "default:huckberry_bush",
+		deco_type = "schematic",
+		place_on = {"default:dirt_with_grass", "default:dirt_with_snow", "default:dirt_with_coniferous_litter"},
+		sidelen = 16,
+		noise_params = {
+			offset = -0.004,
+			scale = 0.005,
+			spread = {x = 100, y = 100, z = 100},
+			seed = 137,
```
paramat: Please use a different noise seed to all other bushes, for a unique distribution.

mods/default/nodes.lua
```diff
@@ -1613,6 +1617,96 @@ minetest.register_node("default:bush_sapling", {
 	end,
 })
 
+minetest.register_node("default:huckleberry_bush_stem", {
+	description = "Huckleberry Bush Stem",
```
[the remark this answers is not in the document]
paramat: Yes best use the generic stem.

mods/default/nodes.lua
```diff
@@ -1613,6 +1617,96 @@ minetest.register_node("default:bush_sapling", {
...
+	drawtype = "allfaces_optional",
+	waving = 1,
+	tiles = {"default_huckleberry_bush_leaves.png"},
+	paramtype = "light",
+	groups = {snappy = 3, flammable = 2, leaves = 1},
+	drop = {
+		max_items = 1,
+		items = {
+			{items = {"default:huckleberry_bush_sapling"}, rarity = 5},
+			{items = {"default:huckleberries"}}
```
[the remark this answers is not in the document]
paramat: I agree, otherwise you have to destroy the leaf nodes to get fruit, resulting in destroyed bushes with only the stem left.

mods/default/nodes.lua
```diff
@@ -1613,6 +1617,96 @@ minetest.register_node("default:bush_sapling", {
...
+			{x = -1, y = 0, z = -1},
+			{x = 1, y = 1, z = 1},
+			-- maximum interval of interior volume check
+			2)
+
+		return itemstack
+	end,
+})
+
+minetest.register_node("default:huckleberries", {
```
paramat: Apples are only nodes because that is how they exist in the world, picked berries should probably be items.

mods/default/nodes.lua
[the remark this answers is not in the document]
paramat: Is the stem recoloured? If it needs to be a different colour to the generic stem to be realistic then a dedicated stem is fine.

mods/dye/init.lua
```diff
@@ -73,6 +73,14 @@ minetest.register_craft({
 	recipe = {"group:coal"},
 })
 
+-- Manually add huckleberries->violet dye
+
+minetest.register_craft({
+	type = "shapeless",
+	output = "dye:violet 2",
```
paramat: If a flower gives 4 dye i think a bunch of berries certainly can give 4.

paramat:
If this is added i feel the biomes need changing.
Since this is a food source and there are already apples in deciduous forest i'm not keen on having this in deciduous forest. Also, we already have a bush in deciduous forest and i really don't want more as that biome is already dense and very rich in flora.

This is better in other biomes to make wild food avaialble in other locations. I see that the species likes tundra type climates so it could perhaps be in snowy grassland. A new bush is best added where flora is lacking in variety, so maybe in grasslands, if that is realistic.

I think this has potential with some changes and improvements.
Please can you post the schematic lua table in a comment so we can see it?

mods/default/nodes.lua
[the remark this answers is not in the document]
Ezhh: It's recoloured, but unlike all the other bush stems doesn't have any wood to be crafted into. Bush stems themselves are not great decorative items either, so it effectively makes this node pointless.

Maybe now the spawn mod sets default starting biomes we don't need bush stems to give wood, and could change them to giving sticks? That would give this new bush stem a crafting progression and seems more realistic (planks of wood from bush stems is quite silly after all).

Ezhh:
About the schematic - it currently feels smaller than other bushes / maybe not dense enough. However we need to balance this against it's use as a food source. Too many leaf nodes for berry collection could make gathering food via bushes much more easy than via any other means. (Maybe we need to consider very slow regrowth of berries? I don't want to make it so not all leaf nodes on the bush grow them, because the texture is nice and it would spoil the look.)

mods/dye/init.lua
Ezhh: This will actually mean it's 4 * number of leaves on the bush, and they regrow much easier than flowers. I think 2 is better.

mods/default/nodes.lua
```diff
@@ -1613,6 +1617,96 @@ minetest.register_node("default:bush_sapling", {
...
+	sunlight_propagates = true,
+	groups = {choppy = 2, oddly_breakable_by_hand = 1, flammable = 2},
+	sounds = default.node_sound_wood_defaults(),
+	selection_box = {
+		type = "fixed",
+		fixed = {-7 / 16, -0.5, -7 / 16, 7 / 16, 0.5, 7 / 16},
+	},
+})
+
+minetest.register_node("default:huckleberry_bush_leaves", {
```
paramat: This needs to be named and described as having berries.
`"default:huckleberry_bush_leaves_with_berries"`?
Then we'll need a leaf node without berries as the nodetimer marker for berry regrowth.

mods/default/nodes.lua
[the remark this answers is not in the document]
paramat: Yeah i'm ok with using the generic stem for this, fairly neutral.

mods/default/mapgen.lua
paramat: Also, scale = 0.01 like the other bushes may be better to make them more common.

mods/dye/init.lua
paramat: I'm ok with either.

paramat:
This needs the berry regrowth by nodetimer feature of the apples.
Also needs crafting recipe for stem into sticks or applewood planks, and fuel definitions for stem and sapling.

Good point about excesive food resource, appletrees have an average of 1 per tree, but then appletrees are much more common than bushes. We could tune the noise to balance this.

mods/default/nodes.lua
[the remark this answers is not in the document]
Ezhh: I'm happy to switch either to generic or use the new stem with the understanding we'll move to crafting stems to sticks (that could be a later PR though, and by one of us, so that it doesn't delay this).

Ezhh:
We need to keep in mind more than how common they are on mapgen as well; the ease of collecting saplings and growing them also matters. It might be worth reducing how much they heal so that apples are  better per use?

@random-geek I'm aware we've thrown a lot of feedback / suggestions / requests at you here. If there is anything you need help with, please let us know.

random-geek:
Most of these changes are fairly simple. However, the regrowth is proving to be harder than I thought, since on_timer is overridden by leafdecay.

tenplus1:
Personally I think leafdecay should be disabled for bushes and one stem used for them all which can be crafted into 4x sticks instead of wood.

Ezhh:
@paramat Why do bush leaves use leafdecay? They are at or just above ground level, so it seems completely pointless (no floating cloud of leaves above players' heads problem).

random-geek:
Whoops, something went very wrong with that last commit. Sorry. :(

paramat:
Leafdecay for bushes. It was requested by someone because players were digging the stem to get wood and leaving the leaves (messy players). However i'm ok with removing leafdecay for bushes.

Concerning stems dropping 1 wood, this is still needed because the spawn mod can place the player in grasslands so wood plank resources are still needed in grasslands. I know it is a little unrealistic but i'd prefer to keep it that way. The bush stems are pretty meaty pieces of wood after all, see the texture. I don't want to reduce the variety of starting biomes as this causes issues for the spawn mod.

random-geek:
@paramat , Ezhh pointed out huckleberries don't have a tree parallel, so there is not much need for having `bush` in the nodenames. Would you agree? Either option is fine, but I much prefer the shorter name.

Assuming I disable leafdecay, the remaining suggested changes should be applied fairly soon. Also, I would like feedback on the new sapling texture.

paramat:
I prefer keeping 'bush' in the nodenames for consistency with other bushes and for descriptiveness.

Ezhh:
New sapling texture looks good to me.

paramat:
Looks good to me too.

Found why leafdecay was added #1531 i had been watching 'let's play minetest' videos and noticed a player digging the stem for wood and leaving the leaves. But with huckleberry we have a good reason to not have leafdecay.

Ezhh:
Agreed - and since bush leaves are on the ground or just above, I really don't think it's a bad thing to leave them without the stem.

@random-geek You can drop leafdecayfor huckleberry bushes.

random-geek:
I've made most of the requested changes, and left the stem as a separate node. The berry gathering code will need to be reviewed. Since there was some discussion of the schematic, here is the code I used with saveschems:
```
local L = {name = "default:huckleberry_bush_leaves_with_berries", prob = 255}
local M = {name = "default:huckleberry_bush_leaves_with_berries", prob = 95}
local S = {name = "default:huckleberry_bush_stem", prob = 255, force_place = true}

mts_save("huckleberry_bush", {
	size = {x = 3, y = 3, z = 3},
	data = {
		_, _, _,
		M, L, M,
		_, M, _,

		_, _, _,
		L, S, L,
		M, L, M,

		_, _, _,
		M, L, M,
		_, M, _,
	},
})
```
This bush is intentionally smaller than the regular bushes, as it looked somewhat strange as a larger decoration.

Napiophelios:
If leaf decay was added to bushes; why even have bush leaves
I mean why not just use default apple/acacia leaf nodes?
are they different somehow?

paramat:
So that the leaf nodes drop bush saplings instead of tree saplings.

paramat:
In the schematic, all the leaves should force-place, otherwise they won't replace surface snow in snowy biomes.

paramat:
Also, we can now use 'place_y_offset = 1' and remove the layer of air nodes from the schematic.

random-geek:
@paramat , where is `place_y_offset`? Is it in the schematic, or in register_decoration?

paramat:
Like this:
https://github.com/minetest/minetest_game/blob/ab4940505e0138d2900332663b461da555165a64/mods/default/mapgen.lua#L1771
It places the decoration 1 node higher so that the lowest layer of unplaced air nodes is no longer needed in the schematic.

paramat:
https://github.com/minetest/minetest/blob/d7d451c647f1fff7983178a9817888f4d0bab695/doc/lua_api.txt#L6114

random-geek:
@paramat Won't enabling `force_place` cause trees to be destroyed?

paramat:
Oops actually i remembered that snow is dropped after decorations are placed, and you are right the leaves would then replace tree trunks etc.
Sorry, no force-place then.
This means the pine bush will have to be re-done (but all bushes need redoing anyway to use 'place_offset_y' anyway).

random-geek:
Done. 👍

paramat:
Could you post the schematic table again so i can archive it?
Before this is merged i need to think about what biomes to put these in.

random-geek:
@paramat , here is the code:

``` lua
local L = {name = "default:huckleberry_bush_leaves_with_berries", prob = 255}
local M = {name = "default:huckleberry_bush_leaves_with_berries", prob = 95}
local S = {name = "default:huckleberry_bush_stem", prob = 255, force_place = true}

mts_save("huckleberry_bush", {
	size = {x = 3, y = 2, z = 3},
	data = {
		M, L, M,
		_, M, _,

		L, S, L,
		M, L, M,

		M, L, M,
		_, M, _,
	},
})
```

paramat:
Why is there param2 usage on the stem and leaves with berries?
How about on digging the fruit version place normal leaves and start the timer?
Maybe you're making sure that fruit regrowth depends on a stem being present nearby? (but that doesn't require a param2 check).

Ezhh:
Approach I'd take so there shouldn't be any need for param2 checks on leaves:

If a player digs leaves they should only get leaves without berries. Placement won't start a timer. Mapgen and grown from sapling bushes will start as leaves with berries. Then a node timer started when those berries are harvested and replaced by leaves without berries. Outside creative mode (where it doesn't matter anyway) players will never have the _with_berries node in their inventory.

I've also experimented with collecting by digging instead of right click and it does simplify. 

Leaves without berries just need:
```
on_timer = function(pos, elapsed)
	minetest.set_node(pos, {name = "default:leaves_with_berries"})
end,
```

Leaves with berries just need:
```
after_dig_node = function(pos, oldnode, oldmetadata, digger)
	minetest.set_node(pos, {name = "default:leaves"})
	minetest.get_node_timer(pos):start(math.random(300, 1500)) 
end,
```

This is obviously a bit simplified as light level checks and so on are still needed, but it neatly removes any issues around param2 and protection checks no longer need to be manually added. 

How does this sound?

(I'm aware I suggested right-click as an approach to begin with, so please accept my apologies for that).

Ezhh:
Another small point that came out of my own experiments:

Do we want to consider changing default_huckleberry_bush_leaves_with_berries.png so it's only berries, not leaves, then do `default_huckleberry_bush_leaves.png^default_huckleberry_bush_leaves_with_berries.png` (though the name of the berries texture can then be simplified to not mention leaves)? This makes it easier to edit the leaves texture, and for texture packs and so on.

paramat:
Maybe i'm missing something but a fabricated bush is allowed to grow berries? The nodes would only be obtained by digging a bush so it's just like moving the bush or growing one from a sapling.

Ezhh:
@paramat It's because by collecting berries with right click it was possible to collect the leaves while they still have berries and then place them in that form, at which point you could right click and start the timer.

random-geek:
@Ezhh 
Are you suggesting that digging the leaves with berries would replace the node with plain leaves? That would mean players would have to dig each node _twice_ if they wanted to remove it, which seems excessively complicated to me.

Ezhh:
I'm not sure why anyone would find it complicated. You dig apples separately from leaves. You dig grass separately from dirt_with_grass. To collect, you dig. Right click would be very inconsistent and players may not realise they can collect the berries at all. 

Let's wait on paramat's thoughts anyway.

paramat:
> That would mean players would have to dig each node twice if they wanted to remove it,

This seems fine to me, it's a special case. The leaves with berries could have 'dig immediate = 3' to make it fast since that first dig only collects the fruit.

Ezhh:
Leaves with berries should also have:
`node_dig_prediction = "default:huckleberry_bush_leaves"`

random-geek:
Should I keep the param2 check for the stem? It seems somewhat unnecessary.

Ezhh:
I don't see a need for it.

paramat:
I don't see a need for it either.

paramat:
Comments on textures:

The leaves textures are fine and good. I'm not keen on the suggestion to use texture layering to add a berry layer to leaves, as that doesn't give as much artistic freedom and doesn't reduce the number of textures needed.

The pile of huckleberries texture and the stem texture both need converting to indexed colour.

Ezhh:
> ... as that doesn't give as much artistic freedom

It gives more freedom in fact, because you can separately change the leaves without needing to change two textures, and you can change the berries without the leaves getting in the way.

tenplus1:
This is where the shears tool would have come in handy, digging normally would only return the berries but shears would have been required to pick up the empty bush node.

random-geek:
Resolved conflicts.

paramat:
I'm ok with using layered textures, separating the leaves from the berries, as suggested https://github.com/minetest/minetest_game/pull/2180#issuecomment-410505657
But note this might make the berries appear to be on top of the leaves, not in them.

Ezhh:
> But note this might make the berries appear to be on top of the leaves, not in them.

I have tested this and it appears fine.

rubenwardy:
Please make this a separate mod

Ezhh:
> Please make this a separate mod

I know default has too much content, but when all other bushes are in default this wouldn't make sense.

paramat:
I'm a little unsure about the need for this too.
However, if it is added to MTG i have requests for changes:

* Use a different berry name, 'huckleberry' is very local to America.
Wikipedia: "The name 'huckleberry' is a North American variation of the English dialectal name variously called 'hurtleberry' or 'whortleberry' (/ˈhwɜːrtəlbɛri/) for the bilberry"
Billberry: https://en.wikipedia.org/wiki/Vaccinium_myrtillus The common worldwide name for this is 'blueberry' which is descriptive, intuitive and more well known, it also sounds more edible. Most people would have no idea if a Huckleberry is edible.
So 'Blueberry' is better.

* Rethink biomes.
We already have bushes in 'deciduous forest' and 'taiga', and i actually don't like placing bushes in forests and prefer coniferous forest without bushes. So not in any forest.
In grassland is good and suitable for 'bilberrys' (they like heathland), so these could be in 'grassland' and 'snowy grassland' only alongside the existing bushes.

Also:
The noise spreads need to be 100 to match existing bushes.

Ezhh:
To add to the above:
- textures should be adjusted as described in earlier comments.
- the berry collection behaviour still needs to be adjusted.

In worse case, I'm willing to adjust these things myself once everything else is done.

random-geek:
@Ezhh How do you want the berry collection behavior adjusted? It has already been adjusted a couple of times. :)

Ezhh:
@paramat I am massively disappointed you let this sit for two months before blocking it and will say outright that this is a poor way to treat contributors. If you do not like a PR based on the concept, you should say so quickly, not two months, a rather large number of comments, and multiple commits later. Nothing substantial changed in the concept of the PR, so there is no excuse to progress so far only to then block it.

tenplus1:
@paramat - minetest needs some new food items added and this is the perfect chance to do so, plus it has many uses.

rubenwardy:
My suggestion to make this a separate mod is because I'm hesitant to add more stuff to default, and this seems to be a new craft tree as there aren't any berries yet. If you add more items such as berry pies, then that would make much more sense to be in a new mod

There are currently two different foods in the game: apples which are forested and quick but low health, and bread which is farmed and high health. So it would make sense to have berries as an intermediary maybe - a forrested food with higher health than apples when made into a pie. Minetest Game doesn't have hunger though, which makes food less useful, but these bushes are pretty anyway

> The reason i have reviewed this is so that the author learns how to code for MTG, it wasn't because i was going to approve it, i have been uncertain the whole time. I do feel bad about this, the PR has been quite hard work with lots of revisions and delays in review.
👎 unfortunately, and i apologise for that 

You should have stated this to begin with, by disapproving the PR before offering critism

paramat:
>  If you do not like a PR based on the concept, you should say so quickly,

> You should have stated this to begin with, by disapproving the PR before offering critism

I couldn't disapprove before reviewing, because i was uncertain the whole time
See https://github.com/minetest/minetest_game/issues/2043#issuecomment-364605416
several months ago i wrote "But anyway, i'm still undecided on this, and low priority."
I did not decide to be against this from the beginning, please read what i wrote:
> i have been uncertain the whole time.

So you have misunderstood.

Also, it may be supported by other core devs, so as mapgen dev i made sure it would be correct if it is merged, that's why i am reviewing it and will continue to.
It's unfortunate but is honesty, it is not intentionally treating contributors poorly. We're allowed to change our minds, to suppress your own true opinion and do what is wrong for MTG for sake of avoiding disappointing the author is much worse.

Rubenwardy, ignoring the PR for 2 months then stating your disapproval is also unfortunate, but of course it's right to be honest. So you have done something similar to me, except, you didn't even review :) However i'm not going to suggest you have acted unacceptably, because you haven't.

Remember 1 disapproval can't block a PR.

rubenwardy:
I haven't disapproved this PR, and I actually like it

paramat:
I see, sorry.

DS-Minetest:
I don't like the name. Wikipedia says: "Huckleberry is a name used in North America for several plants"[¹](https://en.wikipedia.org/wiki/Huckleberry). So, this name is only used in America and not even only for one plant but for several. This is really unclear. (Also think of how this should be translated correctly, if you don't just write the english name, which would be like not translating it, it will be wrong.)
What about (common) bilberry or blue whortleberry?

If huckleberries are Vaccinium myrtillus, shouldn't they be much lower on the ground? I think, I've never seen a Vaccinium myrtillus that is taller than 1/2 meter. [Also see this wikimedia commons picture.](https://commons.wikimedia.org/wiki/Category:Vaccinium_myrtillus_(plant)?uselang=de#/media/File:%D0%97%D0%B0%D0%BA%D0%B0%D0%B7%D0%BD%D0%B8%D0%BA_%22%D0%A7%D0%BE%D1%80%D0%BD%D0%B8%D0%B9_%D0%BC%D0%BE%D1%85%22_-_%D1%82%D0%B8%D0%BF%D0%BE%D0%B2%D0%B8%D0%B9_%D0%B2%D0%B8%D0%B3%D0%BB%D1%8F%D0%B4_2.jpg)

paramat:
Looks like rubenwardy will support this so development can continue. I will continue to review mapgen aspects, my requests are https://github.com/minetest/minetest_game/pull/2180#issuecomment-424496013 A name change would be good, sfan5 first suggested 'blueberry'.

random-geek:
Thanks for the feedback. If I understand correctly, these are the desired changes (assuming there is interest in this PR):

- Rename to bilberry/whortleberry (I prefer bilberry for its simpler name)
- Split texture into two layers
- Make textures indexed (I don't remember if I already did this or not)
- Adjust berry collection behavior according to Ezhh's suggestions
- Move bushes back to grassland/snowy grassland

Ezhh:
There is interest.

The list sounds good to me, but I'd like @paramat to confirm his thoughts on name choice and biomes.

paramat:
I prefer Blueberry, which is closely related to Bilberry:
> Vaccinium myrtillus is a species of shrub with edible fruit of blue color, commonly called "bilberry", "wimberry", "whortleberry", or European blueberry.[3] It has much in common with the American blueberry (Vaccinium cyanococcus).

My 3 requests are here https://github.com/minetest/minetest_game/pull/2180#issuecomment-424496013

rubenwardy:
> Rename to bilberry/whortleberry (I prefer bilberry for its simpler name)

I've never heard of bilberry. I'd personally prefer blueberry. It has the benefit of being intuitive even if not called that in the user's culture (similar to "blue wool")

random-geek:
Hopefully that fixes all the issues mentioned.
Note: I changed the berries item texture to have more variation, and flipped it so the lighting would be from the top-left. The stem and leaves textures are not indexed, because the originals were not indexed either.

tenplus1:
Now I have concerns, while Huckleberry bushes can actually grow as big as the schematic, 
Blueberry bushes cannot and rarely reach 1 meter in width/height (1 block)... Why the name change ?? -1 from me

Huckleberry: http://pabook2.libraries.psu.edu/palitmap/HuckleberryColony.jpg
Blueberry: https://www.quickcrop.ie/site/uploads/Plant%20DB%20Photos/blueberry/blueberry-bush.gif

DS-Minetest:
@tenplus1 In your images the blueberry bush looks much bigger to me.

But I agree that the bush is too big, Vaccinium myrtillus should rather be one node high plantlike bushes.
> If huckleberries are Vaccinium myrtillus, shouldn't they be much lower on the ground? I think, I've never seen a Vaccinium myrtillus that is taller than 1/2 meter. [Also see this wikimedia commons picture.](https://commons.wikimedia.org/wiki/Category:Vaccinium_myrtillus_(plant)?uselang=de#/media/File:%D0%97%D0%B0%D0%BA%D0%B0%D0%B7%D0%BD%D0%B8%D0%BA_%22%D0%A7%D0%BE%D1%80%D0%BD%D0%B8%D0%B9_%D0%BC%D0%BE%D1%85%22_-_%D1%82%D0%B8%D0%BF%D0%BE%D0%B2%D0%B8%D0%B9_%D0%B2%D0%B8%D0%B3%D0%BB%D1%8F%D0%B4_2.jpg)

random-geek:
I had thought about the height issue at various times, though I'm afraid I plantlike node could look strange in this context. Another possibility would be to leave it as a leaf-like node, but change the schematic to be shorter and with no stem. I am open to either option, but obviously we will have to let the devs decide.

Ezhh:
This is difficult for me to comment on, since I really didn't mind the original name. I'd prefer a better berry type than making any major change to the bushes.

paramat:
With some research it seems this plant is low and spreading, certainly no higher than 1m.
If we remove the stem node this would fix the stem issues (what wood should it drop). The schematic could then be flat and 3x3 nodes? This would also give it a distinctive shape different to the other bushes.

Also i'm removing my -1, i don't feel strongly enough to officially disapprove, more neutral.

Ezhh:
I'd be fine with paramat's suggestion.

random-geek:
image

Some concept bushes I just constructed. Does that look good?

paramat:
That's what i had in mind. I guess with centre node always placed, face-connected high prob, diagonal slightly lower prob?

random-geek:
image

Naturally spawned bushes using new schematic.

Ezhh:
Screenshot looks good. Is everything ready for a (hopefully final) review now?

paramat:
I haven't tested but the PR looks fine now.
I might fine tune the noise params in a follow up PR, this will be much easier than insisting this is done in this PR, and i don't want to delay this PR any more.
Now that we have 2 types of bush in grasslands both may need lowering in density.

random-geek:
The only thing I'm concerned about is the colour of the leaves, since it is quite close to the colour of the grass. I don't know if that's a problem, though. I could try tuning the values a bit.

paramat:
Looks ok to me. However that can always be tuned in a later PR.

paramat:
I forgot to confirm, the leaves textures do indeed have to be RGBA and not 'indexed alpha', because that is required by the 'allfaces optional' drawtype.

paramat:
So i'm unofficially +1 for the PR in terms of the changes. But this still needs 2 concept approvals from other core devs. Perhaps rubenwardy can provide the 2nd concept approval?
I'm happier with this now that these are low bushes instead of another 3x3x2 bush.

mods/default/nodes.lua
```diff
@@ -1614,6 +1617,90 @@ minetest.register_node("default:bush_sapling", {
...
+			{items = {"default:blueberry_bush_sapling"}, rarity = 5},
+			{items = {"default:blueberry_bush_leaves"}}
+		}
+	},
+	sounds = default.node_sound_leaves_defaults(),
+
+	on_timer = function(pos, elapsed)
+		local nodes = minetest.find_nodes_in_area(vector.add(pos, {x = -1, y = -1, z = -1}),
+			vector.add(pos, {x = 1, y = -1, z = 1}),
+			{"default:dirt_with_grass", "default:dirt_with_snow"})[1]
```
Ezhh: This means if grass is removed, berries stop growing, then if grass grows back, they still don't grow, and that could be confusing for players. There's also no real-world parallel here. Berries don't only grow if there's grass nearby. Can we remove this check?

paramat: If you decide to have a check here it should be for 'soil' group.

Ezhh: I considered suggesting that, but I think it's better to avoid the check.

random-geek: I considered making the leaf nodes attached like grass, but then some bushes which had grown on hills would be "unstable". I vote for a soil group check, because it seems very strange that players could have blueberry plants producing blueberries without the presence of any dirt whatsoever.

paramat: I agree with removing the check. A soil node could be temporarily removed from under and the timer trigger during that time, this would create leaves that never grow berries again.
Removing a soil node from under doesn't kill the leaves, so growing berries in that situation seems ok.
Also a small advantage in avoiding node searches.

random-geek:
Should I add this bush to `schematic_tables.txt`?

paramat:
@rubenwardy @SmallJoker hopefully someone can provide the other +1?

paramat:
Combining rubenwardy's concept approval with my technical approval to create the 2nd :+1:

tacotexmex:
Why put this in default? Why not take the chance to put bushes in its own mod instead?

Ezhh:
Because all other bushes are already in default. This includes the nodes and the functions and the decoration registrations. They would all need to be moved, and then we'd need to add aliases for them. 

Default is obviously annoyingly large, but unless we separate large chunks of it into distinct mods (which I think has had resistance before, and would need even more aliases...) it doesn't make any sense to have a mod for a single bush.

tacotexmex:
That reasoning is sound. At the same time it is often asserted that new MTG features needs to be added sparsely and with restraint because of the many servers and games built upon it. This addition is just a bit disparate with that in mind (and harder to disable because of its part of default).

Ezhh:
You also can't individually disable aspen trees, coral or kelp.

For what it's worth - I'm strongly in favour of lots of smaller mods as opposed to the huge thing that default has become. It's just very hard to change it once it gets this way.

tacotexmex:
I understand. MarkuBu did it though, I believe: https://github.com/MarkuBu/minetest_modbase

Ezhh:
That's a separate game based on MTG, not something people using MTG can switch to. Switching to that after using MTG first would leave you with a whole world full of unknown nodes because it doesn't do anything to account for node name differences, not to mention unnatural shifts between biomes the next time you generate never-before-visited map, since it uses a completely different biome system.

MTG can't do a radical change like this because it needs to maintain compatibility with existing worlds. This is both a strength (people don't need a new world every time there's an update), and a problem (because it prevents us easily fixing some things we'd love to fix).

If you want to continue this discussion, please open an issue for it. It's a much larger topic than where a new bush should exist within the structure of MTG, and if the use of default for this bush is a real problem for you, you will probably need to argue a case for much wider changes.

rubenwardy:
That separate game has a terrible design that I argue is worse than default

PR #2187, opened 2018-07-28
Amend empty drop from "" to {}

tenplus1:
Changes empty drops from returning empty string to empty table.

### DIFF
--- a/mods/doors/init.lua
+++ b/mods/doors/init.lua
@@ -86,7 +86,7 @@ minetest.register_node("doors:hidden", {
 	diggable = false,
 	buildable_to = false,
 	floodable = false,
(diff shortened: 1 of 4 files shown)


SmallJoker:
What's the difference of these two values?

tenplus1:
@SmallJoker - for some reason get_node_drops returned tnt:tnt_burning as a drop item instead of an empty drop, this fixed it.

SmallJoker:
@tenplus1 Then it's the wrong place to fix this issue. Please create a core PR for this bug.

tenplus1:
@SmallJoker - https://github.com/minetest/minetest/pull/7592

PR #2226, opened 2018-10-01
add glass and obsidian glass slabs

TumeniNodes:
We have slabs for everything else, so I feel these make sense.
I did not add stairs, as they never seem to look right (though these 2 new textures may change that, I will test it out)


TumeniNodes:
screenshot_20181001_154932

screenshot_20181001_154948

mods/stairs/init.lua
```diff
@@ -804,3 +804,22 @@ stairs.register_stair_and_slab(
 	"Snow Block Slab",
 	default.node_sound_snow_defaults()
 )
+stairs.register_stair_and_slab(
+	"glass",
+	"default:glass",
+	{cracky = 3, oddly_breakable_by_hand = 3},
+	{"default_glass.png", "default_glass.png", "stairs_glass_slab.png"},
+    "Glass Stair",
```
paramat: Indentation

mods/stairs/init.lua
```diff
@@ -804,3 +804,22 @@ stairs.register_stair_and_slab(
...
+	"Glass Slab",
+	default.node_sound_glass_defaults()
+)
+
+stairs.register_stair_and_slab(
+	"obsidian_glass",
+	"default:obsidian_glass",
+	{cracky = 3},
+	{"default_obsidian_glass.png", "default_obsidian_glass.png", "stairs_obsidian_glass_slab.png"},
+    "Obsidian Glass Stair",
```
paramat: Indentation

mods/stairs/init.lua
TumeniNodes: yeh, i forgot when using text editor I have to use 2 tabs, will fix

paramat:
I tested trying out normal glass textures for 'inner stair', looked even weirder than the PR approach, so nevermind that idea.

TumeniNodes:
I told you :D

paramat:
Since the parent nodes are not rotatable these will have world-aligned textures, so a dedicated side texture will look odd for an upright slab and other rotated stair nodes. Something to consider.
However once we decide on the 'world aligned textures' issue these could possibly be non-WA to avoid that issue.

screenshot_20181002_011014

screenshot_20181002_011034

Odd how these textures behave even though WA.

paramat:
However stairs are usually not-rotated, and the dedicated side textures look better and may well be worth still using. The alternative is using the normal glass textures, which may overall look worse.

TumeniNodes:
hm, the side textures should not be doing that. This is with WA set?
The reason I made the side textures, is because the slabs look horrible using the existing ones.

paramat:
Yes if you specify non-WA in the tile definitions that will override the auto-WA normally applied to all slabs/stairs.

TumeniNodes:
oh, so it _can_ be set to the tiles in the def without touching the defs of the node then?

Ezhh:
> Odd how these textures behave even though WA.

The behaviour in the screenshot is exactly what I'd expect from WA textures. Glass slabs should definitely not be world-aligned. 

> This oddness may be an engine bug worth reporting.

I don't see any engine bug here. This is expected behaviour (once again, WA textures should not have been merged.)

paramat:
The PR provides divided side textures, if WA was working then the side of a vertically rotated slab would be divided, which doesn't happen in https://github.com/minetest/minetest_game/pull/2226#issuecomment-426104467
> Glass slabs should definitely not be world-aligned.

Even though the parent nodes are not rotatable?

Tumeni best that #2219 is merged before this.

TumeniNodes:
Makes sense too me

TumeniNodes:
I suggest, may be a nice gesture, to have a mod, ready to go for people to use after this is reverted. Would make many happy.

Ezhh:
>  if WA was working then the side of a vertically rotated slab would be divided

Was looking at the wrong part - so an issue with WA when different textures are used for the sides/top?

> Even though the parent nodes are not rotatable?

Yes, if we're using different textures for the sides - we need to make sure those are maintained correctly on rotation.

>  may be a nice gesture, to have a mod, ready to go

Agreed. I don't mind doing this if no one else jumps in first.

paramat:
> Yes, if we're using different textures for the sides - we need to make sure those are maintained correctly on rotation.

Fine for me.

paramat:
Will need updating now that #2219 is merged.

TumeniNodes:
I will do it tomorrow, which is today... but later on

TumeniNodes:
dammit, 
I have to close this and make a new PR. I screwed my rebase
now the branch is 4 commits ahead and 1 behind, ugh

paramat:
No problem i still do this myself sometimes because it's easier.

PR #2229, opened 2018-10-04
add glass and obsidian glass stairs

TumeniNodes:
replacement PR for 'add glass and obsidian glass slabs'

screenshot_20181004_162851


mods/stairs/init.lua
```diff
@@ -855,3 +855,23 @@ stairs.register_stair_and_slab(
...
+
+stairs.register_stair_and_slab(
+	"glass",
+	"default:glass",
+	{cracky = 3, oddly_breakable_by_hand = 3},
+	{"default_glass.png", "default_glass.png", "stairs_glass_slab.png"},
+	"Glass Stair",
+	"Glass Slab",
+	default.node_sound_glass_defaults()
+)
```
paramat: Missing bool for WA.

mods/stairs/init.lua
```diff
@@ -855,3 +855,23 @@ stairs.register_stair_and_slab(
...
+
+stairs.register_stair_and_slab(
+	"obsidian_glass",
+	"default:obsidian_glass",
+	{cracky = 3},
+	{"default_obsidian_glass.png", "default_obsidian_glass.png", "stairs_obsidian_glass_slab.png"},
+	"Obsidian Glass Stair",
+	"Obsidian Glass Slab",
+	default.node_sound_glass_defaults()
+)
```
paramat: Missing bool for WA.

mods/stairs/init.lua
paramat: https://github.com/minetest/minetest_game/pull/2226#issuecomment-426112817
So 'false'.

paramat:
I think the textures should be renamed as they are not just used for slabs.
'... glass_divided' or something similar?

mods/stairs/init.lua
TumeniNodes: Well, that was silly of me, duh
fixed for both

TumeniNodes:
Is anything else needed, as far as fixes?

paramat:
All looks fine, i just need to test.

Ezhh:
It needs changes.

When two slabs combine, we're getting the liquid tank effect. 
glassslabs1

The stairs nodes look like abstract art, with lines missing where they should be, and lines where they shouldn't be. Here's some examples:

Inner stair from the side:
glassslabs2

Stairs from the top:
stairstop

Stairs from the side:
stairsside

In summary: slabs are great if the liquid level issue can be fixed, but stairs would need a full range of textures to look good.

TumeniNodes:
Good catch @Ezhh ,
I did not want to add the stairs originally, just the slabs.

I have no idea what is causing this glitch.
I first thought, it might be related to not using glass_detail.png on the slabs but, that should not have anything to do with it.

Is this a shader issue, related to connected glass?
I don't see how this should even be happening, as far as the textures, or the node defs, and swap are concerned from slabs to full block.

Ezhh:
@TumeniNodes Not sure about the slab issue yet. But I'm looking at how to fix stairs.

This will give normal stairs as good as you can get them:
```
stairs.register_stair(
	"obsidian_glass",
	"default:obsidian_glass",
	{cracky = 3},
	{"stairs_obsidian_glass_split.png", "default_obsidian_glass.png", 
        "stairs_obsidian_glass_stairside.png^[transformFX",
	"stairs_obsidian_glass_stairside.png", "default_obsidian_glass.png", "stairs_obsidian_glass_split.png"},
	"Obsidian Glass Stair",
	"Obsidian Glass Slab",
	default.node_sound_glass_defaults(),
	false
)
```

Where default_obsidian_glass.png is:
stairs_obsidian_glass_stairside
(the extra bit in the corner will be for reuse in the other stair types, to minimise total textures needed)

TumeniNodes:
I'll try that out in a few.
Agreed much better to use the modifiers, than to create additional textures

Ezhh:
And for inner stairs:
(no need of further textures)

```
stairs.register_stair_inner(
	"obsidian_glass",
	"default:obsidian_glass",
	{cracky = 3},
	{"stairs_obsidian_glass_stairside.png^[transformR270", "default_obsidian_glass.png", 
        "stairs_obsidian_glass_stairside.png^[transformFX", "stairs_obsidian_glass_stairside.png", 
        "default_obsidian_glass.png", "stairs_obsidian_glass_stairside.png"},
	"Obsidian Glass Slab",
	default.node_sound_glass_defaults(),
	false
)
```

Ezhh:
Outer stairs:
```
stairs.register_stair_outer(
	"obsidian_glass",
	"default:obsidian_glass",
	{cracky = 3},
	{"stairs_obsidian_glass_stairside.png^[transformR90", "default_obsidian_glass.png",
	"stairs_obsidian_glass_outer_stairside.png", "stairs_obsidian_glass_stairside.png",
	"stairs_obsidian_glass_stairside.png^[transformR90", "stairs_obsidian_glass_outer_stairside.png"},
	"Obsidian Glass Slab",
	default.node_sound_glass_defaults(),
	false
)
```

Where stairs_obsidian_glass_outer_stairside.png is:
stairs_obsidian_glass_outer_stairside

NOTE: These still leave some internally visible textures; however, without adding too many textures, I think it's as good as we can get for the stairs, and it looks pretty good to me.

Ezhh:
Actually, this would also need to account for rotation, which it;s not handling correctly at the moment... but I'm out of time for playing round with it right now.

paramat:
2 slabs combine into a full glass cube https://github.com/minetest/minetest_game/blob/5b3b60927000706f3613423d30e2ed108e2c38fd/mods/stairs/init.lua#L214 which is 'glasslike optional' and therefore contains liquid according to the often non-zero param2. For glass param2 has to be set to zero in this situation.
Probably best to detect if the full cubic node is 'facedir' and set param2 to zero if not.

Ezhh:
Okay, so I had a little more time...

The rotation issue I'm getting is specific to the top texture for normal stairs. Compare:
stairstop1

With:
stairstop2

TumeniNodes:
Should just remove to code which combines slabs into a full block... it's dumb : P 
Kinda joking, kinda not..

Ezhh:
I've never liked it. If I wanted to place a full block, I'd place a full block... but that's offtopic for here :)

paramat:
^ I would agree to that.

TumeniNodes:
> I would approve removing slab combining code.

That is a blessing to hear.
Can anyone else chime in on that matter?

If plenty consent, I can either do so within this PR or, do one right after to do the removal.
But, would it be wise to find out how some server owners feel about it?

I don't see how it could break any existing worlds, as wherever slabs were combined are now full glass nodes. (?)

paramat:
It won't break anything.

Ezhh:
Agreed with @paramat, and it would have my instant support. 

For this PR, if we will keep stairs, and it's agreed the inner textures are acceptable, we need to:
- update to use the textures and registrations I provided.
- figure out why the upper texture in normal stairs insists on being world-aligned, even though the others rotate correctly.
- fix the liquid tanks issue (though if we agree in advance on removing slab combining, that will fix itself, but not sure the best way to approach this - we shouldn't merge one PR when it has a known error that will only be fixed in a later PR which others devs might still block).

@paramat Any thoughts on how to approach the last point?

TumeniNodes:
> fix the liquid tanks issue

Will be fixed right after with a follow up PR to remove slabs combo code

TumeniNodes:
ok, I have the obsidian glass stairs all fixed. They actually look nice, from all angles, through all rotations
screenshot_20181006_180101

screenshot_20181006_180107

screenshot_20181006_180053

screenshot_20181006_180043

screenshot_20181006_180037

sorry for all the shots but, only way to show atm

paramat:
Those look good.

The liquid tank issue: I'll quickly cook up the fixing code and test it for inclusion in this PR, it's simple so not a big issue if it's removed again later.

paramat:
Liquid tank fix (tested):
Code to add is:
```
					-- If paramtype2 of cubic version is not 'facedir' set its
					-- param2 to 0 to avoid bugs such as liquid in connected
					-- framed glass.
					if minetest.registered_nodes[recipeitem].paramtype2 ~= "facedir" then
						p2 = 0
					end
```
Added like this:
```
				-- combine two slabs if possible
				if slab_trans_dir[math.floor(p2 / 4)] == dir
						and wield_item == under.name then

					if not recipeitem then
						return itemstack
					end
					if minetest.is_protected(pointed_thing.under, player_name) and not
							minetest.check_player_privs(player_name, "protection_bypass") then
						minetest.record_protection_violation(pointed_thing.under,
							player_name)
						return
					end
					-- If paramtype2 of cubic version is not 'facedir' set its
					-- param2 to 0 to avoid bugs such as liquid in connected
					-- framed glass.
					if minetest.registered_nodes[recipeitem].paramtype2 ~= "facedir" then
						p2 = 0
					end
					minetest.set_node(pointed_thing.under, {name = recipeitem, param2 = p2})
					if not creative_enabled then
						itemstack:take_item()
					end
					return itemstack
				end
```

Fixer-007:
Those blocks are rotatable, right?

TumeniNodes:
@Fixer-007 yes they is.

Ezhh:
Stairs sides look like this for me:
stairssidewewweewe
There's a split in the middle that should not be there on both sides.

mods/stairs/README.txt
```diff
@@ -7,3 +7,18 @@ Authors of source code
...
+Authors of media (textures)
+---------------------------
+
+Textures
+--------
+
+Copyright (c) 2017 Shara RedCat (CC BY-SA 3.0)
+ stairs_obsidian_glass_outer_stairside.png
+ stairs_obsidian_glass_stairside.png
+
```
Ezhh: 2018, and they are just as derived as your own textures.

TumeniNodes: I will fix

mods/stairs/init.lua
```diff
@@ -855,3 +861,60 @@ stairs.register_stair_and_slab(
...
+	default.node_sound_glass_defaults(),
+	false
+)
+
+stairs.register_stair(
+	"obsidian_glass",
+	"default:obsidian_glass",
+	{cracky = 3},
+	{"stairs_obsidian_glass_split.png", "default_obsidian_glass.png",
+	"stairs_obsidian_glass_outer_stairside.png", "stairs_obsidian_glass_outer_stairside.png",
```
Ezhh: Should be: `"stairs_obsidian_glass_stairside.png^[transformFX", "stairs_obsidian_glass_stairside.png",`

paanrama:
Why obsidian glass is transparent?

TumeniNodes:
> Why obsidian glass is transparent?

That is how it is. I may be putting up a PR to change it but, textures with alpha transparency often bring out glitches in the engine.

paanrama:
I ask because I live in a volcanic country and I have a lot of obsidian at home. Obsidian is a no transparent volcanic glass. It can have some different colors but in my country is generally pure black an never transparent, so for me is strange see a transparent obsidian. In your place exists transparent obsidian? I never listen about it.

JurajVajda:
I guess only the frame is obsidian and the glass is glass :)

mods/stairs/README.txt
```diff
@@ -7,3 +7,26 @@ Authors of source code
...
 Various Minetest developers and contributors (LGPLv2.1+)
+
+Authors of media (textures)
+---------------------------
+
+Textures
+--------
+
+Copyright (c) 2018 Shara RedCat (CC BY-SA 3.0)
+
```
paramat: No empty line

mods/stairs/README.txt
```diff
@@ -7,3 +7,26 @@ Authors of source code
...
+ stairs_obsidian_glass_outer_stairside.png -- Derived from a texture by PilzAdam (CC BY-SA 3.0)
+
```
paramat: As above

mods/stairs/README.txt
```diff
@@ -7,3 +7,26 @@ Authors of source code
...
+ stairs_obsidian_glass_stairside.png -- Derived from a texture by PilzAdam (CC BY-SA 3.0)
+
+
+Copyright (c) 2018 TumeniNodes (CC BY-SA 3.0):
+
```
paramat: As above

mods/stairs/README.txt
```diff
@@ -7,3 +7,26 @@ Authors of source code
...
+ stairs_glass_outer_stairside.png -- Derived from a texture by celeron55 (CC BY-SA 3.0)
+
```
paramat: As above

mods/stairs/README.txt
```diff
@@ -7,3 +7,26 @@ Authors of source code
...
+ stairs_glass_stairside.png -- Derived from a texture by celeron55 (CC BY-SA 3.0)
+
```
paramat: As above

mods/stairs/README.txt
```diff
@@ -7,3 +7,26 @@ Authors of source code
...
+ default_glass_split -- Derived from a texture by celeron55 (CC BY-SA 3.0)
+
```
paramat: As above

mods/stairs/init.lua
```diff
@@ -855,3 +861,97 @@ stairs.register_stair_and_slab(
 	default.node_sound_snow_defaults(),
 	true
 )
+
+	--glass stair nodes need to be registered individually to utilize specialized textures.
```
paramat: No indent. Space after '--', capital letter

paramat:
Code looks ok once my line comments are attended to, will now test.

paramat:
Tested. You've both done an excellent job with the textures, these look really good.
:+1:

paramat:
When this is merged, please can the commit message explain the bugfix for slabs combining into a cube: that facedir is only set for the cube if it has paramtype2 = "facedir".

TumeniNodes:
> Tested. You've both done an excellent job with the textures, these look really good.
> +1

Thank you. And thank you for testing and for the approval.
Yes, definitely add the explanation

Ezhh:
@paramat I'll leave merging for you because I'm not entirely sure what you want the commit message to be.

paramat:
Ok, i'll merge this within a day.

PR #2232, opened 2018-10-08
Connect fence rails to walls

TumeniNodes:
This was part of the original PR, but it was removed.
This adds more building options

screenshot_20181008_150159

### DIFF
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -368,7 +368,7 @@ function default.register_fence_rail(name, def)
 				{1/16,  3/16, -1/16, 1/2,  5/16, 1/16},
 				{1/16, -5/16, -1/16, 1/2, -3/16, 1/16}},
 		},
-		connects_to = {"group:fence"},
+		connects_to = {"group:fence", "group:wall"},
 		inventory_image = fence_rail_texture,
 		wield_image = fence_rail_texture,
 		tiles = {def.texture},


benrob0329:
While on the subject, this looks very wrong:
screenshot_20181008_151537

TumeniNodes:
actually, you can makes some pretty cool stuff with that
screenshot_20181001_124834
screenshot_20181001_124435
screenshot_20181001_124404

paramat:
:-1: Fence rails will look bad when running parallel to walls again. Fence rails connecting to walls as in the screenshot looks bad (a wall post plus lower sections either side) and is of little use.
Fences can still connect to walls by placing a "fence" node there.

Ezhh:
But these are actual dedicated walls. Not full node walls like in the previous issue screenshot.

And why would someone build a connecting wall and a connecting fence side by side when they can just join together?

paramat:
See tumeninodes comments https://github.com/minetest/minetest_game/issues/2201#issuecomment-416114294 and the following 2 comments, including and https://github.com/minetest/minetest_game/issues/2201#issuecomment-416393509
So this inconsistency is a little irritating =s

paramat:
Also see https://github.com/minetest/minetest_game/issues/2201#issuecomment-416751198
> If there are no objections i'll make a PR in a while to only connect to fences.

Supported by both of you and 3 others.

paramat:
Ezhh good points in https://github.com/minetest/minetest_game/pull/2232#issuecomment-427971218 but i still think it's better for rails to only connect to fence posts.

Ezhh:
I agreed (and still do) with not connecting to solid nodes, but wasn't really fully sold on this part of the change even then. I actually noted at the time that I believed the additional connections had been intentional, and I was mostly thinking of walls when I said that.

It's certainly not a big deal to me either way, but I support this.

TumeniNodes:
I had misunderstood and mispoke in those comments, and apologize for that.
I had thought it was about just not connecting them to full nodes, for a number of reasons.

Connecting with walls and fences was the original intention but, my originals were also full width and used facedir.
I like that, that part was changed, to behave as fences do, but do not like the connect_to with full nodes

If the confusion I cause irritated you, again, I apologize

With that, I still feel the rails should connect to fences & walls

paramat:
Ok i see, no problem :]

paramat:
I find this is a much better looking way to connect walls and fences, which is possible now:

screenshot_20181009_041847

Concerning 'building options', it's also the case that omitting a connection creates building options.
As a random, abstract example, the first structure below is possible now and looks good, the screenshot after is with this PR and looks bad, it also creates rail T-junctions without posts.

screenshot_20181009_042008

screenshot_20181009_042143

On balance i find the benefit of not having this PR-added connection outweighs the benefit of having it. Structures generally look better when walls, fences and rails are close together. It creates more interesting and less messy structures.

Ezhh:
I find wooden connections between stone posts make sense, whereas in your first screenshot the distance between fence and wall is so short as to seem not worth bothering, and you are going to get the same connection issues you show in the final screenshot when you use all short fences - so I don't see how this approach can be better.

For the final screenshot, why have a wall and a fence next to each other like that? They are functionally the same thing, so assuming you have a walled and fenced area, it would make more sense to only have one division between them, not a fence and wall back to back.

TumeniNodes:
I don't think I have ever seen anyone put a fence, right along side a wall like that.
In my opinion, if the fence rails, are to behave the same as the fence, then they should also use 'connect_to" in the same manner, as far as 'walls' go

Again, this is just my own opinion. The decision rests with the dev team

paramat:
Yes i suspected my example would be criticised as something that is not often built :] That's why i described it as 'random, abstract'. It's just a random construction to illustrate the behaviour of these 3 nodes used together and that i find the possibilities of not having that connection more interesting and better looking than having it. I don't expect a wall and fence running parallel is common either but the construction only looks like that by accident, i could show many other constructions that illustrate what i mean that are not a wall and fence in parallel.

IRL if you had a fence near a wall it makes sense that the fence post may be reinforced by being connected to the wall, however you wouldn't connect the middle of every rail to a nearby wall.
You can still do this, which looks better:

screenshot_20181009_042514

Ezhh:
That looks bad to me, not better.

> i could show many other constructions that illustrate what i mean that are not a wall and fence in parallel

Please do. Examples no one is ever likely to build don't help things.

paramat:
I actually disagree with walls and fences connecting at all, they are seperate systems. But it may be too late now as changing this would break builds.

In my opinion walls and fences should only connect to the cubic version of the wall/fence material. However then someone wrote: but what if a wooden house has trunks at it's corners? So now fences connect to tree nodes, meaning you can't run a fence past a tree without an ugly connection.
This excess of connections was a mistake so i'm now disapproving any connection without a very good reason behind it, to not repeat these mistakes.
Connections also remove possibilities and add ugliness, they're not universally good as has been assumed in the past.

Ezhh:
> In my opinion walls and fences should only connect to the cubic version of the wall/fence material. 

I thought this was the thing that specifically triggered removing connections. I don't really think full nodes should connect to fences or walls anyway. No one is asking for this to be restored, so I'm not sure why you are arguing against it.

> Connections also remove possibilities and add ugliness

I disagree. It all depends on which connections. Walls and fences work well when connecting. Full nodes and walls/fences don't work well.

> they're not universally good as has been assumed in the past.

Not sure who assumed this.

TumeniNodes:
fences connecting to tree nodes drives me insane/er.

fences connecting to full/solid nodes is blasphemy (or it should be), and highly frowned upon.
_wall_ connecting to full nodes make sense
fences (and all their components) connecting to walls makes sense.

It's all logic

paramat:
Outvoted, go for it :]

TumeniNodes:
Thank you

PR #2237, opened 2018-10-11
add wall extension

TumeniNodes:
This adds a new nodebox to extend walls, when placed atop one another, to fill in the gap (as seen in the left of the image below)

ba

### DIFF
--- a/mods/walls/init.lua
+++ b/mods/walls/init.lua
@@ -13,10 +13,10 @@ walls.register = function(wall_name, wall_desc, wall_texture_table, wall_mat, wa
 			type = "connected",
 			fixed = {{-1/4, -1/2, -1/4, 1/4, 1/2, 1/4}},
 			-- connect_bottom =
-			connect_front = {{-3/16, -1/2, -1/2,  3/16, 3/8, -1/4}},
-			connect_left = {{-1/2, -1/2, -3/16, -1/4, 3/8,  3/16}},
-			connect_back = {{-3/16, -1/2,  1/4,  3/16, 3/8,  1/2}},
-			connect_right = {{ 1/4, -1/2, -3/16,  1/2, 3/8,  3/16}},
+			connect_front = {{-3/16, -5/8, -1/2,  3/16, 3/8, -1/4}},
+			connect_left = {{-1/2, -5/8, -3/16, -1/4, 3/8,  3/16}},
+			connect_back = {{-3/16, -5/8,  1/4,  3/16, 3/8,  1/2}},
+			connect_right = {{ 1/4, -5/8, -3/16,  1/2, 3/8,  3/16}},
 		},
 		connects_to = { "group:wall", "group:stone", "group:fence" },
 		paramtype = "light",
@@ -25,55 +25,6 @@ walls.register = function(wall_name, wall_desc, wall_texture_table, wall_mat, wa
 		walkable = true,
 		groups = { cracky = 3, wall = 1, stone = 2 },
 		sounds = wall_sounds,
-				after_place_node = function(pos, placer, itemstack, pointed_thing)
-			local pos_under = {x = pos.x, y = pos.y - 1, z = pos.z}
-			local pos_above = {x = pos.x, y = pos.y + 1, z = pos.z}
-			local node_under = string.gsub(minetest.get_node(pos_under).name, "_full$", "")
-			local node_above = string.gsub(minetest.get_node(pos_above).name, "_full$", "")
-
-			if minetest.get_item_group(node_under, "wall") == 1 then
-				minetest.set_node(pos_under, {name = node_under .. "_full"})
-			end
-			if minetest.get_item_group(node_above, "wall") == 1 then
-				minetest.set_node(pos, {name = wall_name .. "_full"})
-			end
-		end,
-		after_dig_node = function(pos, oldnode, oldmetadata, digger)
-			local pos_under = {x = pos.x, y = pos.y - 1, z = pos.z}
-			local node_under = string.gsub(minetest.get_node(pos_under).name, "_full$", "")
-			if minetest.get_item_group(node_under, "wall") == 1 and
-					digger and digger:is_player() then
-				minetest.set_node(pos_under, {name = node_under})
-			end
-		end,
-	})
-
-	minetest.register_node(wall_name .. "_full", {
-		description = wall_desc,
-		drawtype = "nodebox",
-		node_box = {
-			type = "connected",
-			fixed = {{-1/4, -1/2, -1/4, 1/4, 1/2, 1/4}},
-			connect_front = {{-3/16, -1/2, -1/2,  3/16, 1/2, -1/4}},
-			connect_left = {{-1/2, -1/2, -3/16, -1/4, 1/2,  3/16}},
-			connect_back = {{-3/16, -1/2,  1/4,  3/16, 1/2,  1/2}},
-			connect_right = {{ 1/4, -1/2, -3/16,  1/2, 1/2,  3/16}},
-		},
-		connects_to = {"group:wall", "group:stone", "group:fence"},
-		paramtype = "light",
-		is_ground_content = false,
-		tiles = wall_texture_table,
-		groups = {cracky = 3, wall = 1, stone = 2, not_in_creative_inventory = 1},
-		sounds = wall_sounds,
-		drop = wall_name,
-		after_dig_node = function(pos, oldnode, oldmetadata, digger)
-			local pos_under = {x = pos.x, y = pos.y - 1, z = pos.z}
-			local node_under = string.gsub(minetest.get_node(pos_under).name, "_full$", "")
-			if minetest.get_item_group(node_under, "wall") == 1 and
-					digger and digger:is_player() then
-				minetest.set_node(pos_under, {name = node_under})
-			end
-		end,
 	})
 
 	-- crafting recipe
@@ -95,5 +46,4 @@ walls.register("walls:mossycobble", "Mossy Cobblestone Wall", {"default_mossycob
 		"default:mossycobble", default.node_sound_stone_defaults())
 
 walls.register("walls:desertcobble", "Desert Cobblestone Wall", {"default_desert_cobble.png"},
-		"default:desert_cobble", default.node_sound_stone_defaults())
-
+"default:desert_cobble", default.node_sound_stone_defaults())
--- a/mods/walls/init.lua
+++ b/mods/walls/init.lua
@@ -25,6 +25,55 @@ walls.register = function(wall_name, wall_desc, wall_texture_table, wall_mat, wa
 		walkable = true,
 		groups = { cracky = 3, wall = 1, stone = 2 },
 		sounds = wall_sounds,
+				after_place_node = function(pos, placer, itemstack, pointed_thing)
+			local pos_under = {x = pos.x, y = pos.y - 1, z = pos.z}
+			local pos_above = {x = pos.x, y = pos.y + 1, z = pos.z}
+			local node_under = string.gsub(minetest.get_node(pos_under).name, "_full$", "")
+			local node_above = string.gsub(minetest.get_node(pos_above).name, "_full$", "")
+
+			if minetest.get_item_group(node_under, "wall") == 1 then
+				minetest.set_node(pos_under, {name = node_under .. "_full"})
+			end
+			if minetest.get_item_group(node_above, "wall") == 1 then
+				minetest.set_node(pos, {name = wall_name .. "_full"})
+			end
+		end,
+		after_dig_node = function(pos, oldnode, oldmetadata, digger)
+			local pos_under = {x = pos.x, y = pos.y - 1, z = pos.z}
+			local node_under = string.gsub(minetest.get_node(pos_under).name, "_full$", "")
+			if minetest.get_item_group(node_under, "wall") == 1 and
+					digger and digger:is_player() then
+				minetest.set_node(pos_under, {name = node_under})
+			end
+		end,
+	})
+
+	minetest.register_node(wall_name .. "_full", {
+		description = wall_desc,
+		drawtype = "nodebox",
+		node_box = {
+			type = "connected",
+			fixed = {{-1/4, -1/2, -1/4, 1/4, 1/2, 1/4}},
+			connect_front = {{-3/16, -1/2, -1/2,  3/16, 1/2, -1/4}},
+			connect_left = {{-1/2, -1/2, -3/16, -1/4, 1/2,  3/16}},
+			connect_back = {{-3/16, -1/2,  1/4,  3/16, 1/2,  1/2}},
+			connect_right = {{ 1/4, -1/2, -3/16,  1/2, 1/2,  3/16}},
+		},
+		connects_to = {"group:wall", "group:stone", "group:fence"},
+		paramtype = "light",
+		is_ground_content = false,
+		tiles = wall_texture_table,
+		groups = {cracky = 3, wall = 1, stone = 2, not_in_creative_inventory = 1},
+		sounds = wall_sounds,
+		drop = wall_name,
+		after_dig_node = function(pos, oldnode, oldmetadata, digger)
+			local pos_under = {x = pos.x, y = pos.y - 1, z = pos.z}
+			local node_under = string.gsub(minetest.get_node(pos_under).name, "_full$", "")
+			if minetest.get_item_group(node_under, "wall") == 1 and
+					digger and digger:is_player() then
+				minetest.set_node(pos_under, {name = node_under})
+			end
+		end,
 	})
 
 	-- crafting recipe
--- a/mods/walls/init.lua
+++ b/mods/walls/init.lua
@@ -13,10 +13,10 @@ walls.register = function(wall_name, wall_desc, wall_texture_table, wall_mat, wa
 			type = "connected",
 			fixed = {{-1/4, -1/2, -1/4, 1/4, 1/2, 1/4}},
 			-- connect_bottom =
-			connect_front = {{-3/16, -5/8, -1/2,  3/16, 3/8, -1/4}},
-			connect_left = {{-1/2, -5/8, -3/16, -1/4, 3/8,  3/16}},
-			connect_back = {{-3/16, -5/8,  1/4,  3/16, 3/8,  1/2}},
-			connect_right = {{ 1/4, -5/8, -3/16,  1/2, 3/8,  3/16}},
+			connect_front = {{-3/16, -1/2, -1/2,  3/16, 3/8, -1/4}},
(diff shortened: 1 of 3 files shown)


Ezhh:
Note: I've given permission for the code of mine tumeninodes has made use of to be released under LGPLv2.1+ for the purposes of this PR.

TumeniNodes:
Thank you very much Ezhh. When I saw your use of it in hedges, it just fit perfectly.

sofar:
Not sure that you need to add a new node to do this:

A simpler solution would be to make the "middle" wall segment a connected nodebox node that can connect *up*, and does so automatically. That keeps the placement code unchanged.

TumeniNodes:
Attempting to make the current node do this to itself/within itself, produces strange things.

connect_top and connect_bottom, do not operate as one would believe they should, they create oddities.
That is why I felt this method was a reasonable way to go

paramat:
Tumeni, what are the strange results?

I would prefer sofar's suggestion, if it works. However this might add the gap filling nodebox when creating a pillar from single wall nodes stacked vertically, therefore altering the appearence of some builds.

TumeniNodes:
I tried several ways to use the 'connect_top' and 'connect_bottom' to do what is suggested... yet it always wound up facing the top node into another direction. rotating it 90*

As I said, I will try messing with the connect_top and connect_bottom pos, and fidget with the code again, over the next couple days, and see if I can get it right.

paramat:
Ah i think i see, the upwards connection needs to fill the gap in all 4 directions, so would add filler boxes where not wanted.

TumeniNodes:
> Ah i think i see, the upwards connection needs to fill the gap in all 4 directions, so would add filler boxes where not wanted.

exactly

Ezhh:
> Ah i think i see, the upwards connection needs to fill the gap in all 4 directions, so would add filler boxes where not wanted.

This being why my hedges mod (the one tumeninodes has taken code from) works this way and adds a second node.  I spent a good bit of time trying to do what sofar suggested in that mod first, but it can't work.

paramat:
I don't find high walls important enough to add this more messy workaround.
I generally want to avoid adding low priority features to MTG because i feel we need to shift some effort elsewhere to new games, and avoid endlessly bloating MTG (which is old, messy and does so much wrong) with unnecessary stuff. Sorry :-1:

Ezhh:
@paramat There's really nothing messy about this.

Your current spate of blocking PRs just because you personally can't be bothered with or don't see the value in things is starting to put me off MTG development.

If you personally don't like something, don't block it - leave it for others to review.

paramat:
I can't block PRs alone.
> just because you personally can't be bothered with

This isn't one of my reasons.
> don't see the value in things

If a core dev doesn't see the value of something, as in, thinks it's wrong for the game, that is a valid reason to disapprove. I'm allowed to have an opinion and state it.
> If you personally don't like something, don't block it

That's unreasonable, if a core dev thinks something is wrong for the game it's best to be truthful and state their disapproval.
> leave it for others to review.

I'm not disapproving because i can't be bothered to review.

You're being unreasonable, not me.
However perhaps my comment was worded not so well. I do think this is wrong for the game, the first part "I don't find high walls important enough to add this more messy workaround." is my reason. The second part of my comment is more my attitude towards MTG and is not so much a reason.

TumeniNodes:
> I don't find high walls important enough to add this more messy workaround.
> 
> Sorry -1

But... without high walls... how are we gonna keep Amurica safe? :'(

It's ok, really.
I can simply pop it out as a mod release.
And then you will see... You'll _all_  see, and I will unleash the power of my walls upon your entire kingdom. You will rue this day on which you denied my greatness.

(a little fun to ease the tension) You mortals are quite interesting to observe.

paramat:
Don't worry :] I expect this will be approved.

Ezhh:
I'll review/test/almost certainly approve if any other dev will support it, but don't want to do a full review unless I know there is enough interest.

Ezhh:
Who decides nodes shouldn't be stacked? That's a very arbitrary decision. I dislike using real world arguments because we obviously can't match everything to how it is in reality, but walls can be any height, therefore it makes sense for a player to expect them to be stackable in game. 

When I first used the mod I loved it (still do), but that was quickly followed by disappointment when I realised I couldn't build them high enough to prevent people/mobs easily jumping over them. I know I'm not the only one who wished to stack them, and there's nothing about the concept that is wrong.

This is also not something that has a neat and easy fix in other mods, short of completely replacing the walls mod with yet another redo mod, so for me it makes a lot of sense to have this in MTG.

Napiophelios:
why don't you just alter the node box 'down',
you would only notice when mixing build materials
or digging underneath the wall
screenshot_20181013_232700

no one seems to care that the gates glitch into stuff
what would it matter here?

Ezhh:
You partially answered your own question. :)

But also consider that players might want walls with gaps in them to look through, or might want to build on top of glass, or see-through ice. Then the lower part becomes visible.

paramat:
Here's a random experiment, probably with a few issues, but sharing it anyway in case it leads to a better idea:

screenshot_20181014_060152

screenshot_20181014_060216

The connecting walls are now 1 node tall, the pillars are 9/8 nodes tall such that the pillars overlap when stacked, and no gap. No tile z-fighting on the overlap either.
However, when stacking it is necessary to point at the top of the connecting wall, if you point at the top of the pillar the stacked node ends up 1 node too high. This *might* be solvable with a defined selectionbox, not sure.
Anyway, super-simple change to a few numbers:
```
		node_box = {
			type = "connected",
			fixed = {{-1/4, -1/2, -1/4, 1/4, 5/8, 1/4}},
			-- connect_bottom =
			connect_front = {{-3/16, -1/2, -1/2,  3/16, 1/2, -1/4}},
			connect_left = {{-1/2, -1/2, -3/16, -1/4, 1/2,  3/16}},
			connect_back = {{-3/16, -1/2,  1/4,  3/16, 1/2,  1/2}},
			connect_right = {{ 1/4, -1/2, -3/16,  1/2, 1/2,  3/16}},
		},
```
I thnk Napiophelios' idea is excellent. Better than mine. Not a problem when mixing materials as it just looks like interlocking pieces. Visual intrusion into glass or ice is a minor issue, walls are not often placed in that way.

paramat:
Ezhh, when i wrote "Walls were never meant to be stacked" i meant the walls mod wasn't written with that intention, which is why i don't consider it a bugfix as you have suggested.

paramat:
screenshot_20181014_065057

Looks good with gaps, even more castle-like having the top of the gap undulating.

Napiophelios:
>The only reason I see this as a better method for MTG, is that it fits better visually.
But it looks kind of terrible on top of transparent nodes

Ezhh:
Sorry, but @Napiophelios suggestion is :-1:  for me. Extended into other nodeboxes is really not a good approach.

The original suggestion was neither "complex" nor "messy", despite your claims.

TumeniNodes:
> @TumeniNodes If I am right in my understanding of what you are saying, you prefer the original approach. If so, please switch it back.

I would rather switch it back.
I did find it odd that other times in the past, when I have come up with something which pushes into the next nodebox, it is viewed as taboo, but this time it was ok.

The original approach results in a cleaner visual.
After messing with it this way a bit, made me want to revert back to the original state of the PR
I will be gone for a bit today and when I return later tonight I will switch this back.

paramat:
Since we have a simple and effective alternative solution, i now am strongly opposed to the more complex method, and am now confident in my disapproval :] (don't mean this in an unpleasant way, i'm in a good mood).
Walls on top of glass is very rare, the benefits of simplicity outweigh the occasional visual intrusion, which is only 1/8th of a node.
It will be a mistake to start auto-swapping in and out unnecessary near-duplicate nodes into worlds when a beautifully simple solution exists. Once those extra nodes are added, there's no going back.

paramat:
However, if this PR goes ahead, i have a suggestion.

What i dislike the most about this, what i consider 'messy', is the 'full' wall being a dummy node `not_in_creative_inventory`, and being auto-swapped according to nearby building. Instead it seems better to simply give builders the 'full' wall node to use as they wish. It actually makes a very nice non-stacked wall alternative:

screenshot_20181015_025608

Stacked it can create a wall with a flat top if desired.
This way builders have more options, the code is simpler and there's no need for potentially problematic auto-swapping code.
If this PR goes ahaed, please can it use this approach?

TumeniNodes:
So, if I remove the 'not in inventory' def... will there need to be a recipe for these full wall nodes now?
If so, then someone will need to come up with a recipe, cuz I don't do 'recipes' not sure if it would just be 'wall_mat' in all 9 slots?

I have no idea how people keep coming up with recipes for so many things, when there are only 9 slots for crafting and so many types of nodes.

A stone slab is 3 stone across the bottom, the stone walls are 6 stone from bottom up/across...
I have yet to mess with recipes as I only use MTG in creative

Napiophelios:
do walls currently have a 'reverse' recipe to revert walls back to build material?
if not, you could place a crafted wall block into the grid and it would just convert it into the new type wall.

TumeniNodes:
> do walls currently have a 'reverse' recipe to revert walls back to build material?
> if not, you could place a crafted wall block into the grid and it would just convert it into the new type wall.

I don't believe they do..., that's a good idea. Well, seems good to _me_ anyway.

paramat:
Let's see what Ezhh thinks of my suggestion in https://github.com/minetest/minetest_game/pull/2237#issuecomment-429686537 . With this implementation i would remove my -1 and might even +1. It offers more possibilities than the downward-extended wall idea, which i know Ezhh will be opposed to.

TumeniNodes:
Yep, that's why I have not made that particular change as of yet

Ezhh:
The whole point of this to me was making it automatic. Manual placement of the second node removes a lot of value from this and I'm not keen to approve that approach, though won't oppose it as such either.

Automatic placement of the other nodes is neither complicated nor messy. Automatic connections is the whole point of having connecting nodes. When this is, despite your claims, simple to have, why not have it? I'm sorry, @paramat , but I just can't understand your issues with this.

For the original approach (subject to reviewing and testing it first) I'll be :+1:  and for the extended nodebox approach I will be :-1: 

For the moment, beyond this I'm removing myself from the conversation. Another dev will be needed here either way.

DS-Minetest:
Just make normal wall to full wall with shapeless:
`W`
And vice versa.

- Players would be annoyed having specific nodes that can't be crafted to anything in their inventory. They would see it as trash.
- As cobble is such a basic material, there are mods that add other building nodes crafted by cobble, don't block that rare shapes.

TumeniNodes:
Not sure which way I want to go with this now/yet.

There is more I planned to add, in steps (other PRs) related to walls but, now I am not sure if better to go this rout, which has caused some minor conflict or, to simply release it all as a mod?

Kinda torn here.
While it is nice to be able to contribute, it can be difficult at times when there are some changes suggested, which you do not feel so keen on.
But, both sides of this are understandable.

I'm gonna sit on this for a bit to decide. It is not a priority matter so, it can just rest for a bit, if that can be allowed, without closing it just yet.

I will add this though:
With the absence of a few true 'game'-like aspects to MTG, many people get excited to see new types of building/decorative nodes come along. Just something to consider with PRs such as this.

And please do not take that as an insult in any way... it is merely an observation.
I like MT/MTG and the community, and I like to contribute/share where I can

So, let's chill on this one for a little bit.

paramat:
Yes it's fine to pause this one.
Contributing is sometimes difficult, but disagreement and disapproval are inevitable, natural, healthy and essential apects of development =]
> With the absence of a few true 'game'-like aspects to MTG, many people get excited to see new types of building/decorative nodes come along. Just something to consider with PRs such as this.

That has a slight implication, in response i have to clarify that i don't on average oppose new nodes, as my history proves =] I'm not doing that here, my suggestion actually adds more new nodes to the PR. It's just implementation i'm concerned about.

If you have more to add to walls it might be helpful to open an issue to discuss your plans first. However it's also completely fine to not do so.
Your plans may be good ideas, just because there was some disagreement on one aspect doesn't mean the rest will be the same.

rubenwardy:
There's the following glitch, due to a limitation caused by only having 2 nodes for this and no good drawtype:

screenshot_2018-11-04_21-31-06

I don't think this is necessarily a problem, however

TumeniNodes:
That could probably be fixed with some fancy... 'if above but no node to the left or right, use these positions for connect_left/connect_right...' or something like that but... then it becomes complicated

mods/walls/init.lua
```diff
@@ -25,6 +25,55 @@ walls.register = function(wall_name, wall_desc, wall_texture_table, wall_mat, wa
 		walkable = true,
 		groups = { cracky = 3, wall = 1, stone = 2 },
 		sounds = wall_sounds,
+				after_place_node = function(pos, placer, itemstack, pointed_thing)
+			local pos_under = {x = pos.x, y = pos.y - 1, z = pos.z}
+			local pos_above = {x = pos.x, y = pos.y + 1, z = pos.z}
+			local node_under = string.gsub(minetest.get_node(pos_under).name, "_full$", "")
```
rubenwardy: I don't like the required use of ~regex~ patterns here. As the *walls* group is in the full version, you don't need the canonical version first.

I think this would be better:

```lua
local node_under = minetest.get_node(pos_under).name
if minetest.get_item_group(node_under, "wall") == 1 then
	local node_under_can = string.gsub(node_under, "_full$", "")
	minetest.set_node(pos_under, {name = node_under_can .. "_full"})
end

local node_above = minetest.get_node(pos_under).name
if minetest.get_item_group(node_above, "wall") == 1 then
	minetest.set_node(pos, {name = wall_name .. "_full"})
end

```

paramat:
> That could probably be fixed with some fancy... 'if above but no node to the left or right, use these positions for connect_left/connect_right...' or something like that but... then it becomes complicated

That would make it too complex for me, that visual glitch is very minor, doesn't look bad to me.

mods/walls/init.lua
TumeniNodes: I will swap out in favor of this suggestion. Should be done in the next day or so

TumeniNodes:
> > That could probably be fixed with some fancy... 'if above but no node to the left or right, use these positions for connect_left/connect_right...' or something like that but... then it becomes complicated
> 
> That would make it too complex for me, that visual glitch is very minor, doesn't look bad to me, looks quite good.
> If you add a 2nd flat-topped wall node as i suggest there's no glitch.

I agree on  the first part.
On the second part, I have come to also agree, let me explain why...
Making the 'full' wall has added some building options for me personally on a couple of occasions now, since I went this route with my own personal build, and I feel players/builders will like that added ability.

@Ezhh I realize you are against this but, hopefully my explanation may be enough to help you to reconsider.
 There are 2 more parts to this I want to add via additional PRs to follow, and with one of them, this option makes perfect sense.

 I also have to go back and look at the recipes @paramat  suggested and add those as well.
I'm hoping we can move forward with this after I make all the adjustments stated above in the next couple days

TumeniNodes:
> @TumeniNodes You are confusing me at this point. You say you will follow rubenwardy's suggestion, which would still be the original approach just with a small code improvement, but then you also say within the same hour that you'll follow paramat's suggestion, which is completely different and wouldn't need that code to begin with.
> 

I am going to use the code rubenwardy suggested. Then for some reason I forgot the recipe suggested was related to adding the 'full' wall.

When I mentioned that I agree to the 'full' node being added to inv it was not meant that I am adding the 'full' node to inv... I was simply explaining why I feel it would be a good idea (not saying I was changing that in this PR)
Sorry for the confusion, that was me being forgetful again, and confused myself.

paramat:
To be clear, i will still be :-1: with rubenwardy's code improvement applied :)

It might be helpful to describe what you want to do in addition later, as that might affect the implementation.

TumeniNodes:
There are 2 addition things I will propose...
a - a lower/thinner wall, to be used between the current wall nodes (same idea as fence rails), and as a bonus they add another option for a different type of wall.

b - adding xpanes to the 'connects_to' group (for reasons I will show when that comes up)

'a' addresses offering players a new wall type. That new lower-mid wall will function the same as this PR (swapping out for a full node, if another is placed atop it)
And these 2d walls look very nice

The 3rd part I plan, is to add a couple more nodes to walls.. (most likely brick, and block types) i.e. 'brick', stone_brick, stone block, etc... and each of the varieties... stone, sandstone brick/block, desert sandstone brick/block, and so on

Don't hyperventilate when reading that last one

mods/walls/init.lua
```diff
@@ -25,6 +25,56 @@ walls.register = function(wall_name, wall_desc, wall_texture_table, wall_mat, wa
...
+			end
+		end,
+	})
+
+	minetest.register_node(wall_name .. "_full", {
+		description = wall_desc,
+		drawtype = "nodebox",
+		node_box = {
+			type = "connected",
+			fixed = {{-1/4, -1/2, -1/4, 1/4, 1/2, 1/4}},
```
t4im: Excuse the nitpick :-)

No need for a table of nodeboxes with a single nodebox.
It's better to get rid of the extra `{}`s to not waste the memory (The definition is stored for the entire server lifetime).
This is only a little here, but can add up as people copy MTG code everywhere.

Same for the other nodebox definitions below of course.

TumeniNodes: I will attend to this soon, thank you. Sorry so long in responding

MoNTE48:
@TumeniNodes, you can do rebase? I have been using your version of the mod for a while and I like how it works.

TumeniNodes:
@MoNTE48 I can, but it might be about a week or so... I haven't had time to touch anything MT/MTG related in a while.
Apologies for any inconvenience

paramat:
I think this should be closed on the principle that we should not further bloat MTGame with low-priority special features, when these can be an optional mod instead.

I think we should officially move MTGame to a state of being 'bugfixes, maintenance and essential refinements only', because MT really needs to move on from MTGame. If we keep focussing on MTGame we will not move on.

TumeniNodes:
I'm in full agreement with the above comment, (all of it)
I won't close this myself, just in case of "odd reasons"
I am far too busy right now with other things, and only have enough time today to type this.
I'll be back eventually
I hope everyone is doing ok during all the craziness
0/

MoNTE48:
The comment above rather shows how much MTG Core Devs put their dick on players and contributors. They took a cool and promising project and simply kill it.
For them, it is more important to create issues, talk with yourself and close them.
Or add, delete, and then again readd unnecessary to anyone floatlands.
And also put Blocker on such Issues (https://github.com/minetest/minetest_game/issues/2703), which are necessary for 2.5 cripples (прости Wuzzy🙂).

rubenwardy:
Please tone it down

My position regarding this PR is unchanged - I like this addition, and I don't think it's too complicated

paramat:
Keeping open due to support from 2 core devs.

PR #2246, opened 2018-10-18
cook coral into skeleton coral

tenplus1:
Since removing the abm that changes orange and brown coral into skeleton coral we need another way of obtaining it, so this pull adds cooking recipes to bake them both into white skeleton coral instead.


SmallJoker:
What is the reason for closing this PR right after opening it?

PR #2252, opened 2018-10-26
New glass & glassdoor texture

SmallJoker:
For comparison old/new:
comparison image


Ezhh:
Sorry, but :-1: 

It's a glass door, so the texture needs to match glass nodes.

Compare current:
oldglass

To this:
newglass

This is a nice texture, but it simply doesn't match.

Ezhh:
Note: I'd support it if you change the glass nodes texture as well, and make the patterning on the door consistent. Given the moved/improved position of the handle, they can be lined up better to match the nodes.

VanessaE:
I've always been a little dissatisfied with mtg's glass texture as well.  I mean...blue?  with broken edges (missing pixels)?  Blah.

SmallJoker:
00724da

paramat:
Since glass nodes are so widely used, it would be better to colour the glass door to match glass nodes instead of the other way around. With these new glass textures the frames look less glasslike because they are less blueish. Also, builders may not like the colour change to glass.

The correct approach is: Independently assess the colour of glass and change it if necessary, then make glass doors match glass nodes in colour.
This PR is doing things backwards. And now you've changed a huge amount of textures unnecessarily :)

The streaks on the glass should exactly match those of glass nodes, same design, same positioning. So that glass doors continue to look consistent when used with glass nodes.

Not having gaps around the edge is fine.
New design of hinges and handle: I'm neutral, but no objection.

:-1:

paramat:
default_glass128

The new glass node texture is too white, the edge stands out too much and it looks less like glass now. The gaps around the edges which help to make it look glass like have been removed. I think it looks much worse.

paramat:
screenshot_20181027_124404

screenshot_20181027_124627

Actually, trying it in-game, surprisingly it looks no worse than the current texture. I do like how the shine lines are shorter. Sorry, i'll edit my review.
All i request then is the shine on the door matching cubic nodes, then i'm neutral.

Ezhh:
What paramat said +glass pane textures + license update for stairs textures. Then it's :+1:  for me too.

VanessaE:
Hell, I just figured you guys wouldn't mind a new, well-tested door image.  I didn't expect to turn into "let's rework all glass things" in mtg. :smiley:

The whole issue of making glass look like glass is this fear of using *gasp* texture alpha.  Yes I know MT has problems due to lack of alpha/Z sorting, but that affects water, too, and no one seems to have a major issue with that.  See also, doyousketch2's stained glass mod.

DS-Minetest:
:-1: 
[Glass often looks more blueish that white.](https://s14-eu5.startpage.com/cgi-bin/serveimage?url=https:%2F%2Fcontent3.jdmagicbox.com%2Fcomp%2Flatur%2Fc7%2F9999p2382.2382.140214200751.p1c7%2Fcatalogue%2Fgujarat-glass-and-aluminum-latur-ho-latur-glass-dealers-2jgezzs.jpg%3Finterpolation%3Dlanczos-none%26amp;output-format%3Djpg%26amp;resize%3D1024:370%26amp;crop%3D1024:370px;*,*&sp=507ed671bbc06b319c0e60cfedc15930)
The missing pixels look somehow good, without them it looks like a solid steel frame.
Also the blue colour makes it look less visible, less existing, softer. (Maybe because the sky is white-blue.)

VanessaE:
Glass is only blueish if it's tempered, as in car windows/windshields.  Normal window glass isn't.

Looks like a steel frame?  Then use alpha.

VanessaE:
Glass is only blueish if it's tempered or hardened, as in car windows/windshields and some 3d printer build plates. Normal [soda-lime window glass](https://en.wikipedia.org/wiki/Soda%E2%80%93lime_glass) isn't.  Even borosilicate (old Pyrex) glass isn't, usually.

Looks like a steel frame? Then use alpha properly.  A bit translucent on the frame, moreso on the streaks.

Ezhh:
I've always found the bluish colour makes it look more like it has a metal frame. Typical window glass isn't this colour.

paramat:
We can't use semi-translucency. The reason it's used for water is because almost nothing else uses it, once something else uses it the weirdness multiplies. Semi-transparency can be done with texture packs.

VanessaE i'm confused by your seemingly contradictory comments in IRC:

19:13 VanessaE request:  import homedecor's old glass door textures (prior to today) to use for mtg's glass doors.  imho mine look better ;)
19:14 VanessaE (I've since gotten rid of mine in favor of mtg's)

I made the current MTG glass door texture, the hinges are not great i don't mind those being changed, black wasn't a great colur for those either. But i still prefer my clean and simple horizontal handle. However i'm not going to make a fuss about that, but only on the condition that my lovely obsidian glass door texture is not altered (or that i alter that myself).

I'm still a little concerned about the brightness of the glass edges and the effect on the appearence of builds. It now has a lot more contrast. Keep in mind those who do not use connected glass, every node will have this bright frame.

paramat:
Are the greyish pixels on the doorframe above and below the handle shadow? below only might be better.

paramat:
'stairs glass outer stairside' has a misplaced shine lne.

paramat:
screenshot_20181027_183708

A single placed door will have shine in a non-matching direction.

paramat:
screenshot_20181027_184801

Hmmm, the level shine lines on the side of slabs look good, maybe the normal glass texture should have level shine lines to match?

VanessaE:
> We can't use semi-translucency.

\*Grumble\*.  It works well enough in practice in the stained glass mod, and I use it in a few places in my mods, as well.  It's not hard to see it in use on my public servers.

Incidentally, the reason I didn't use it in Home Decor's glass door texture is simply because it isn't used in default glass.  Having one without the other made it seem "out of place".

> Semi-transparency can be done with texture packs

No, it can't.  Besides drawing with good use of alpha in the texture image file, you have to enable texture alpha in the node definition as well.  A texture pack can't do that.

> VanessaE i'm confused by your seemingly contradictory comments in IRC

"gotten rid of mine" refers to the *nodes* in favor of mtg's (i.e. my doors were redundant), not their textures.

> Are the greyish pixels [...] shadow [...] ?

(noting you struck this out)

Yeah, kinda-sorta.  The handle is supposed to be steel (formerly represented by a steel ingot in Home Decor's recipe, in place of one of the usual six glass), but with the glass also being grey, I needed a bit of contrast to make it look like more than just random pixels.

paramat:
Oops, and, i see.

DS-Minetest:
>No, it can't. Besides drawing with good use of alpha in the texture image file, you have to enable texture alpha in the node definition as well. A texture pack can't do that.

I've got a question: Does [`use_texture_alpha`](https://github.com/minetest/minetest/blob/2322078fe480a338422b35ee29fee3c0767f3e1d/doc/lua_api.txt#L5501-L5505) also refer to texture modifiers (like [colorize](https://github.com/minetest/minetest/blob/2322078fe480a338422b35ee29fee3c0767f3e1d/doc/lua_api.txt#L563-L573)) ([which can be overridden by texture packs](https://github.com/minetest/minetest/blob/2322078fe480a338422b35ee29fee3c0767f3e1d/doc/texture_packs.txt#L140-L170))?

>The missing edge pixels just made it look damaged, imho.

What exactly should be damaged? You get glass by melting sand and don't add any frame to it. The edge might even be splintery. There is nothing that can be damaged.

This new glass is brighter than [iron glass](https://github.com/minetest-mods/moreblocks/blob/10dd84a7f190204f713dd551ec3b3d17cef7d959/nodes.lua#L186-L195) and ergo takes possibilities.
Actually the current glass isn't that blue, it only seems to be so blueish in comparison with very white things like the new glass.
What about adding a new steel framed glass to mtg with this texture?
<details><summary>some screenshots</summary>
Click on an image to make it bigger.

what is shown| connected glass on | connected glass off
---|---- | ----
current glass texture|screenshot_20181027_212854 | screenshot_20181027_212930
new glass texture|screenshot_20181027_214054 | screenshot_20181027_213237
current glass texture compared with moreblocks iron glass (new glass texture is even brighter than iron glass)|screenshot_20181027_212822
new glass texture compared with moreblocks iron glass|screenshot_20181027_214050 | screenshot_20181027_213247
</details>

VanessaE:
>I've got a question: Does use_texture_alpha also refer to texture modifiers

I'm not sure, but it *does* work with param2 coloring.

> What exactly should be damaged? You get glass by melting sand and don't add any frame to it. The edge might even be splintery. There is nothing that can be damaged.

One assumes that the act of smelting sand into glass in the game includes a hidden step of pressing the molten glass into a mold to make a cube out of it.  Those missing pixels make it look like someone took a hammer to the edges of the cube after cooling.

After watching some videos to familiarize myself with the process of *actually* melting ordinary sand into glass, I conclude that while amateurs using torches are on a wild goose chase, the process produces a reasonably-rounded blob when done in a well-controlled manner (i.e. with a kiln/forge and proper glassmaking tools).

Nevermind mass production, let alone what a good glass-blowing artist can do.

> Actually the current glass isn't that blue

To my eyes, it looks blue regardless of what color I compare it to (I tried black, white, grey, red, yellow).  In any case, the edges are #C0C0E3 in the original file, with the streaks varying (#DFDFF0, #CCCCE8 mostly).

In any case, glass color comes from impurities mixed into the sand.  For blue, it's generally some amount of some cobalt-based compound, and I don't imagine we're pouring any of that in our sand. :stuck_out_tongue: 

> What about adding a new steel framed glass to mtg with this texture?

RBA's old framed glass mod covers that.

VanessaE:
When this goes in, I'd like to also request the new images be pushed to stable-0.4 also (yeah, yeah, I know, but it would be trivial to do)

TumeniNodes:
I love this design.

paramat:
LOL 5.0.0 is really good you know.

VanessaE:
Good it may be, but released it is not :stuck_out_tongue_closed_eyes:

paramat:
@VanessaE i wouldn't mind adding 'use texture alpha' to glass nodes so that texture packs can choose to use it, but actually using alpha in default textures would force significant visual glitches on users who didn't choose it, that's what i disapprove of.

VanessaE:
I can accept that.  Of course, the proper course of action is to fix the cause of the glitch properly.

paramat:
That's a very big task and considering the current extreme lack of core dev time, not high enough priority to justify working on. You would have to do it yourself ;)

paramat:
So for this PR, seems fine except:
* Shine direction needs reversing on the door.
* We should consider whether to make the locations of the shine lines consistent across all glass nodes, and what locations to use. I suggest all shine lines are located at the centre of each 8x8 sub-square, because this is what seems necessary for corner stairs.

I'm finding i actually prefer these glass textures to the old ones now.

DS-Minetest:
The shine stripes are differently rotated at door_a and door_b, only one of the states has stripes that have the same direction as the glass nodes ones. To change this, door_a or door_b could be overridden. However this would cause the texture to be different after closing or opening the door.

paramat:
Yes, but fixing is not worth it, the shine direction is reversed on the back of the door anyway. However the optimum is when placing a single door the shine matches viewed from outside a building.

paramat:
@SmallJoker I can see why you have located the shine lines where they are on the various nodes, it's not a big issue. So my only request is attending to https://github.com/minetest/minetest_game/pull/2252#issuecomment-433640418

SmallJoker:
@paramat Fixed it. Here's a double-door as a gift (sarcasm).


Double-door

PR #2263, opened 2018-11-13
Add craftguide mod

LoneWolfHT:
https://github.com/minetest/minetest_game/issues/1435

craftguide now supports sfinv https://forum.minetest.net/viewtopic.php?f=11&t=14088&p=335570#p335498


Ezhh:
Just some initial thoughts:

1. I'd prefer it as a sfinv tab if possible and it should be as consistent as possible with existing tabs. Compare the differences in the buttons (and also width, which is a factor to consider if trying to adapt to sfinv).


2. Visually the alignment of things is all over the place. There are ten items across, but the arrow for recipes is mostly under item # 6 instead of centred, so more toward the right, even though the right side of the menu is where you are more likely to get more content. This leaves a big imbalance in terms of where the empty space is:

Everything needs moving slightly to the left. It would also be nice if the items at the bottom aligned more closely with the ones above.

3. I really don't like the inconsistency between fuel recipes and normal recipes. Let's look at two entries:


When I select the tree is shows me what the tree can be turned into (fuel), but when I select the stick, it shows me what can make a stick, meaning the two types of recipe are moving in opposite directions.

Additionally you only get to see something can be fuel if you can't produce it from other items. In this case, the stick can also be fuel, but because it can be crafted you don't get shown that.

--

I want to stress that I am in favour of seeing a craft guide added to the game, and these are just some starter points for consideration rather than hard requirements for a merge.

pauloue:
1. My fork does this: https://github.com/pauloue/craftguide. This mod has an option to use an sfinv tab as well.
2. It's not centered to make room for the button to switch between alternate recipes or usages.
3. For items that have both a normal and fuel recipe you can see if it can be used as a fuel by looking at the tooltip.

paramat:
In the issue rubenwardy requests this should use sfinv and i agree, the issue has somewhat planned this already.
rubenwardy wrote https://github.com/minetest/minetest_game/issues/1435#issuecomment-301525159
> I like craftguide, however think that it should be a sfinv page instead

So a version of craftguide that uses sfinv seems to be the plan currently.

Ezhh:
> It's not centered to make room for the button to switch between alternate recipes or usages.

It's poorly aligned even then (that button should be smaller, then the arrow could be moved over), but this is low importance anyway.

> For items that have both a normal and fuel recipe you can see if it can be used as a fuel by looking at the tooltip.

The issue is the inconsistent display. Surely it would be better to find a consistent solution? Maybe invert the arrow to point toward fire and keep the selected item in the usual placement, or where there is a fuel recipe, generate a new area on the left which could then be visible as well as the crafting recipes on the right. With some minor placement adjustments (there's going to need to be some to get this neatly into a sfinv tab anyway) it could fit.

LoneWolfHT:
@kilbith If you haven't seen this PR yet you may be interested in the 'thoughts'

LoneWolfHT:
Does the craftguide need to add the sfinv tab by default?

paramat:
In IRC we've been informed by the author that craftguide already supports sfinv, see discussion http://irc.minetest.net/minetest-hub/2018-11-13#i_5438694

Napiophelios:
Are you guys testing this with the proper setting?
Kilbith mod looks pretty good as is.

Currently you have to add "craftguide_sfinv_only = true" 
to the minetest.conf file and you get this:
screenshot_20181113_184811

LoneWolfHT:
> In IRC we've been informed by the author that craftguide already supports sfinv, see discussion http://irc.minetest.net/minetest-hub/2018-11-13#i_5438694

Oh. I should have explained that. They do. 

Again though: Should I set it up so that the sfinv tab is enabled by default?

Napiophelios:
>Again though: Should I set it up so that the sfinv tab is enabled by default?

IMO, if this is added to MTG, all the frills should be stripped away
and just have the "SFINV page only" in inventory
if the player possesses a craftguide (similar to the map and binocular mods handicaps).
maybe also limit it to progressive mode as well

paramat:
> IMO, if this is added to MTG, all the frills should be stripped away
and just have the "SFINV page only" in inventory

Yes.

LoneWolfHT:
> maybe also limit it to progressive mode as well
I don't think a progressive mode craftguide would 'fit'. Right now Minetest is what I would call 'Creative Survival'

Ezhh:
I wouldn't mind progressive mode being kept as an option, but I don't believe it should be the default.

TumeniNodes:
and someone please fix those damned spaces between tabs

Napiophelios:
You guys should give Pauloue's fork a look over
https://github.com/pauloue/craftguide

his formspecs need to be tweaked a bit
but over all it looks good
He also uses images for the buttons.

screenshot_20181113_175713

pauloue:
My fork also strips away all the frills (separate form, intllib, book, sign, sfinb_buttons integration)

LoneWolfHT:
Checked pauloue's fork. I'm in favor of using that instead of jp's craftguide mod.

Ezhh:
The minetest-mods version also made some changes based on my feedback that are worth looking at - for both layout and for fuel recipes.

paramat:
https://github.com/minetest/minetest_game/pull/2263#issuecomment-438893924
PR author prefers using a different fork of craftguide, and this PR hasn't made the requested changes, so closing.

Ezhh:
Sad to see - just commenting to add that I will support this if anyone picks it up and works in the changes I'd like to see (which again, are mostly present in the minetest-mods version from what I know).

LoneWolfHT:
It looks like the minetest-mods repo has included paloue's changes and added a little on top of that
I'm willing to start this up again. But it looks like I deleted the PR's fork. Shall I create another PR?

paramat:
Yes a new PR is fine.

PR #2273, opened 2018-12-06
Prevent crash if some mod inadvertently allowed doors to rotate.

sofar:
Reported by a user in IRC. Apparently some screwdriver mod allows this. Neither the mod (xdecor) nor the standard screwdriver does this in stock 0.4.x or 0.5.x - rotation isn't permitted. It is however still possible that mods override this and the door API code will croak when it finds a param2 value that isn't `0..3`. Cheap check, so, really no big deal.


HybridDog:
#1297

PR #2293, opened 2019-02-01
Tweak permafrost and stones textures

random-geek:
I made a couple of modifications to the permafrost textures. The changes shouldn't change the look of the node significantly, but they do seem to be an improvement.

Firstly, the base permafrost texture was updated to use the new dirt texture (minus the stones). The colours are nearly identical, the only difference is it tiles more nicely.
Secondly, an additional side texture was added to permafrost with stones, making the node look cleaner and more consistent with grass, snow, etc. The stones had to be slightly rearranged in the top texture in order for this to work.

As usual, all textures are indexed.

| Existing textures | PR |
| - | - |
| permafrost_old | permafrost_new |


paramat:
This is something i forgot to do thanks.
The new permafrost texture is fine.

It's good to have some stones at the side but i feel they extend too far down the sides. These stones are meant to be mostly on the surface instead of embedded, so extending far down the sides doesn't make sense, i think this is why i didn't add a side texture.

The new texture also reduces the darkness of the landscape too much when seen from the side, which is intentional and very important to the look of tundra. Visually there are now too many stones, they visually over-dominate.

Please can you make the side texture extend half as far down, just 3-4 stones, like a single row of stones?
Please, for consistency/simplicity, use the top or bottom of the previous top texture by erasing the rest.

Please keep the previous top texture even if it doesn't perfectly tile with the side textures, it is intentionally is sparse and has some patches of exposed permafrost. It also tiles over node edges, the new one doesn't (i understand why).
The new texture is too uniform, too high density and again, the stones now visually over-dominate.

random-geek:
Thanks for reviewing.

> It's good to have some stones at the side but i feel they extend too far down the sides.

I agree with you on that point, the stones might be a little overpowering.

> Please, for consistency/simplicity, use the top or bottom of the previous top texture by erasing the rest.

That probably wouldn't work very well. For one thing, the clipped stones at the top of the texture would look wrong. Also, I had to make the lighting direction on the side texture ambiguous so it wouldn't clash with the top texture on two out of four faces. I understand wanting consistency, but could you explain why this is necessary?

> Please keep the previous top texture even if it doesn't perfectly tile with the side textures, it is intentionally sparse and has some large patches of exposed permafrost. It also tiles over node edges, the new one doesn't (i understand why).

Again, the clipped stones on the edge would look quite out of place. If you want the stones to be somewhat sparse/uneven, I would rather rearrange them to your liking instead of using the old texture. Tiling over node edges is not not a bad thing either, but I don't see why it is needed as long as there are no straight lines in the texture with no stones. The tiling also makes it very hard to make a decent side texture.

I apologize if I sound harsh. I will try to make different textures when I have time, but I am not sure how to make the textures look good given these parameters.

paramat:
Yes sorry, i responded too soon and without enough consideration, i realised some problems with my suggestions. Now reconsidering ..

random-geek:
All the issues should be fixed now.
As an aside, I don't recall ever having changed the colour of any of the pixels. I compared the new texture to an old one and it looks the same as well.

Here's a screenshot with the new textures:
permafrost_new2

paramat:
Thanks looks much better.

PR #2382, opened 2019-06-02
Patch 1

acmgit:


DS-Minetest:
The line that you are trying to remove is already removed.

SmallJoker:
Invalid PR.

PR #2396, opened 2019-06-22
Add (simpler) craftguide mod

p-ouellette:
A craftguide mod based on [minetest-mods/craftguide](https://github.com/minetest-mods/craftguide). This mod is much simplified and has many changes and improvements.

#### Todo:

- [x] Use `sfinv_crafting_arrow.png`
- [x] Search for translated item descriptions


paramat:
1000 less lines, nice.

mods/craftguide/depends.txt
```diff
@@ -0,0 +1 @@
+sfinv
```
Panquesito7: `depends.txt` is deprecated, `mod.conf` should be used instead.

Panquesito7 (changes requested):

SmallJoker:
@pauloue What's the progress of this mod? Meanwhile you could make use of `formspec_version` if needed.
I'd like to have a built-in craftguide, but it should be possible to disable or replace it on-load. Would that be possible using sfinv functions?

pauloue:
It's ready for review. I will update to formspec version 3 soon. You can hide it or replace it with e.g.:
```lua
sfinv.override_page("craftguide:craftguide", {
	is_in_nav = function() return false end
})
```
I could add a setting to disable it too.

paramat:
Unfortunately i cannot review this, but i support it being added.

SmallJoker:
This craftguide version is also very reliable - for crafting and cooking. Recipes show up correctly and the search function works as well.

A few thoughts about the design:

1. The formspec height could be decreased by 1 - there's unused space at the bottom
2. There should be a minimalistic crafting grid for shaped recipes to be more intuitive
3. According to `craftguide_search_icon.png`'s size it could probably be optimized a bit (optipng/pngcrush)

pauloue:
1. I tried setting the height to what it was before #2517, but that results in the whole formspec re-centering when switching between the craft guide and other tabs.
2. Do you mean a little icon that appears above the arrow like the shapeless icon? Or grid lines or boxes in the recipe?
3. It's already optimized from the same icon in the engine.

SmallJoker:
1. Too bad. In this case vertically centering the content would look better.
2. I had grid lines/boxes in mind for shaped recipes to clarify where to place them exactly.

kilbith:
I wouldn't want this merged without taking the full version. `craftguide` will still be developed and people will keep using the original with a lot more features and innovation... and I fear this might cause some technical conflicts between the version in Minetest Game, and the full one.

pauloue:
I forked from the original because I didn't like the direction it was going and I didn't think it was right for MTG. The original's init.lua is now over 2k lines. Not necessarily a bad thing, but I think a lot of the features and hacks don't belong in MTG. If this does get merged the name should probably be changed.

SmallJoker:
As long the MTG version (maybe called `craftguide_lite`) can be overwritten and/or disable by other mods (`craftguide` in particular) then I don't see any issue. sfinv pages can be overwritten, so deleting them shouldn't be too much effort.

kilbith:
I don't think it was a good choice gameplay-wise to have removed the Progressive mode.

paramat:
A simpler version, suitable for MTG, is preferable. We always just add what is right for MTG without worrying about whatever optional mods are out there. Consider it a completely different mod such that there is no conflict.

paramat:
Tested, seems mergeable as it is. I support the changes you have made compared to the 'craftguide' mod.
> In this case vertically centering the content would look better.

I do not mind this being done, but i do not think it is essential for merge, could be done later.
In creative inventory and crafting tabs, the slots are hard up against top or bottom, so it is consistent for that to continue.
The information thta appears in the space below could perhaps be centred.
> There should be a minimalistic crafting grid for shaped recipes to be more intuitive
I had grid lines/boxes in mind for shaped recipes to clarify where to place them exactly.

-1 I cannot see any need, the crafting shape is already clear. That would just be extra clutter and complexity.

paramat:
Minor issue ...

screenshot_20200209_230101

Several of the dyes display an unknown node for the last recipe, however maybe this is suitable as it can be any node? Maybe itentional?

pauloue:
I didn't update the formspec to version 3 because of an inconsistency with the way sfinv tabs using old and new formspecs looked. That's fixed in https://github.com/minetest/minetest/pull/9319, but I'm not interested in updating the coordinates and there's not much benefit.

paramat:
That is fine, it works fine as it is, we can always update later.

I have not studied the code in depth, but have looked through it. The mod has been developed and refined so i expect the code is good.
It works well and is fairly harmless being a self-contained mod that does not affect anything else.
:+1:

Wuzzy2:
### List of bugs:

* If you hover a slot that represents a group instead of a specific item, you might see burning time (e.g. for `group:wood`). This doesn't make sense, groups shouldn't have a burning time, it should only be shown for specific items
* The mod uses the name `craftguide`, but this name is already taken by this mod: <https://forum.minetest.net/viewtopic.php?t=14088>. I know your PR is a fork of this mod, but it isn't exactly the same, nor can it act as “drop-in replacement”. So please change the shortname to something unique. Note the original `craftguide` mod is compatible with pretty much all games, not just MTG, so this naming collision *really* matters

### Suggestions and comments:

* Common groups such as `group:wood` or `group:stone` should have a simplified tooltip.
The current tooltip “Any item belonging to the group(s): wood” is quite technical. For specific, well-known groups I suggest to write a hardcoded string, e.g. “Any wooden planks” for `group:wood` and “Any stone” for `group:stone`, etc. and only fall back to the technical tooltip for unknown groups or if multiple groups are used at once. At least that's how I did it in MCL2
* The buttons to switch thru alternative recipes (inconsistently) have no tooltips, which is inconsistent with the page flipping buttons
* Burning time is inconsistent. It only appears in the tooltip if you hover the item in the recipe or result, but not in the upper item list. I think if you want to show burning time at all, you should do it consistently at least, either show it for all item buttons, or none
* I don't think this mod needs `craftguide_furnace.png` as it's going to be “baked in” into the game
* No `minetest.get_translator` is used so far
* Overall, I like this! It's simple and straight-forward and has no visual clutter. If this PR doesn't make it, then nothing will.

> I don't think it was a good choice gameplay-wise to have removed the Progressive mode.

Maybe, but given the extremely minimalistic nature of MTG … I'm very much in favor of spoiler-free crafting guides. But core MTG has so little content, that there's simply not much to spoil anyway. But it gets a problem if you have tons of mods, but then you probably have added a custom crafting guide anyway.

### Crafting guide checklist (from <https://wiki.minetest.net/Crafting_guide>):

* Shows shaped recipes: Yes
* Shows shapeless recipes: Yes
* Shows indicator when recipe is shapeless: Yes
* Output count is shown: Yes
* Cooking recipes are supported: Yes
* Cooking time is shown: Yes, in tooltip of furnace icon
* Groups are supported: Yes
* Fuels are suppored: Yes, in tooltip of item
* Replacements are supported: Nope
* Repairing recipes are suppported: Nope

Note that not all things need to be completed for a decent crafting guide. Replacements are the most complex thing about crafting, but is also the least important thing IMO. MTG uses it only once in the entire game, that is when you use a lava bucket as fuel, it turns into an empty bucket.
Repairing recipes are less esoteric. They should IMO also be supprted as they are still crafting recipes, technically, and they are not obvious at all (that's the point of having a crafting guide!). My recommendation is to just display them to the player as shapeless recipes for each repairable tool. The dumb thing about repairing in Minetest is that there's no way for mods to detect whether repairing is enabled and by how many %. For now, I suggest to just ignore repairing recipes. Hopefully Minetest lets you detect repairing recipes in a later version.

Wuzzy2:
Another bug:
Several dye recipes have unknown items, for example, recipe 3/3 for brown dye.

Wuzzy2:
Usability remark: The search bar is at the top, but this is inconsistent with the Creative Inventory, where it lies at the bottom. It's generally a better usability if you don't let widgets jump around randomly.
It would look really nice if the crafting guide and the Creative Inventory visually match.

paramat:
Wuzzy2, thanks, i removed my approval, i suspected it would turn out unwise for me to do that based mostly on trust.

Wuzzy2:
I feel this naming collision is only done due to pure laziness, NOT because there is any justification for it.

Naming collisions of the technical name (!) are bad because:
* Confusing as hell
* Creates ambiguity
* Annoying in dicussion (“Which `craftguide` mod did you mean?”)
* Screws up mod search and other tools

It's a GOOD thing to have unique mod names, why do you oppose it?

You have provided no reason at all to justify the naming collision.

pauloue:
> If you hover a slot that represents a group instead of a specific item, you might see burning time (e.g. for `group:wood`). This doesn't make sense, groups shouldn't have a burning time, it should only be shown for specific items

Maybe it would be better to only show it for specific items, but this isn't a bug. `group:wood` does have a burning time that applies to all items with the group.

> The mod uses the name craftguide, but this name is already taken by this mod:

How about `mtg_craftguide`?

> Common groups such as group:wood or group:stone should have a simplified tooltip.

I can add that.

> The buttons to switch thru alternative recipes (inconsistently) have no tooltips, which is inconsistent with the page flipping buttons

I'll fix that.

> Burning time is inconsistent. It only appears in the tooltip if you hover the item in the recipe or result, but not in the upper item list. I think if you want to show burning time at all, you should do it consistently at least, either show it for all item buttons, or none

I agree it's inconsistent, but I'm not sure about showing the burntime in the item list, so I'm tempted to not show it at all. Already not all fuels are shown (ones that don't appear in any other recipe) and it's easy enough to try placing an item in a furnace. The exact burntime is not important and can be found on the wiki.

> I don't think this mod needs craftguide_furnace.png as it's going to be “baked in” into the game

Are you suggesting using `default_furnace_front_active.png`? That would mean depending on default.

> Several dye recipes have unknown items, for example, recipe 3/3 for brown dye.

That's because there is no brown flower. Maybe MTG shouldn't register this recipe or maybe craftguide can hide recipes with nonexistent items.

>It would look really nice if the crafting guide and the Creative Inventory visually match.

Good idea.
screen

Wuzzy2:
> That's because there is no brown flower. Maybe MTG shouldn't register this recipe or maybe craftguide can hide recipes with nonexistent items.

I think impossible recipes should just be hidden, as they just add useless clutter.

I do not think the recipe is a bug, as it's purely group-based. It's just that the group is empty in vanilla MTG. I think the idea behind this is that mods could add an item that matches this group.

> Are you suggesting using default_furnace_front_active.png? That would mean depending on default.

Hmmm, yeah, that sucks. I don't really care. Maybe it's better to not touch it then.

> The exact burntime is not important and can be found on the wiki.

That's a terrible argument, the wiki is not a good source, it's always very painful to update it by hand only to bump boring numbers. It's very error-prone as well. At least the fact that items *are* burnable should be visible. The wiki is OK *for now*, but it always requires manual intervention, which is bad.

> How about mtg_craftguide?

I don't care, anything that doesn't collide with `craftguide` is OK.

> Maybe it would be better to only show it for specific items, but this isn't a bug. group:wood does have a burning time that applies to all items with the group.

But that's a fallacy. Although you can register burntimes for groups, Minetest still translates it to the individual items, and Minetest does not report the burntime of groups (IIRC), only of individual items. As far I understand, registering a burntime of a group is only a shortcut, basically.
And, in fact, the burn time for different woods differs in MTG as well, so wood has no common burntime. It would be wrong to claim a burntime of e.g. 30 for `group:wood`.
The most logical solution is to not show burntime for groups.

Wuzzy2:
Hmmm, but OK, it's not that big of a deal if not all fuels are religiously shown. The shaped and shapeless recipes are obviously much more important.

pauloue:
Maybe just indicate whether items are fuels without showing the burning time?
screen

paramat:
> It's a GOOD thing to have unique mod names, why do you oppose it?

I do not oppose it, i was just unsure. However, you have convinced me a different name is better.
> group:wood does have a burning time that applies to all items with the group.

Ah yes, i forgot that, there is a fuel recipe for that group.
> How about mtg_craftguide?

Fine for me.
>  but I'm not sure about showing the burntime in the item list, so I'm tempted to not show it at all.

Fine for me, i found that a somewhat unnecessary feature. We could always add it in later if we really want.
> Are you suggesting using default_furnace_front_active.png? That would mean depending on default.

It is always better to try to not depend on default, so please do not use the MTG texture.

paramat:
> Maybe just indicate whether items are fuels without showing the burning time?

Fine for me, and it seems that solves the burntime problems discussed above?

pauloue:
I implemented Wuzzy's suggestions.
I added simplified tooltips for all group ingredients in MTG since there aren't too many.
Translation support added.

@Wuzzy2 I attributed the shapeless icon to you (https://github.com/minetest-mods/craftguide/pull/15) and assumed CC BY-SA. Let me know if that's incorrect.

rubenwardy:
> I feel this naming collision is only done due to pure laziness, NOT because there is any justification for it.

I haven't reviewed this PR, but I believe this is a modification of JP's craftguide with unused features removed? In which case, using the same name is desirable as it is essentially the same mod. As long as it doesn't introduce breaking changes that would make it incompatible with upstream craftguide

rubenwardy:
If the name is different, how easy would it be for upstream craftguide to disable it? If the simplifications in this PR makes it incompatible with upstream - I guess it will - then renaming does make sense

Wuzzy2:
@kilbith, I summon you!

Basically, this PR wants to introduce a radically downsized version of an older version of your `craftguide` mod but the PR author refuses to change the shortname.

@rubenwardy: Have you checked out the latest version of `craftguide` lately? The difference is like night and day.

rubenwardy:
> but the PR author refuses to change the shortname.

The name has already been changed to `mtg_craftguide` in this PR

rubenwardy:
> Have you checked out the latest version of craftguide lately? The difference is like night and day.

I'm using it right now, the style looks a lot different but I can't see any functional differences. I use the sfinv mode for it, however

Wuzzy2:
This doesn't work:

> S("No usages.\nClick again to show recipes.")

Split it into 2 strings to make it translatable.

pauloue:
> This doesn't work:
> 
> > S("No usages.\nClick again to show recipes.")
> 
> Split it into 2 strings to make it translatable.

Is that an engine bug? Seems like it should work according to the docs and it does work if I replace the `label` with a `textarea`.

rubenwardy:
> Is that an engine bug? Seems like it should work according to the docs and it does work if I replace the `label` with a `textarea`.

It is bad practice, generally, to have more than one sentence in the same translation

paramat:
>  but I believe this is a modification of JP's craftguide with unused features removed? In which case, using the same name is desirable as it is essentially the same mod. As long as it doesn't introduce breaking changes that would make it incompatible with upstream craftguide

I think it is best to use a different name (as it now does), consider it a different mod, and not be restricted by having to maintain compatibility with the source mod (being forced to maintain compatibility is what causes so many problems in MTG, we should avoid this when not necessary, as here).

SmallJoker (approved):
Works nicely and covers most the most needed features :+1:

paramat:
Keeping open as concept supported by at least 2 core devs.

PR #2399, opened 2019-06-28
new_default_grass_texture

TumeniNodes:
I here the groans now "another new grass texture PR?"
Yes,
I designed a new texture to cut down on the  appearance of linear effects, using 16px it is difficult to avoid them but, this reduces them a bit. And I stuck to the original colors as best as I could.

It's a trivial matter, I just thought I'd put it up to see if it is a desired change. (I know it has been changed often in the past)

default_grass

new grass proposal 06-28


SmallJoker:
How about a real global texture which spans a mapblock instead of updating the single-node texture?

TumeniNodes:
@SmallJoker  I have been out of the loop for a couple months, has this been implemented yet?

paramat:
Is this an edit or a completely new pattern?

The grass texture is a hard one to work on. I have very high standards for it due to the amount of work the last one took, i think it took me 20+hrs of work. Just mentioning to explain why i won't change it casually =) I would prefer a minimal edit to reduce linear patterns.

I disapprove of global textures in MTG.

TumeniNodes:
This is a completely new design from scratch... total approx 1 week of time combined

TumeniNodes:
Another method (though may be costly) is to make mapgens read multiple textures as it does for the grass-types (i.e. default grass 1-5)
It could read default_grass_01-05, default_stone_01-05
This was something I was picking away at quite a while ago but got distracted.
Using the same texture but altering the inner pixels to give a nice variety, while the outer pixels connect nicely. Rotating would disrupt the effect though, so none of that.

SmallJoker:
@TumeniNodes  Example:
```Lua
minetest.override_item("stairs:stair_stone", {
	tiles = {{
		name = "dirt_medium.png",
		tileable_vertical = true,
		tileable_horizontal = true,
		align_style = "world",
		scale = 8
	}}
})
```
dirt_medium.png:
dirt_medium

random-geek:
Good texture, but I definitely think it needs to be a little lighter/less saturated to match the overall feel of the old version. I wish it had more color variation as well, but that is likely to un-solve the problem you were trying to solve.
Also, does the side grass texture need to me modified as well?

paramat:
> Another method (though may be costly) is to make mapgens read multiple textures as it does for the grass-types (i.e. default grass 1-5)

There's an engine issue for that, random texture selection from an animation strip.

TumeniNodes:
@SmallJoker interesting, I'll have to play with that over some time.

@random-geek this matches up with the side texture well so I see no need to change it. This was done as a personal preference to reduce the linear effects of the top grass texture. And yes, adding another color will lead back to causing a pattern.
Using 16px textures it is difficult to avoid running into patterns. With 2d platformers this is expected and actually desired I guess but, being a 3d voxel game it looks out of place.

@paramat yes, I recall that issue (it's quite old isn't it?) I do agree using sprites is a decent option, or even simply adding roughly 4-5 more randomized textures for existing ones. They could even be added into a sub-folder but that just makes it more complicated.

paramat:
Tested, it's very good, i think i prefer it to the current one.
Here it is enlarged:

default_grass256

I do want the side texture changed to match, but i'm happy to do that myself in a separate PR.
So if you change the texture credits for grass i'll +1.

TumeniNodes:
I honestly don't think I would change the grass side texture, it's perfect and they blend very well together.
About the _only_ thing I might see changing, would be to remove the darkest pixels on the side texture.
Other than that, I have always loved the grass side texture as is, it looks very good.

It's up to you, we can leave it as is or, I will do another PR to do what I mentioned above, it would be no trouble to do that one small detail.

SmallJoker:
64x64 px texture example (spans 4x4 nodes)
screenshot_20190630_115554

TumeniNodes:
That actually does look nice.
Perhaps we should be utilizing this method?
What are the pros, and what are the cons?

paramat:
:-1: for global textures, i can't stand the way it looks, looks like a modern visually-fancy FPS game. MTG is simple and retro. Please leave that to other games.
Global textures will cause complexities and make texture and texture pack creation much more difficult, 16x16 is already hard enough.

TumeniNodes don't bother with side texture, i'll do that in a follow-up PR.

benrob0329:
That global texture example isn't great, but I think a minimal amount of variation could be added from the current PR texture to make it fit more with MTG's style.

>i can't stand the way it looks

You can't judge all GTs based off of the few examples we've had so far. IKEA's use of them is different than MTG's might be which is different than any other game or mod. It simply adds the ability for more variation, and could be used in ways we haven't even thought of yet.

TumeniNodes:
I prefer to use random textures.
With some very brief testing, when placing nodes individually, the edges don't match up.
At least with using random textures, the edge pixels can be kept the same, while altering the inner pixels, like so
random textures

paramat:
Yes, i'm sure there are some occasional uses of GT i would like (like large artificial panel areas), my comments are about using it just to reduce repetition, especially for natural nodes.

TumeniNodes:
Any others who've tested and could approve?

random-geek:
I know lighter grass textures haven't been popular in the past, but...
Here's the same texture, just lightened a *tiny* bit. I think it goes better with the colors of the plantlike grass without being too extreme a change.

default_grass

screenshot

LoneWolfHT:
> I know lighter grass textures haven't been popular in the past, but...
> Here's the same texture, just lightened a _tiny_ bit. I think it goes better with the colors of the plantlike grass without being too extreme a change.
> 
> default_grass
> 
> [screenshot]

Maybe darken the plantlike grass a tiny bit too. :+1:

paramat:
@Ezhh (sorry to disturb you).

LoneWolfHT:
image
The side looks a little weird compared to the top

Also:
image

paramat:
Don't worry about the side, i'll update that afterwards.

What's the issue in the 2nd screenshot?

TumeniNodes:
@LoneWolfHT  how so? They match up perfectly to me, I even made sure to set some of the px colours to match up with the side texture...
Aside from the darker pixels, which _could_ be lightened (but I honestly see no reason to)

LoneWolfHT:
> Don't worry about the side, i'll update that afterwards.

Okie
> What's the issue in the 2nd screenshot?

The green grass is smoother but the dry grass is 'sharper'. It doesn't look good IMO

LoneWolfHT:
> @LoneWolfHT how so? They match up perfectly to me, I even made sure to set some of the px colours to match up with the side texture...
> Aside from the darker pixels, which _could_ be lightened (but I honestly see no reason to)

Heh, kind of a repeat but: The top grass is smoother but the side grass is 'sharper'

TumeniNodes:
Dry grass should be sharper... it's "dry grass". More brittle etc than green/lush grass, I think the mix is perfect
(perfect, is mt word for the day.... don;t judge me)

paramat:
Yes, dry grass is intentionally higher contrast due to being dry and crisp.

TumeniNodes:
Looking at that 2d image though, I'll tell you what I think we do need.... "dry_dirt"
It would be just a bit lighter, and the dirt for the coniferous litter should be darker than current dirt
(But I'll make another issue on those) gettin ot here, sorry

paramat:
Actually don't worry, i have been thinking about a different dirt for savanna, often you see reddish dirt in savanna. But a drier colour would be good too.

Ezhh:
I don't have a huge preference either way on this one, but no objections t the change either.

paramat:
Ezhh, what about global textures?
Note, MTG using global textures is likely to make me quit MTG because there will be no way to modify it to use 16px textures.

TumeniNodes:
MTG should remain as is, regarding the visual scheme (16px)
Some can make a TP using global textures. Leave such things up to a game maker.
If MTG were a first person shooter game like doom, I could see going that route.

(just because you can, doesn't mean you should)
I do like the look of global textures, but they really do not fit MTG at all.

paramat:
Besides my personal visual taste, here are some more points to consider:

A global texture causes texturing to become beyond the ability of most MT users. Most users can paint a 16x16 texture in an image editor, but a large texture can only be created by using a noise generator.

Noise generators are difficult to understand and use. I used one recently to create the tundra moss, despite my deep understanding of perlin noise i barely understood what the parameters of the noise generator did even after reading the documentation. It took a lot of time to get right.
Most users will flounder around adjusting parameters, not understanding what they're doing, and end up with an unsatisfactory and rather generic looking texture.

Creating textures for MTG is most modders' first experience of texturing. We need at least 1 game that is easy to texture for, it might as well be MTG as this matches its simple, retro character, and the textures are already 16px.

//////////////////////////

Pixel art will no longer be possible, that is, attention to individual pixels. Tidying up edges for better tiling, adjusting certain details, anything like this becomes impossible due to the size of the texture.
The result will be a huge reduction in the variety and artistry of textures, everything becomes the output of a few common noise generators.
The excellent grass texture of this PR would probably become impossible, as it doesn't look like noise generator output.

Most MTG textures are primarily pixel art, so global textures will look out of place.

/////////////////////////

What about HD texture packs? A not-excessive 128px pack with a mapblock size global texture means textures become 16 times larger at 2048px, which is impractically large. 1024px would be too. Imagine the memory that will use.

////////////////////////

There is no problem to solve here, this submitted grass texture tiles beautifully and looks lovely on a large scale. There is nothing offensive about how it repeats.

benrob0329:
>A global texture causes texturing to become beyond the ability of most MT users

This is a fair point, but you can always tile a 16px texture if you want to stick with that.

>but a large texture can only be created by using a noise generator

This is simply wrong, I have created global textures with hand-painted noise, and 64px pixel art is not uncommon.

> A not-excessive 128px pack with a mapblock size global texture means textures become 16 times larger at 2048px, which is impractically large

Texture sizes are only as impractical as the GPU's memory allows. To counter this, consider that every "HD" grass texture looks like crap because of the tiling effect. A global texture would allow the needed variation we need to have good HD texture packs.

>There is no problem to solve here, this submitted grass texture tiles beautifully and looks lovely on a large scale. There is nothing offensive about how it repeats.

This is true, we're just discussing the possibility of making it better :P

TumeniNodes:
Make 'use global textures' an option, and then we can all work on an official GT pack for MTG which users can d/l if they want to try it out.
But leave the base MTG at 16px tiled.
Some people will like global textures, others won't, it's always nice to give users an option.
Putting such a pack together will take some time, unless you want it to be rushed and look half-assed and cheesy (which I know, no one wants)

benrob0329:
The biggest problem is that GTs are not a matter of taste in general, but rather a tool for creating different artistic styles and effects. You could use them in a wide verity of circumstances to add more variation to a texture, or to change colors depending on position, or any other mixture of things. It's not a single art style and should not be treated as one.

We need to stop acting like GTs will only produce one type of result, or that they can only b created in a certain way for a certain purpose. It's a tool, just like everything else in the engine.

Ezhh:
First thing: **discussion of global textures should be totally separate to this PR**. This PR should be merged based on what it is, not used to launch a separate and much larger discussion around whether global textures should be used instead.

---

With that said, since I've been asked, my thoughts on global textures:

> a large texture can only be created by using a noise generator.

This isn't true. I've previously created large scale textures for another game using another method and they worked just fine. More complicated to make than current MTG textures? Certainly. Higher level of skill needed? Yes. But you do not need a noise generator. That doesn't mean it wouldn't be a good approach, maybe even the best approach (I'm not qualified to say), but it's not the only valid approach. With that said, I feel paramat is right in that they are more complicated to produce and would be outside the reach of many people who can make a simple 16*16 texture. However, that alone doesn't make them wrong for MTG.

The issue for me as far as MTG (and MT itself in general to some extent) goes is one of visual style. We have a world made of blocks and mostly straight lines, everything pretty much locked and limited to a grid. This to me brings with it a certain set of expectations; I  expect things to repeat and I expect them not to push too far toward realism. I worry that global textures could create the appearance of MTG trying too hard to be what it's not. This can result in highlighting the lack of realism instead of improving it.

Of course this is a personal opinion and others will disagree. I even feel I might be proven wrong with a set of global textures that work well together and "get it right".

But this brings me to the next point, and this is why this really needs to be a separate discussion outside of this PR: **Global textures must be done right**. We'd need a whole set of complimentary textures that work well and look good together.

I don't want to just see grass replaced and everything else left as is. in my opinion, that would leave us with an unplanned and uncoordinated mess. 

So if people truly want global textures, open an issue that's about that and plan out how to do it properly.

paramat:
@TumeniNodes this PR has had unrelated commits added to it. If it's hard to fix feel free to open a new PR and i'll merge it.

TumeniNodes:
yeh, I goofed. I will make a new one.
closing this one

PR #2412, opened 2019-07-16
add_new_dry_dirt

TumeniNodes:
New PR for dry dirt node. (ban the rebase command)
See original PR : https://github.com/minetest/minetest_game/pull/2409

PR #2415, opened 2019-07-16
add dry dirt node and texture

TumeniNodes:
OK, 3rd times a charm.
Gonna take this step by step as we confirm changes

I also threw in freebies, fixing missed ' > " for dropped node names (_FYI_ I plan to fix the rest in a PR after this, there are a few throughout the code)

~~@paramat  if you prefer, I can stop here and let you handle the mapgen code, I do not mind.
But if you would like me to finish this, we will need to confirm each change as it goes, either way is fine with me, I do not mind doing the work.~~

Nvm, I'll work on this again tonight.


paramat:
Yes this has been a tricky task. I have made several mistakes in reviewing, this is partly due to the complexity and all the related issues, but also because with tasks like this i find it hard to review because i usually realise what needs doing by coding it myself.

If you can cope with the pain you can do this all yourself if you want, otherwise you're welcome to just add the new dry dirt node and the texture and leave the rest to me. I don't mind either way.

About the changes to farming:

I'm not sure we need a new 'dry soil' node, it seems a lot to do just for dry dirt. It seems possibly acceptable to just have dry dirt turn into 'soil' when hoed. This will cause side textures to look a little weird, but the side of farmed soil is not often seen, the top is altered already by the added 'furrows'. Also, soil will soon be 'wet' which can explain the side texture change.
Until further consideration / inout, i think it's best (and no harm for now) to just override dry dirt like a normal dirt node.

mods/default/nodes.lua
```diff
@@ -435,7 +436,7 @@ minetest.register_node("default:dirt_with_grass", {
 		{name = "default_dirt.png^default_grass_side.png",
 			tileable_vertical = false}},
 	groups = {crumbly = 3, soil = 1, spreading_dirt_type = 1},
-	drop = 'default:dirt',
+	drop = "default:dirt",
```
paramat: When i wrote something like:
'We are changing ' to " for nodenames'
I only meant do that in touched code, not for all nodes. I appreciate you doing this, but this PR is complex enough and shouldn't have any out-of-scope changes added.

mods/default/nodes.lua
```diff
@@ -447,32 +448,38 @@ minetest.register_node("default:dirt_with_grass_footsteps", {
...
 	sounds = default.node_sound_dirt_defaults({
 		footstep = {name = "default_grass_footstep", gain = 0.25},
 	}),
 })
 
 minetest.register_node("default:dirt_with_dry_grass", {
 	description = "Dirt with Dry Grass",
-	tiles = {"default_dry_grass.png",
-		"default_dirt.png",
+	tiles = {"default_dry_grass.png", "default_dirt.png",
```
paramat: These 'tiles' changes in several nodes are also out of scope for this PR.

mods/default/nodes.lua
```diff
@@ -447,32 +448,38 @@ minetest.register_node("default:dirt_with_grass_footsteps", {
...
 		{name = "default_dirt.png^default_dry_grass_side.png",
 			tileable_vertical = false}},
 	groups = {crumbly = 3, soil = 1, spreading_dirt_type = 1},
```
paramat: This node should no longer spread.

paramat:
For my own sake too i'll create a list of needed changes, compiled from previous reviews:

* Add 'dry dirt' texture.
* Add 'dry dirt' node.
* Add 'dry dirt with dry grass' node, not in 'spreading_dirt_type' group.
* Keep 'dirt with dry grass' node but remove from 'spreading_dirt_type' group.
* Remove 'dry grass' from 'grass spread' ABM.
* In savanna biome registrations change top and filler nodes to 'dry dirt with dry grass' and 'dry dirt'
* In decoration registrations, for dry grass, acacia tree and log, acacia bush, change 'place on' node to 'dry dirt with dry grass'.
* Alter 'grass covered' ABM as shown here https://github.com/minetest/minetest_game/pull/2412#issuecomment-512036962


* In farming:
Keep `minetest.override_item("default:dirt_with_dry_grass", {` so that existing savanna can still be farmed.
Add `minetest.override_item("default:dry_dirt_with_dry_grass", {` with:
```
		dry = "farming:soil",
		wet = "farming:soil_wet"
```

mods/default/nodes.lua
paramat: > I also threw in freebies, fixing missed ' > " for dropped node names (FYI I plan to fix the rest in a PR after this, there are a few throughout the code)

Best do all in a separate PR then.

DS-Minetest:
Some ideas:
- Dry dirt could be turned to dirt if water is nearby.
- Dry dirt with dry grass could spread as dirt with grass to nearby dirt nodes.
- You could make dry dirt with grass spread to dry dirt but turn dry dirt to something like dry dirt with crust with an abm. So, if you dig much of the surface, there might be grassless and grass areas.

paramat:
Of course, this change has had no input from other core devs yet, i hope they are ok with it.

Concerning farming, i'm actually not sure about making dry dirt nodes farmable, realistically it wouldn't be particularly fertile soil. Preventing farming in savanna adds some challenge instead of making farming too easy (like the silly farmable desert sand thing did).
Perhaps it's best to not make any changes to the farming mod, for now at least?

Napiophelios:
"Desert farming is the practice of developing agriculture in deserts. As agriculture depends upon irrigation and water supply, farming in arid regions where water is scarce is a challenge. However, desert farming has been practiced by humans for thousands of years."

https://en.wikipedia.org/wiki/Desert_farming

TumeniNodes:
I'll be putting some time into this tomorrow (the initial parts)
It looks as though we'll really have to think about the farming aspects more.
At the very least if we add it to farming, it gives it a use, which will likely garner some support by others, as opposed to just introducing a new node/texture, just for the sake of asthetics.
I do know, I fully agree with paramat that, later on, farming with sand should really be removed. But that is something for another time.

I also like some of the ideas DS-Minetest shared.
But for now, just working in the initial node so it plays nice is my goal, and then it can be abused later.

paramat:
Hmm well i'm neutral on whether to make savanna farmable. Let's see what other core devs prefer.
And if we do i'mok with adding a 'dry soil' node, might as well do it properly.

We now have 2 more deserts, using 'sand' and 'grey sand', but i don't think it's worth making those farmable too. I prefer making deserts non-farmable to add more challenge and not make farming easy everywhere.

LoneWolfHT:
image

paramat:
This PR currently only adds the new nodes, no mapgen changes yet.

TumeniNodes:
> image
> Shouldn't this be dry dirt, not dirt?

Not yet, I will code the mapgen changes over the next couple days.

LoneWolfHT:
> Not yet, I will code the mapgen changes over the next couple days.

Ok, looks good so far then

TumeniNodes:
> Of course, this change has had no input from other core devs yet, i hope they are ok with it.
> 
So let's find out first, beforehand.
I am only saying this because I am literally fed up with trying to figure out 'rebase' and tired of having to continually, completely redo the entire PR.
Rebase sucks, as do any instructions (official or third party) on how to use it... because people leave out very important details such as (do I have to perform rebase from the branch I am trying to rebase, or from master or does it even matter?)
People who try to explain such things, do not comprehend a novice user does not know these small details, and they create more problems than solutions in the end. (which always pisses me off)

So let's see how others feel about this change before I tackle it again.
I can do it all in one fell swoop and put it up.
Either that or someone can step-by-step hold my hand through one successful 'rebase' and then I would be able to understand it and be fine from then on.
Right now I am looking at the 4th redoing of this, and holding back because I know longer grass sides will possibly be merged the next few hours or day, where I would have to redo it again... :P

TumeniNodes:
@paramat  here is the newest PR https://github.com/minetest/minetest_game/pull/2422

PR #2436, opened 2019-08-10
Torch API

MoNTE48:
Now any mod can very easily create its own 3D Torch.
I made this for mesecons_torch. There are other mods where this can be useful.

```
default.register_torch("mymod:newtorch", {
	description = "My New Torch",
	tiles = {"torch_texture.png"},
	inventory_image = "default_torch.png",
	wield_image = "default_torch.png",
	light_source = 15,
	groups = {choppy = 2, dig_immediate = 3, flammable = 1, attached_node = 1, torch = 1},
	mesecons = {some code}
})
```

It's all. Your torch will use the standard torch model, with support for placement on a wall or ceiling.

_The documentation is updated as well._


paramat:
I would prefer that an API allows the torches to be nodeboxes. So that use of Blender (or other 3D modelling program) isn't required, that being beyond the ability of most modders.
But i understand that this is a port from a fork and you may not want to alter the code much.

TumeniNodes:
>I would prefer that an API allows the torches to be nodeboxes.

No way. I have to disagree on that, the models have been perfect and were a very nice and welcome addition as far a torches go.
There are _some_ areas where models a the much better option, torches is one of them.

mods/default/torch.lua
```diff
@@ -1,131 +1,100 @@
...
+	local torch = {
+		description = def.description,
+		drawtype = "mesh",
+		tiles = def.tiles,
+		paramtype = "light",
+		paramtype2 = "wallmounted",
+		sunlight_propagates = true,
+		walkable = false,
+		liquids_pointable = false,
+		light_source = def.light_source,
```
SmallJoker: Consider making `def` a table that overrides these defaults. This way anyone can pass custom definitions, without extending the torch API.

mods/default/torch.lua
```diff
@@ -1,131 +1,100 @@
...
+		groups = def.groups,
+		drop = name,
+		sounds = def.sounds,
+		floodable = true,
+		mesecons = def.mesecons,
+		on_blast = def.on_blast,
+		on_flood = function(pos, oldnode, newnode)
+			local torch = (oldnode.name:gsub("_wall", "") or oldnode.name:gsub("_celling", "")) or oldnode.name
```
SmallJoker: `oldnode.name == name .. "_wall"` etc.

paramat:
TumeniNodes, you seem to misunderstand, i was only suggesting supporting nodeboxes as an additional option, i wasn't suggesting removing support for models.
Your comment reads as opposing the possibility of registering nodebox torches. The majority of modders are unable to do 3D modelling so would be forced to re-use the current model, so altering the shape of the torches is currently impossible for most modders, that owuldn't be a good API.

TumeniNodes:
> i was only suggesting supporting nodeboxes as an additional option, i wasn't suggesting removing support for models.

If that is the case then yes, I misunderstood.
I'm definitely onboard with adding nodebox torches as an additional option, if it is an option which seems popular and has been complained about.

MoNTE48:
**lbm** still needed after so many years?
If so, is it not better to replace it with aliases? It spends less resources and is more reliable.

MoNTE48:
@SmallJoker you can review latest version of this PR? Thanks

MoNTE48:
I do not want to rebase again.
Can someone approve of this? Or is this PR useless and will I close it?

SmallJoker:
`default:torch_wall` no longer uses the 3D model. I assume other "3D torch" registrations would not work either since the default torch is already broken.

grafik

MoNTE48:
@SmallJoker I checked it again and everything works perfectly. The model is still in use. Make sure to use the correct version of my branch, I did rebase.

SmallJoker:
Applying the PR correctly is not the issue (re-tested right now). The problem is that older texture packs use `default_torch_animated.png` to override the wallmounted 2D torch animation.
Please undo renaming the texture, or use a different name to prevent side-effects.

MoNTE48:
@SmallJoker 
default_torch.png -> default_torch_inv.png
default_torch_animated.png -> default_torch.png

SmallJoker:
@MoNTE48 Now there's no animation. The previous `default_torch.png` (not animated) was good, only `default_torch_animated.png` needs to be renamed -> Maybe `default_torch_mesh_animated.png`?

grafik

MoNTE48:
@SmallJoker, pls test again. Thanks!

mods/default/torch.lua
```diff
@@ -3,134 +3,104 @@
...
+	torch.drawtype = "mesh"
+	torch.paramtype = "light"
+	torch.paramtype2 = "wallmounted"
+	torch.sunlight_propagates = true
+	torch.walkable = false
+	torch.liquids_pointable = false
+	torch.drop = name
+	torch.floodable = true
+	torch.on_flood = function(pos, oldnode, newnode)
+		oldnode.name = name or name .. "_wall" or name .. "_celling"
```
SmallJoker: This will always be `oldnode.name = name` because `name` cannot be `nil`.
`minetest.add_item(pos, name)` will be sufficient here.

MoNTE48: I know. Therefore `name or name .. "_wall" or name .. "_celling"`

SmallJoker:
```Lua
$ luajit
LuaJIT 2.1.0-beta3 -- Copyright (C) 2005-2017 Mike Pall. http://luajit.org/
JIT: ON SSE2 SSE3 fold cse dce fwd dse narrow loop abc sink fuse
> name = "torch"
> print(name or name .. "_wall" or name .. "_celling")
torch
> name = nil
> print(name or name .. "_wall" or name .. "_celling")
stdin:1: attempt to concatenate global 'name' (a nil value)
stack traceback:
	stdin:1: in main chunk
	[C]: at 0x56131ea0a1d0
```

MoNTE48:
Ready for merging or closing 10+ days.
Any additional suggestions?

Thomas--S:
I've tested this PR (including using different texture packs and registering a custom torch). Everything seems to work fine. The code also looks good, as far as I can see.

<sub>Maybe this review helps the official game devs with reviewing and merging this PR.</sub>

MoNTE48:
after half a year it seems irrelevant and can be closed?

paramat:
This does not seem useful enough to add.

It does not allow altering the shape of the torch, a texture pack can already override torch textures, so the only benefit is being able to alter torch brightness and the sounds.
Altering brightness is not of much use because torches usually do not make sense any brighter (as they would then be as bright as fire or mese lamps) and darker makes them too dark for most users.
Altering sounds is not of much use because torches are almost always wood, so sounds usually do not need changing.
This seems too low priority and random to add to MTG. I think custom torches should be left to mods.

If i misunderstand please explain how this is a fundamental feature.
:-1:

game_api.txt
```diff
@@ -313,6 +313,27 @@ Allows creation of new fences with "fencelike" drawtype.
 	sounds = default.node_sound_wood_defaults(),
 
 
+Torch API
+---------
+
+Allows creation of new torches with "wallmounted" drawtype.
```
paramat: 'wallmounted' is not a drawtype, the drawtype of these torches is 'mesh'. 'wallmounted' is a rotation type.
I think it is better to describe this in a simpler way as 3D torches:
"Allows creation of new 3D torches."

mods/default/README.txt
```diff
@@ -9,6 +9,7 @@ Various Minetest developers and contributors (LGPLv2.1+)
 
 The torch code was derived by sofar from the 'torches' mod by
 BlockMen (LGPLv2.1+)
+The torch API by MultiCraft Development Team (LGPLv2.1+)
```
paramat: Please credit to the people who wrote the code, not to an organisation.

Thomas--S:
> I am not convinced this is useful enough to add.
> 
> It does not allow altering the shape of the torch, a texture pack can already override torch textures, so the only benefit is being able to alter torch brightness and the sounds.
Altering brightness is not of much use because torches usually do not make sense any brighter (as they would then be as bright as fire or mese lamps) and darker makes them too dark for most users.
Altering sounds is not of much use because torches are almost always wood, so sounds usually do not need changing.
> 
> If i misunderstand please explain how this is a fundamental feature.


This PR allows the registration of additional torches that match the style of the 3D torches.

For example (taken from the initial post), mesecons uses a special type of torches that conducts/transforms signals in a certain way.

Another example is the kalite torch that is (if I can remember correctly) part of the dreambuilder modpack. It is similiar to the default torch, but it is crafted with kalite lumps instead of coal and has a red flame.

Also, if I remember correctly, in some older versions of LOTT subgame, blue and white torches crafted with special lumps were used.

You see, there are some use cases for additional torches and it doesn't look very good when only the default torch is allowed to have a 3D look, but all other torches must remain flat (torchlike drawtype), if the modders want to avoid code duplication.
This PR introduces a simple API that allows the registrations of such additional torches in a 3D style matching the default torch.

LoneWolfHT:
I think it would be better if this was moved to a 'torch_api' mod that modders could copy into their games and use for the easy registration of torches. The actual default:torch registration would stay in the default mod, so the torch API would have no dependencies
If this is merged I'm willing to create a PR that would do this

MoNTE48:
ШО, ОПЯТЬ?!!
Another rebase? It was 7 months waiting for approval or closure...

MoNTE48:
@SmallJoker @sfan5 Perhaps any reviews from you?
And someone can remove "Rebase needed" label?

mods/default/torch.lua
```diff
@@ -3,152 +3,116 @@
...
-	local nodedef = minetest.registered_items[newnode.name]
-	if not (nodedef and nodedef.groups and
-			nodedef.groups.igniter and nodedef.groups.igniter > 0) then
-		minetest.sound_play(
-			"default_cool_lava",
-			{pos = pos, max_hear_distance = 16, gain = 0.1},
-			true
-		)
+function default.register_torch(name, def)
+	def.drawtype = "mesh"
```
sfan5: def should be copied with `table.copy` before making changes

mods/default/torch.lua
```diff
@@ -3,152 +3,116 @@
...
+	def.paramtype = "light"
+	def.paramtype2 = "wallmounted"
+	def.sunlight_propagates = true
+	def.walkable = false
+	def.liquids_pointable = false
```
sfan5: can be removed, this is the default

mods/default/torch.lua
```diff
@@ -3,152 +3,116 @@
...
-	on_flood = on_flood,
-	on_rotate = false
-})
+	local torch_wall = table.copy(def)
+	torch_wall.mesh = "torch_wall.obj"
+	torch_wall.selection_box = {
+		type = "wallmounted",
+		wall_side = {-1/2, -1/2, -1/8, -1/8, 1/8, 1/8}
+	}
+	torch_wall.groups.not_in_creative_inventory = 1
```
sfan5: since groups is optional according to the documentation, there needs to be `if torch_wall.groups == nil then torch_wall.groups = {} end` here

mods/default/torch.lua
```diff
@@ -3,152 +3,116 @@
...
 
-minetest.register_node("default:torch_ceiling", {
-	drawtype = "mesh",
-	mesh = "torch_ceiling.obj",
+default.register_torch("default:torch", {
+	description = S("Torch"),
 	tiles = {{
-		    name = "default_torch_on_floor_animated.png",
-		    animation = {type = "vertical_frames", aspect_w = 16, aspect_h = 16, length = 3.3}
+		name = "default_torch_mesh_animated.png",
```
sfan5: is there a reason the names were changed?
It would be better to keep the old names so texture packs don't have to update.

MoNTE48: https://github.com/minetest/minetest_game/pull/2436#issuecomment-569530941

mods/default/torch.lua
MoNTE48: What does that make sense?

mods/default/torch.lua
MoNTE48: Maybe better is `def.groups = def.groups or {}`?

sfan5: Sure.

mods/default/torch.lua
sfan5: You don't want to pollute the caller's `def` table with changes, e.g. if it is reused.

MoNTE48:
@sfan5 please check the changes

MoNTE48:
Ping XD

paramat:
> I think it would be better if this was moved to a 'torch_api' mod that modders could copy into their games and use for the easy registration of torches. The actual default:torch registration would stay in the default mod, so the torch API would have no dependencies

Yes, this should certainly be done. This follows the principle of making MTGame as modular as possible, and separates out the API from the actual registration of torches. Separating API from registrations is already being proposed in other places in MTG.

paramat:
Any core dev concept support? I know core devs have reviewed, but that does not necessarily mean concept support.

MoNTE48:
I decided not to wait for the anniversary of the commit. Thanks to Core Devs for their good attitude towards contributors.

PR #2443, opened 2019-08-14
add steel bar doors and trapdoors

TumeniNodes:
New PR for https://github.com/minetest/minetest_game/pull/2442

~I decided to use default:tin_ingots in the recipes because,
it provides a simple solution to a recipe conflict with steel doors and steel trapdoors,
it is a legit process used to coat steel with tin to reduce corrosion (in cans)~

~citations:~
~>A "tin can" used to store food. Most cans used for this purpose are now tin-plated steel.~

~>So tin may be unassuming, but it's not unimportant. This metal is used to prevent corrosion
https://www.livescience.com/37355-tin.html~

~(I feel it is a perfect and viable solution and now xpanes recipes do not require any modification)
It also provides another use for tin, which now justifies the addition of tin even more.
Profit ;)~
*Switched to just use default:steel_ingots per https://github.com/minetest/minetest_game/pull/2443#issuecomment-521336622

I added a handle to the trapdoors to distinguish from xpanes:bar_flat (though these are not seen in the wield condition, only in inv. and ingame

I also took the liberty of making changes requested regarding indentations but, I went ahead and did the same for all of the existing doors' code

screenshot_20190814_002804
screenshot_20190813_012142


DS-Minetest:
Ugh, the recipe seems so bad.
I don't know how tin coating is used exactly against corrosion, but in mtg there's never anything done against corrosion. And making steel (or iron?) bars out of some steel and tin seems unintuitive. And the xpanes bars and the doors should be made out of the same material.
I still want back the recipes using xpanes. Changing the shape would be weird and unintuitive for players. Also with xpanes the chances for recipe overlapping is lower.
(Btw. git isn't that difficult to use.)

TumeniNodes:
I would like them to be made using the xpanes too but, I agree it seems overkill to add xpanes as a 2d hard dep for doors.
And I definitely do not feel it is a good idea to register these doors within xpanes itself.

I do not feel these new recipes are _that_ bad, and they will become intuitive as players begin to use them.
It's not really that dramatic of a problem. At least I don't believe it is.

And git _can_ become a bit confusing in certain situations for novice users and/or individuals with memory issues (like myself)

DS-Minetest:
Don't add a hard dep, add a soft optional one and only register the recipes if the mod is installed.

TumeniNodes:
> Don't add a hard dep, add a soft optional one and only register the recipes if the mod is installed.

hmmmm, tbh I hadn't even thought of that.
actually the doors themselves should only be reg if xpanes is loaded

I'll do that when I return from appt

benrob0329:
Perhaps we should seperate the doors API from the door definitions themselves so that extra deps don't screw with games trying to be compatible.

paramat:
Using tin in the recipes is a good solution and gives tin another use too.
> I also took the liberty of making changes requested regarding indentations but, I went ahead and did the same for all of the existing doors' code

That's fine.

TumeniNodes:
Well if we use shaped recipes we can just move the steel ingots to the right for the bar doors and to the left for the solid steel doors?

TumeniNodes:
New recipes added. Hope these are acceptable and do not conflict with any others out there

Thomas--S:
I'd suggest to add a steel stick (similiar to those of the [darkage mod](https://bitbucket.org/adrido/darkage/src/19f2fa743590c4b7bea2329015dd68e7071d40ee/init.lua#lines-41) or the [basic_materials mod](https://gitlab.com/VanessaE/basic_materials/blob/master/metals.lua#L33) to Minetest Game.

It could be used as an ingredient for xpanes' steel bars as well as for the steel bar doors here.
Another advantage would be that mods won't have to define their own steel sticks anymore, but they can use Minetest Game's one, so the duplication of item definitions can be avoided more easily.

paramat:
The new recipes seem reasonable and similar to what i was considering.

TumeniNodes:
> I'd suggest to add a steel stick (similiar to those of the [darkage mod](https://bitbucket.org/adrido/darkage/src/19f2fa743590c4b7bea2329015dd68e7071d40ee/init.lua#lines-41) or the [basic_materials mod](https://gitlab.com/VanessaE/basic_materials/blob/master/metals.lua#L33)) to Minetest Game.
> 
> It could be used as an ingredient for xpanes' steel bars as well as for the steel bar doors here.
> Another advantage would be that mods won't have to define their own steel sticks anymore, but they can use Minetest Game's one, so the duplication of item definitions can be avoided more easily.

I don't see this as a viable reason to add a steel stick (which I would call a steel_rod)
But to provide a material so 3rd party mods don't need to provide it for a specific use they have?
If this were done, then another similar step would be suggested and, where does it end?

The way to avoid duplication among 3rd party mods, is for those mods to make use of existing items from existing mods, if they require them.
But then we have a ton of 3rd party mods all dependent upon on another (a mess)
So sometimes duplication is unavoidable and ... "ok" ;)

TumeniNodes:
> The new recipes seem reasonable and similar to what i was considering.

Wonderful. Does that warrant an approval? :D

TumeniNodes:
@paramat I'm thinking of changing the recipe for the doors to use the outer two rows of the grid vertically instead, to kinda/sorta resemble making upright bars.
Thoughts on that? I think it makes a bit more sense.

```
{"x", "", "x"},
{"x", "", "x"},
{"x", "", "x"}

Thomas--S:
I can understand to some degree that you want to avoid new items.

However, I think that adding recipes like the ones suggested in this PR might cause some problems:
Firstly, they are not very intuitive because they have a different shape than all other (trap-)door recipes and secondly, the recipes will very likely break the recipes of mods where these shapes make more sense.

May I ask you to reconsider adding steel sticks (or however you wish to call them), please?
They would allow to use the recipes the players are already used to for crafting (trap-)doors.
I think they would also help quite a bit with avoiding duplicate recipes.

TumeniNodes:
Honestly I do not oppose adding steel rods but that is up to core devs.
If it is agreed I will add them

If not, this can stand as is and with proper documentation the recipes for these doors will become intuitive over a short period of time

It is not that big of an issue to have a diff recipe layout for a new and specialized object

paramat:
> I'm thinking of changing the recipe for the doors to use the outer two rows of the grid vertically instead, to kinda/sorta resemble making upright bars.
Thoughts on that? I think it makes a bit more sense.

Yes excellent, please do, more intuitive.
I haven't approved this PR, i have no objections but just feel very neutral, i expect other core devs will support it.

I don't think steel sticks/rods should be done, that's far worse than slightly weird recipes.

Thomas--S:
> I don't think steel sticks/rods should be added, that's far worse than slightly weird recipes.

Please excuse me for asking again, but could you please explain this in a little bit more detail? I don't understand how adding the new item is "far worse than slightly weird recipes".

When we "block" rather common recipes with _slightly weird_ ones, where we could easily use a more intuitive one, it will be very difficult for mods to find replacements for their "useful" recipes that rely on the same shapes.

DS-Minetest:
Imo, all (trap)doors recipes should base on the same shape and should only differ in the material they use.
I still can't understand why we can't have the recipe with the xpanes as materials (with optional dependencies). It would be by far the most intuitive.
The `⋮ ⋮` recipe however, wouldn't resemble a door, it's just two separate rods.

TumeniNodes:
@SmallJoker 
@sfan5 
@rubenwardy 
@Ezhh 
@sofar 
Would appreciate your thoughts, and reviews

SmallJoker:
My suggestions
1) Optional-depend on `xpanes`
2) Overwrite or duplicate `xpanes:bar_flat` as trapdoor (I'd prefer former)
3) 3x Steel Bars + Steel Ingot for a locked steel bar door

paramat:
Here's an alternative i'm ok with:

Register the bar doors in xpanes mod, crafted from xpanes bars.
After all, registering items should ideally be done in a different place to the API for those items, this is something we're already considering.
Optionally registering these bar doors in doors mod but crafted from xpanes items and with an optional dependency on xpanes is a very messy implementation, i'm -1 for that.
SmallJoker's first 2 suggestions are messy too, i'm -1 for those.
The bar door (which we are making lockable) being 5 xpanes bars + a steel ingot (for the lock) is a good idea.

TumeniNodes:
If an agreement can be made between at least two devs, I will make the adjustments agreed upon.
Until then I can't do anything.

I do feel using xpanes, as I originally had it, makes sense... now it just depends on "where" to register these doors, and the recipes.
I'm glad to do the work, but just need to know for certain which work needs to be done.

btw, new texture for the door itself (with lock)
doors_door_bar

and in-game
screenshot_20190818_160727

DS-Minetest:
>     2. Overwrite or duplicate `xpanes:bar_flat` as trapdoor (I'd prefer former)

What exactly do you mean with this? There is no such thing as `xpanes:bar_flat`.

TumeniNodes:
xpanes:bar_flat is what they are called

https://github.com/minetest/minetest_game/blob/master/mods/xpanes/init.lua#L144

TumeniNodes:
rebased

TumeniNodes:
@paramat I have another alternative to propose...
Register the new steel bar door and trapdoor in doors, with xpanes soft dep with this PR,
and when the time comes to separate doors reg from the api... _then_ we address where to place the steel bar doors reg.

Does that sound reasonable?

paramat:
> Register the new steel bar door and trapdoor in doors, with xpanes soft dep with this PR,
and when the time comes to separate doors reg from the api... then we address where to place the steel bar doors reg. (perhaps in xpanes as you suggest)

I can't agree to that because it is doing something in a bad way only because it might (or actually might not) be put right later. It has to be right at all times.

I won't disapprove this PR as long as it is done the right way, but my personal feeling is that this is a nice but low-priority feature, and i disagree with adding low-priority features to MTG, because we need to move on and time is better spent on new games. Also because i disagree with increasing the maintenance requirements of MTG without very good reason.

I think our extremely limited time is only worth spending on MTG if it is maintenance or bugfixing, or an occasional high-priority feature. I can't justify spending my time on anything else. I spent far too much time on the snow levelling PR, which is also low-priority.

I would prefer these doors are provided by an optional mod instead of being in MTG.

TumeniNodes:
I can reg them in xpanes, as Ive said I dont use crafting.
Ill pop the reg there and if it sits to long Ill just provide it as a mod, this hasnt gotten much notice or attention/interest anyway I just felt it would be nice to offer a diff door ingame

sfan5:
When rebasing, please also add `S()` for translations.

TumeniNodes:
will attempt to rebase shortly (it was throwing an error at me previously)

TumeniNodes:
sorry, I typed in a wrong character when rebasing and it caused a mess.
Closing this and the new PR is https://github.com/minetest/minetest_game/pull/2490

PR #2449, opened 2019-08-20
add_new_tnt_sounds

TumeniNodes:
Sounds edited for tnt mod.
I have to dig to find out the license for tnt_gunpowder_burning.ogg though
it was not removed with the other two.
Do we know where that file is from? Or does it need to be replaced as well, because I have a file all set if it does.

tweaks avail upon request, but these are better than the current temps.

HybridDog edited the tnt_explode file, I just cut it down a bit more
I cut the tnt_ignite sample down to play just until the exact moment tnt_explode kicks in
I also used ffmpeg to reduce the file sizes


sfan5:
> I also used ffmpeg to reduce the file sizes

Which command line did you use?

TumeniNodes:
-c:a libvorbis -q:a 1 <output>.ogg

TumeniNodes:
Hold on.. I've looked at the commands a bit more
I think I found some misinformation before. I'm going to try something else

HybridDog:
You can use the `oggenc` command instead of ffmpeg.

TumeniNodes:
the actual cl I needed was
ffmpeg -i 'file' -c:a libvorbis -q:a 0 -ab 112 output.ogg

>You can use the oggenc command instead of ffmpeg.

I've never used it before

sfan5:
`ignite`: too quiet/muffled, too long, can't clearly hear it igniting unlike the previous sound
`explode`: too loud, there seems to be some clipping

TumeniNodes:
> `ignite`: too quiet/muffled, too long, can't clearly hear it igniting unlike the previous sound
> `explode`: too loud, there seems to be some clipping

I'll work on these.
explode I just increased the sound to the same level as the original because it was too quiet. The clipping at the end is far less noticeable than the original, I'm trying to figure out how to use a fadeout with it (it's not easy)

I just replace gunpowder burning as well after a brief conv with sofar, it still needs to be edited/cut down

HybridDog:
Maybe @SparkizX can help with the sounds.

paramat:
The license text should have lines wrapped at 80-90 columns.
Easiest way is to copy-paste the CC0 text from another mod
https://github.com/minetest/minetest_game/blob/3a3f71aa50c7ef5d894d1f1e0760f164654bbeb7/mods/default/license.txt#L122

mods/tnt/license.txt
```diff
@@ -64,3 +64,45 @@ rights may limit how you use the material.
 
 For more details:
 http://creativecommons.org/licenses/by-sa/3.0/
+
+
+Other licensing used for media
```
paramat: No need for this as this text also comes under 'Licenses of media'.

paramat:
The explode sound does sound 'clipped', as in waveform clipping from excessive amplitude. HybridDog's version didn't sound clipped.

paramat:
The best way to fade out is use a long fadeout that starts soon after the initial explosion, this will be less noticeable than a short fadeout right at the end.
HybridDog's version, which i think is suitable, does need shortening certainly.

paramat:
Looking at HybridDog's version, the waveform is already maximised in amplitude so it can't be made louder, this is why it is now 'clipped'.

paramat:
I agree the ignite sound is far too muffled.

TumeniNodes:
I added an ignite sound to the start of tnt_iginte but I cut the file a tad too short, will redo when I am back at my pc.

These will all take a day or so to get just right, but theyll get there

TumeniNodes:
If you can all test out the new `ignite` and `explode` sounds please.
I have heavily edited both.
Pay no attention to the extra files, they will be deleted once this PR is finished. It is all still very WIP

mods/tnt/license.txt
TumeniNodes: > No need for this as this text also comes under 'Licenses of media'.

line removed

paramat:
I think the new sound seems quieter because it has much more dynamic range than the old sound. If Audacity has a 'compressor' effect that might help.

TumeniNodes:
At least it's much cleaner now.

The edit that was originally done had clipping at the very beginning.
I managed to get rid of it all.
When working with the some sound files you have to alternate between a couple different effects, and continuously tweaking the settings.
I'll get it there... eventually.

I'm interested in what people think of the sounds themselves so far.
I think the new `ignite` is much better. And I like the way it fizzles _right_ before `explode` kicks in... it kind of builds up a tension, like in a thriller where the music stops one or two seconds before a sudden/violent event. So I feel it works perfect.

That is what I'm trying to do first... is to just get the right sounds, and then all the decibels can be adjusted (I hope)
Audio files are definitely more involved labor wise, which is why I'm just focusing on having the right ones to work with first, and why as much input from people as possible is needed along the way.

TumeniNodes:
Everything cleaned up, edited, etc..
Hopefully all is good now
I think they all sound much better than the originals (but of course I'm bias)

LoneWolfHT:
The debris particles are still there, and they sound a little weird when listening to the sound...

TumeniNodes:
> The debris particles are still there, and they sound a little weird when listening to the sound...

What are you talking about? All of the trailing debris falling in the sound was removed.
There is one fraction of a second of it within the explosion sound itself, which is hardly noticeable and exists in real explosions anyway

LoneWolfHT:
I'm playing the sound on a loop, so it's noticeable pretty fast. But if you aren't able to remove it then I guess it'll do
Hmmmm, when the sound is at a lower volume I don't hear it anymore. But the explosion sounds fine. So this might work :shrug:

TumeniNodes:
I'd suggest and even recommend listing/testing it in-game

paramat:
A few subtle debris sounds are ok.

paramat:
Explode and ignite sound fine to me, more than good enough to merge, will check gain levels in-game.

TumeniNodes:
`gunpowder_burning` _might_  use a little lower db but, I've experience with it IRL and it does burn loud-ish

paramat:
Gunpowder sound also sounds suitable.

TumeniNodes:
While I was talking briefly about this with sofar on irc, it dawned on me that there are countless games out there which use those original sounds.
It was enough for him to even go in and double check ITB's code to be safe...
Should we be making an attempt to get the word out that they should be replaced? Or is the project covered just by replacing them now?

This is something which may definitely be in need of a discussion between devs
It was a mere oversight back a few years ago, but now that it has been found/noticed, how much obligation is there on the project to at least let people know, even with a brief announcement, and then from there it falls upon individual game/server creators?

I'm not a lawyer so idk, just thought it worth mentioning

TumeniNodes:
Well, this is ready to go...

paramat:
I'll inform people in a forum news thread, but won't mention the source of the sounds.

TumeniNodes:
I realize people are quite busy with engine work but this could use some reviews, and hopefully get this little issue out of the way

paramat:
Tested in game, all sound levels are good. Sounds are good, much better than before.
:+1:

TumeniNodes:
> Tested in game, all sound levels are good. Sounds are good, much better than before. Code is fine.
> +1

Thank you

TumeniNodes:
@sfan5 I know you've been busy, but if you could have another look over this please?

paramat:
LOL.

sfan5:
I don't know if you're joking or not
> I'll merge it within 24hrs, 2 approvals or not, it's obviously fine so i'll take the initiative

No it does absolutely not work like this.
MTG is not your pet project (even though you act like it) and it is not okay to break the rules as you consider it useful (even though you did it before).

sfan5:
Feedback on the sounds:

`ignite`: not a fan of how long it is but that's a matter of taste, it's fine :+1:
`explode`: the clipping is still there
I wouldn't say its bad enough that it <i>needs</i> to be fixed, but I certainly notice it every time the sound plays
audacity

paramat:
SmallJoker indicates support for a merge http://irc.minetest.net/minetest-dev/2019-08-27#i_5583353 so i will merge it as sfan5 doesn't disapprove.

The clipping isn't actually noticeable to me when listening, even though visually the waveform is clipped, this surprised me.

sfan5 i'm not joking.
In this situation of having no TNT sounds, even if there is some sort of problem with this PR it could be considered temporary or 'to be improved in a follow up PR'. Merging it causes absolutely no harm to anyone or the project, and it's better than the author rage-quitting and we lose all this good work.

This is why, in this particular situation, i judged it ok to merge. I will not do this for all PRs obviously.
Please realise how this is reasonable instead of being pointlessly rigid with the rules. Most of the other core devs are more flexible with the rules and often take the initiative. You're being too strict and inflexible,

jastevenson303:
Clipping can't be justified.  That's just elementary.

HybridDog:
Audacity supports Soft Clipping (Effect->Distortion...) and other effects which make the sound louder without hard clipping.

HybridDog:
> […] I'm not sure what the difference would be?

I didn't link to an explanation before. It's there: https://manual.audacityteam.org/man/limiter.html

sfan5:
> Paramat + SmallJoker - sfan5 = 1

Note that the approval rules don't work like that, even when a coredev gives explicit disapproval.
Also, I did say that I do not disapprove of the new sounds based on the issue with clipping.

paramat:
I want to apologise for my earlier intent to merge without 2 approvals, that was a mistake. I should not have responded that way to a threat of PR closure.
Despite what i wrote, if SmallJoker had not supported a merge on IRC i probably would not have merged this.

TumeniNodes, please try to not threaten closure of PRs in future. Because our time is limited it won't make a review happen, and in this case a PR closure would have wasted a large amount of your own time and all the time core devs spent reviewing.
If you get impatient waiting for review, the best thing to do is forget about the PR and move on to something new.

SmallJoker:
The current sounds are OK. Some tweaking would surely not be wrong, either.

TumeniNodes:
>TumeniNodes, please try to not threaten closure of PRs in future.

I was feeling bitchy, happens to the best of us. Plus I felt this particular issue was a bit of a priority and should have gotten a bit more attention. It may have been easier and quicker to have come to better editing with more input from you guys.
I am just as busy as others. It involved no complicated code to review... just merely a few moments to listen to 3 sounds, and give some feedback. Which was all I was looking for
Water under the bridge... let's move on please : )

>The current sounds are OK. Some tweaking would surely not be wrong, either.

They're awesome ; )  (but yes, some further tweaking will be done eventually) I _will_ eventually get `explosion` cleaner and I want to make `gunpowder_burning` smoother during the transition from one to the next.

PR #2471, opened 2019-09-10
Fix and improve translation strings

sfan5:
Notable:
* breaks stair API compatibility on two occasions:
  * register_stair_inner, register_stair_outer no longer prepend "Inner ", "Outer " to description (unavoidable)
  * register_stair_and_slab uses the "stairs" translator on the given description (could be avoided)
* cart & screwdriver description were changed to use newline for consistency

### DIFF
--- a/mods/butterflies/init.lua
+++ b/mods/butterflies/init.lua
@@ -5,9 +5,9 @@ local S = minetest.get_translator("butterflies")
 
 -- register butterflies
 local butter_list = {
-	{"white", S("White")},
-	{"red", S("Red")},
-	{"violet", S("Violet")}
+	{"white",  "White"},
+	{"red",    "Red"},
+	{"violet", "Violet"}
 }
 
 for i in ipairs (butter_list) do
@@ -15,7 +15,7 @@ for i in ipairs (butter_list) do
 	local desc = butter_list[i][2]
 
 	minetest.register_node("butterflies:butterfly_"..name, {
-		description = S("@1 Butterfly", desc),
+		description = S(desc .. " Butterfly"),
 		drawtype = "plantlike",
 		tiles = {{
 			name = "butterflies_butterfly_"..name.."_animated.png",
@@ -61,7 +61,7 @@ for i in ipairs (butter_list) do
 	})
 
 	minetest.register_node("butterflies:hidden_butterfly_"..name, {
-		description = S("Hidden @1 Butterfly", desc),
+		description = S("Hidden " .. desc .. " Butterfly"),
 		drawtype = "airlike",
 		inventory_image = "insects_butterfly_"..name..".png",
 		wield_image =  "insects_butterfly_"..name..".png",
--- a/mods/wool/init.lua
+++ b/mods/wool/init.lua
@@ -9,7 +9,7 @@ for i = 1, #dyes do
 	local name, desc = unpack(dyes[i])
 
 	minetest.register_node("wool:" .. name, {
-		description = S("@1 Wool", desc),
+		description = S(desc .. " Wool"),
 		tiles = {"wool_" .. name .. ".png"},
 		is_ground_content = false,
 		groups = {snappy = 2, choppy = 2, oddly_breakable_by_hand = 3,
--- a/mods/binoculars/init.lua
+++ b/mods/binoculars/init.lua
@@ -59,7 +59,7 @@ minetest.after(4.7, cyclic_update)
 -- Binoculars item
 
 minetest.register_craftitem("binoculars:binoculars", {
-	description = S("Binoculars\nUse with 'Zoom' key"),
+	description = S("Binoculars") .. "\n" .. S("Use with 'Zoom' key"),
 	inventory_image = "binoculars_binoculars.png",
 	stack_max = 1,
 
--- a/mods/boats/init.lua
+++ b/mods/boats/init.lua
@@ -154,13 +154,13 @@ function boat.on_step(self, dtime)
 			if ctrl.up and ctrl.down then
 				if not self.auto then
 					self.auto = true
-					minetest.chat_send_player(self.driver, S("[boats] Cruise on"))
+					minetest.chat_send_player(self.driver, S("Boat cruise mode on"))
 				end
 			elseif ctrl.down then
 				self.v = self.v - dtime * 1.8
 				if self.auto then
 					self.auto = false
-					minetest.chat_send_player(self.driver, S("[boats] Cruise off"))
+					minetest.chat_send_player(self.driver, S("Boat cruise mode off"))
 				end
 			elseif ctrl.up or self.auto then
 				self.v = self.v + dtime * 1.8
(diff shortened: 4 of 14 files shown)


mods/butterflies/init.lua
```diff
@@ -5,17 +5,17 @@ local S = minetest.get_translator("butterflies")
...
+	{"violet", "Violet"}
 }
 
 for i in ipairs (butter_list) do
 	local name = butter_list[i][1]
 	local desc = butter_list[i][2]
 
 	minetest.register_node("butterflies:butterfly_"..name, {
-		description = S("@1 Butterfly", desc),
+		description = S(desc .. " Butterfly"),
```
SmallJoker: The translation script cannot detect this.

mods/butterflies/init.lua
```diff
@@ -61,7 +61,7 @@ for i in ipairs (butter_list) do
 	})
 
 	minetest.register_node("butterflies:hidden_butterfly_"..name, {
-		description = S("Hidden @1 Butterfly", desc),
+		description = S("Hidden " .. desc .. " Butterfly"),
```
SmallJoker: Neither this dynamic text.

mods/wool/init.lua
```diff
@@ -9,7 +9,7 @@ for i = 1, #dyes do
 	local name, desc = unpack(dyes[i])
 
 	minetest.register_node("wool:" .. name, {
-		description = S("@1 Wool", desc),
+		description = S(desc .. " Wool"),
```
SmallJoker: Dynamic name again. Consider inserting dummy gettext lines like builtin; to ensure it's parsed.

sfan5: I'm aware, can you link me to a source file where builtin does this?

Wuzzy2:
:-1: :-1: :-1: :-(

Pretty much all changes that are not trivial are wrong:

> 		description = S(desc .. " Butterfly"),
> 		description = S("Hidden " .. desc .. " Butterfly"),
> 		description = S(description .. " Dye"),
>               description = S(desc .. " Wool"),

Ugh. No! That's the mortal sin of translation: String concatenation (in the middle of something). Now you force all languages to use the same word order as English. This is wrong and even worse than the previous approach. The correct way would have been to collect each string “White Butterfly”, “Red Butterfly” and “Violet Butterfly” individually, in their FULL form. That way, translators can apply whatever weird grammar rules would change the words and don't have to hack around. If you try to be clever and basically construct the strings programmatically, you make assumptions about other languages and make some translations simply impossible. The word “Butterfly” might not always be translated the same depending on weird grammar rules in other languages.

>        {"white",  "White"},
>        {"red",    "Red"},
>        {"violet", "Violet"}

And now, the butterfly color will not even be collected by scripts …

A proper way would have been:

>        {"white",  S("White Butterfly")},
>        {"red",    S("Red Butterfly")},
>        {"violet", S("Violet Buterfly")}
>
>        -- ...
>
>         local desc = butter_list[i][2]
>         minetest.register_node("butterflies:butterfly_"..name, {
> 		description = desc,

The same complaints apply for the other string concatenations …


```
+       stairs.register_stair_inner(subname, recipeitem, groups, images,
+              S("Inner " .. desc_stair), sounds, worldaligntex)
```

```
-       S("Wooden Stair"),
-       S("Wooden Slab"),
+       "Wooden Stair",
+       "Wooden Slab",
```

Nooooo! Whyyyyy? That's not how it works! This translation will not work because …

1. Adding a variable inside S() will mean that translation tools to generate the template (like `findtext.lua`) will fail. For that reason, text inside S() should always be constant. Or did you intend to just write the template by hand? :)
2. You removed the S() for the base names "Wooden Stair" and "Wooden Slab" and all other slabs and stairs. This means these string will also not be collected by tools, and no proper template can be generated …
3. It's just another string concatenation

> breaks stair API compatibility on two occasions: 

Sigh. :-1:

> register_stair_inner, register_stair_outer no longer prepend "Inner ", "Outer " to description (unavoidable)

It's completely avoidable. Just add 2 more optional arguments for explicitly specifying the full inner and outer node description. If missing, fall back to the old behaviour (i.e. prepend "Inner"/"Outer"). That's how I did it in MCL2. And obviously, mods are still strongly adviced to provide ALL arguments, but things won't break when they don't.

> register_stair_and_slab uses the "stairs" translator on the given description (could be avoided)

Yes, it could be avoided. Just don't use the S() in the actual stair API anywhere. Instead, mods that actually call the stair register function are supposed to use the S(). That's how I did it in MCL2.

sfan5:
> Ugh. No! [...]

Great job writing such a negative review when you have no idea what you are talking about.

`S(desc .. " Butterfly")` ends up as `S("Red Butterfly")` which can be properly translated in any word order (e.g. `Red Butterfly=Butterfly but with a different order (red)`), try it yourself.

> And now, the butterfly color will not even be collected by scripts …
> The same complaints apply for the other string concatenations …

I'm aware of this limitation, I explicitly preferred doing this instead of excessively repeating strings everywhere.
The collection could be solved using a dummy file with the correct `S()` invocations, but I don't know if others would be fine with that.

> It's completely avoidable. Just add 2 more optional arguments for explicitly specifying the full inner and outer node description.

That's a good idea, I'll do that.


> (like `findtext.lua`) 

By the way, where do I find this tool?

Wuzzy2:
> S(desc .. " Butterfly") ends up as S("Red Butterfly") which can be properly translated in any word order (e.g. Red Butterfly=Butterfly but with a different order (red)), try it yourself.

In theory, yes. In practice, no, because the script won't see it.

> The collection could be solved using a dummy file with the correct S() invocations, but I don't know if others would be fine with that.

Well, but then you just wrote the exact string you wanted to avoid … I don't see the benefit.

> where do I find

Well, the script is not very good and is not finished and lacks features. And you can find multiple versions of it floating around GitHub. Here's one of the versions:

https://github.com/Ganome/OPSkyBlock/tree/7127da760c85ed56f4a128b744aa3e4d004c64a7/mods/intllib/tools

We really need proper and OFFICIAL tools to collect strings. But that's another issue, unrelated to MTG.

sfan5:
> Well, but then you just wrote the exact string you wanted to avoid … I don't see the benefit.

True, the actual benefit would be that the code looks cleaner, not that you don't have to type the strings out.

mods/wool/init.lua
Panquesito7: Just a note: I've tested this way before and it doesn't translates the string.
This should be reverted.

sfan5: works for me:
png

Panquesito7: Nice! 👍

SmallJoker:
The script that I used for my client-side translations: #6325, improved: https://notabug.org/pgimeno/minetest/src/translation-toolchain/util/findtext.lua

Wuzzy2:
This looks fine now. I am happy with the changes.

PR #2487, opened 2019-09-17
Fix more translation strings

### LINKED ISSUE #2474: More translation bugs
Translation work has been going quite well so far, but it's not 100% perfect yet. Here are some more bugs or missing strings I have found:

* Tooltips in creative inventory are currently broken in all languages:
   * Search, Next page, Previous page, Reset
   * The tooltip displayed is literally like “`S("Search")`” (that's right, the S() appears in the tooltip)
* Inner Straw Stair and Outer Straw Stair are not translatable
* Wheat and Cotton items are not translatable
* Crafting tab in inventory menu is not translatable

About the Wheat and Cotton items: Looks like these item descriptions are constructed automatically in `farming` by deriving it from the itemstring …

```
326         -- Register harvest
327         minetest.register_craftitem(":" .. mname .. ":" .. pname, {
328                 description = pname:gsub("^%l", string.upper),
329                 inventory_image = mname .. "_" .. pname .. ".png",
330                 groups = def.groups or {flammable = 2},
331         })
```

That's really hacky IMO. Apart from being problematic for translations, it also makes it hard for mods to set an arbitrary item description for the harvest item. I strongly suggest to fix the `farming` API to allow setting an explicit description for the harvest item, then immediately make use of this for the Cotton and Wheat items, and mark it as strongly discouraged in the API documentation to not set the harvest description.
The automatic description generator should be kept as fallback, but only for legacy purposes.

Version: e4adb01fbf5d036383ec1aab545f14d353f5af7c

sfan5:
fixes #2474

### DIFF
--- a/game_api.txt
+++ b/game_api.txt
@@ -359,6 +359,8 @@ The farming API allows you to easily register plants and hoes.
 
 	{
 		description = "",                      -- Description of seed item
+		harvest_description = "",              -- Description of harvest item
+		                                       -- (optional, derived automatically if not provided)
 		inventory_image = "unknown_item.png",  -- Image to be used as seed's wield- and inventory image
 		steps = 8,                             -- How many steps the plant has to grow, until it can be harvested
 		-- ^ Always provide a plant texture for each step, format: modname_plantname_i.png (i = stepnumber)
--- a/mods/beds/functions.lua
+++ b/mods/beds/functions.lua
@@ -5,6 +5,9 @@ if enable_respawn == nil then
 	enable_respawn = true
 end
 
+-- support for MT game translation.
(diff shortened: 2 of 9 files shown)


Wuzzy2:
Looks OK, except for:

```
			"field[0.5,1;7.5,0;title;" .. esc(S("Title")) .. ":;" ..
				esc(title) .. "]" ..
			"textarea[0.5,1.5;7.5,7;text;" .. esc(S("Contents")) .. ":;" ..
				esc(text) .. "]" ..
```

The colon should be included in the translation.

Wuzzy2:
:+1: :+1: :+1:

PR #2497, opened 2019-09-24
xpanes add new steel bar door sounds

TumeniNodes:
Suggested by @Wuzzy2 to change the open/close sounds for the new steel bar doors
https://github.com/minetest/minetest_game/issues/2495

I definitely agree the current steel door sounds via default do not fit the character of the new steel bar doors at all.
So these are the sounds I have created/produced for those new doors.


Wuzzy2:
^^

paramat:
They're more suitable than current sounds i agree.

TumeniNodes:
I had mentioned possibly changing these sounds a bit but, I'm going to stick with these. They work well and I like them.

paramat:
TumeniNodes are you happy with the sounds? It would be good to merge the intended sounds for 5.1.0.

TumeniNodes:
> TumeniNodes are you happy with the sounds? It would be good to merge the intended sounds for 5.1.0.

Yes I am.
I was unsure at first but after using them for a while and even some feed back from a couple people outside of the community, I am happy with them.

paramat:
Screenshot from 2019-09-27 18-32-01

The close sound is fine, but i'm not happy with the open sound. Looks like it is 2 bits of sound heavily edited and joined, the first sound ending abruptly (see image). This gives it a very abstract 'creak-creak' sound.
The opening creaking sound has potential, just not as heavily edited/processed as this.

TumeniNodes:
I made some more adjustments, try this (it definitely sounds and works better)

paramat:
Ooh this has potential.

paramat:
:+1: Very good sounds when tried in-game, levels are good too.

TumeniNodes:
Thank you

TumeniNodes:
Thank you

TumeniNodes:
Thank you both

PR #2514, opened 2019-10-04
Add French translation

DrHackberry:
I have just created a complete French translation.

Thank you in advance for your reviews.


mods/default/locale/default.fr.tr
```diff
@@ -0,0 +1,211 @@
+# textdomain: default
+Locked Chest=Coffre verrouillé
+Locked Chest (owned by @1)=Coffre verrouillé (possédé par @1)
+You do not own this chest.=Ce coffre ne vous appartient pas.
+a locked chest=Un coffre verrouillé
```
Calinou: Is the uppercase "U" intended here? The original string starts with a lowercase letter.

mods/default/locale/default.fr.tr
```diff
@@ -0,0 +1,211 @@
...
+Mese Crystal=Cristal de mese
+Gold Lump=Morceau d'or
+Diamond=Diamant
+Clay Lump=Morceau d'argile
+Steel Ingot=Lingot d'acier
+Copper Ingot=Lingot de cuivre
+Tin Ingot=Lingot d'étain
+Bronze Ingot=Lingot de bronze
+Gold Ingot=Lingot d'or
+Mese Crystal Fragment=Fragment de cristal de mese
```
Calinou: I think the consensus is to use "Mese" as a proper noun, which means it should be capitalized.

Make sure to apply this change in all strings where "mese" is written.

mods/tnt/locale/tnt.fr.tr
```diff
@@ -0,0 +1,4 @@
+# textdomain: tnt
+Gun Powder=Poudre à canon
+TNT Stick=Baton de TNT
```
Calinou: Missing diacritic:

```suggestion
TNT Stick=Bâton de TNT
```

mods/default/locale/default.fr.tr
DrHackberry: Indeed, it wasn't intended.

DrHackberry:
Thank you for your review Calinou 🙂

I didn't know the Git conventions, I updated all corresponding files and also took into account your suggestions. 

I also created a [topic](https://forum.minetest.net/viewtopic.php?f=20&t=23435&p=359055#p359055) in the French part of the forum in case some users would like to participate without using Git and/or Github.

PR #2529, opened 2019-10-16
Add a crosshair texture

Calinou:
Redesign of #401.

This makes the crosshair easier to see when pointing light surfaces. It also has a more compact profile.

Unlike #401, the black outline uses 50% alpha as to be less intense.

## Preview

Crosshair with dark background

Crosshair with medium background

Crosshair with light background


paramat:
crosshair128

This is very similar to the previous PR, the crosshair design is unchanged apart from the edge alpha, it is still very thick and visually obtrusive, which is more of a problem than it being hard to see against white surfaces. Selection boxes help in seeing what has been selected :-1:

paramat:
Yes i assumed that was the reason, but i do not think 0.5 screen pixels off is a problematic offset.
Your design is the way it is for good reason, but i find it visually very obtrusive. This is a difficult task.

Calinou:
@paramat Would it be OK if I made the white part slightly transparent (like 10-20%)?

paramat:
That wouldn't help in my opinion. The issue for me is that a wide crosshair with a dark border seems to float in front of the screen and is very distracting.
Personally i would try a 1-pixel wide cross consisting of alternating black/white pixels.

Here's a good idea that keeps the crosshair shape unchanged. but unfortunately might depend on shaders being enabled:
> kilbith
14:46 		better inverting the colors in a fragment shader for the crosshair
14:46 		you bind another RTT from the current frame and you pass it to the shader

PR #2563, opened 2020-01-24
Update Chinese translation

IFRFSX:
rail = 轨道
grass = 草
and other repair


mods/default/locale/default.zh_CN.tr
```diff
@@ -165,11 +165,11 @@ Acacia Wood Fence=相思木围栏
 Jungle Wood Fence=丛林木栅栏
 Pine Wood Fence=松木栅栏
 Aspen Wood Fence=白杨木栅栏
-Apple Wood Fence Rail=苹果木铁路
-Acacia Wood Fence Rail=相思木铁路
-Jungle Wood Fence Rail=丛林木铁路
+Apple Wood Fence Rail=苹果木轨道
```
sfan5: Actually this isn't a rail (transport) but a railing, which should be "欄杆" according to [wiktionary](https://en.wiktionary.org/wiki/railing#Translations).

PR #2603, opened 2020-02-24
fix xpanes connections

TumeniNodes:
This fixes the ugly squares and "x"s when xpanes are connected.
as seen here https://github.com/minetest/minetest_game/pull/2500#issuecomment-535486742

I realize this creates a situation where 3rd party TPs will need to create 2 new textures but this is not such a serious problem, given this fixes an old and ugly artifact.
Plus, it looks a little more realistic/believable where glass panes connect.
As a good gesture, I am offering to help as I can with other TPs to create these new textures for them, if needed.

Imo this produces a much nicer and more polished look to xpanes.

screenshot_20200223_220528

screenshot_20200223_220459

screenshot_20200223_220602


LoneWolfHT:
Not sure I like the new look 🤔
Is there a way to fix the visual bug without the extra texture?

TumeniNodes:
> Not sure I like the new look thinking
> Is there a way to fix the visual bug without the extra texture?

The "connected" nodebox requires a texture to reference.
As far as I have found, "this" is the best option which does not give an awkward artifact when connected. (trust me, I have played with this off and on back to I think 2018?) because the look when connected drives me nuts.

And you may not be to sure of it right now... but it will grow on you. (especially if/when it's the only option you have ;D )
Besides, it definitely looks much cleaner, nicer, and more "finished"

An0n3m0us:
@LoneWolfHT There's no other way to fix this (other than adding OBJ models); I tried in my PR already.

paramat:
The change might be a little controversial but having a frame at the corner does look better to me, and makes more sense than bent glass. So no objections from me.

TumeniNodes:
> However, you have altered the xpanes API, which might break a lot more than just texture packs, so this might be very disruptive.

Yes, I realized the implications of changing the [1] to [2] on line 140 will have...


> The change might be a little controversial visually but having a frame at the corner does look better to me, and makes more sense than bent glass.

But it seemed justifiable to me for the visual improvement, and sometimes some small breaking changes are need to move forward.
I don't believe this is that impacting that it should make it a deal breaker tbh.
Some very minor tweaking on 3rd party ends may be necessary, but nothing devastating.

LoneWolfHT:
This isn't an 'engine bug'?
The new look may grow on me a little but I will always prefer the previous look

TumeniNodes:
> This isn't an 'engine bug'?

No, it's simply a texture bug

LoneWolfHT:
But you can't fix the texture by changing the pixel causing the problem?

TumeniNodes:
I've been thinking and, perhaps this bug could be fixed using backface culling.
I'm gonna try it and if it works, I'll push the change

paramat:
Yes, the xpanes API is probably not heavily used, so although this will break some things it might be worthwhile, not sure, needs more inout.

TumeniNodes:
> needs more input.

Yeh, but we seem to be in one of those slow cycles atm, not much activity in the general community and quite a few regulars have been inactive for a while.

DS-Minetest:
Imo the current version looks better.

TumeniNodes:
Well, after a while, if this is not merged... I can always put it out there as an overide mod for those who may like this method/visual
So I'm not all that concerned.

It is only natural for people to become accustomed to something which has pretty much always been a default sound, visual, mechanic, etc., so I sort of expected the reception expressed so far.

An0n3m0us:
Considering that it fixes something and it looks good (I'm sure if I played Minetest, I'd get used to it)

:+1: on merging this PR. The only issue now is the new texture because of the compatibility with other texturepacks.

---

How about creating the texture by combining the glass textures we already have; like so:
```
default_obsidian_glass.png^[combine:16x16:-8,0=default_obsidian_glass.png:8,0=default_obsidian_glass.png
```

That way we won't have to create new textures ; )

Here's what it looks like ingame (I replaced the obsidian glass block texture):
kép

TumeniNodes:
I think this would put people off even more than just changing connecting texture.
It would cause a massive visual change for large structures using a great deal of these nodes, such as skyscrappers.

mods/xpanes/init.lua
```diff
@@ -137,7 +137,7 @@ function xpanes.register_pane(name, def)
 		is_ground_content = false,
 		sunlight_propagates = true,
 		description = def.description,
-		tiles = {def.textures[3], def.textures[3], def.textures[1]},
+		tiles = {def.textures[3], def.textures[3], def.textures[2]},
```
sfan5: game_api needs to be updated too

TumeniNodes: https://github.com/minetest/minetest_game/pull/2603/commits/7e9eeac465bd9f1cb756d189722bd0829610899b
I completely forgot about that, thank you (updated_

TumeniNodes:
It is at this point, safe to remove the `xpanes_pane_half.png` and `xpanes_space.png` from the textures folder.
These are completely unused, and have been for some time.
I will go ahead and remove these in this PR if desired, but not some devs agree.

TumeniNodes:
Very much definitely controversial.
I am wondering... if we may be able to make this an optional setting somehow? hmmm
have to think and look over a few things later

sfan5:
IMO having settings for too many behaviours in MTG isn't good, it should provide a mostly uniform experience out of the box.

Is it not possible to fix the ugly edge interconnects without adding a "bar" there?

TumeniNodes:
> Is it not possible to fix the ugly edge interconnects without adding a "bar" there?

The _only_ way I could think of, would be to use 3 separate nodes for the connection.
One at center which uses only no texture to represent the glass.

And one for each top and bottom. These two would only use the texture of the frame on all six sides.
But then you have to code three individual connecting nodes...

I mean, _something_ like that _might_ work but I am not positive.
I could do it but not sure how people feel about going that far for such a small mod?
let me know

TumeniNodes:
Later tonight, I will replace this node with a mesh, made of 3 parts (top/bottom, and middle) then I can set the UVs up so the middle section uses the glass of `default_glass.png` and the top/bottom use the frame of that texture.

This should work as I imagine only, I am not sure if it will work correctly if individuals use the `use_texture_alpha` for colored glass.
Only because the "alpha" may block seeing the frame texture for the top/bottom bits.

Does that make sense the way I'm explaining it?

Either way, I will get this in as I said, later tonight, and then people can check it out/review it to see what they think

TumeniNodes:
On thinking a bit more:
Will connected nodebox code even work with a mesh node? (I don't think it does)

If not, then this mesh idea is a bust.
And that leaves my only other idea, which is to make two individual node instances for this connection,
Where as now, each of those individual instances will need their own custom textures (like using the `xpanes_edge.png` for the two small top/bottom pieces, and another for the middle piece of this custom node)
Altogether now renders this nice, simple, little mod... "complicated"

These are all solutions I have thought on before, off and on over the past year or so...
This current PR, was the simplest, and most viable I could think of, which is why I decided finally to go with it and give it a shot, after the PR to fix the edges went through. (which I had already done as well a while ago in my own custom MTG version, that I play with)

So it really comes down to devs deciding which way they'd want to go, which I am assuming "to keep it simple" rather than to over complicate the mod.

My other idea, was to use an "if _this_ then _that_ / else _the other_" function
so the user can decide which to use visually.
But, that means adding yet another optional setting, which adds complication as well.

So I am more than happy to put in the time and work... but not without some direction on which way to attempt from the devs, (which gives more of a likelyhood it will/may be merged)

TumeniNodes:
Ok, drawtype = mesh does not work with connected type nodebox (already knew but tested anyway)
screenshot_20200303_162355

But, what you can see here (aside from the flipped normals causing a darker color) if this _had_ worked, these little top/bottom nodes would use the frame texture all around and fix that odd/ugly visual glitch.

My other optional idea, two use three individual nodeboxes, would be very complex, and still may not work... but would definitely fail once a user tried to use the colored glass options tied to this mod (xpanes) due to our ongoing alpha problems

So, here is the bottom line...
It's either **this** PR
else leave it as is and embrace the ugliness

An0n3m0us:
There's an easy way to fix this. It's quite obvious that the two lines of pixels in the center is causing this. They are both 2 pixel wide together.
So to fix it; one can increase the texture size to 32x32 and make the center line become 3 pixels wide. That way, they are now '3 pixels wide in total'. Here's what it looks like:

test

Here's the modified texture (`xpanes_connect.png`):

kép

The same thing can be done for the obsidian connect texture.

TumeniNodes:
Tbh, I feel it's better to fix this while sticking to the default texture size of 16x16

I'm unsure how what you show above is different than the current PR, aside from using a larger texture, and the fuller visual of the connecting node as the effect.
Although it does look nice, changing the texture size to fix a small visual bug, seems a bit out of scope.

I put this PR up knowing full well it may not be merged, and attempting to offer a fix to a long time visual bug.
It's not really a bug which many notice or complain about... but it's there, and has been.
Only recently has someone mentioned it, which provide evidence that yes, people know it's there, but it's not something which keeps them up at night.

As I said earlier, if this doesn't pass muster after a while, I can always release it as an override mod for those who prefer it.
No big deal.

I've thought it all through many times over the last year or more, this is what I came up with, and the only solution which has a minimal impact on the overall mod, itself.
Much like a code cleanup PR, this is a visual cleanup...

An0n3m0us:
Agreed;

If you look at the original texture; it requires a two pixels wide line down the middle. That makes the bar look wider ingame. But if one was to scale the texture, the middle line becomes 4 pixels wide. So then, one can cut it down to 2 pixels wide. That way, it's technically 1 pixel wide because it's scaled back down ingame.

paramat:
Earlier i stated that the xpanes API is probably not heavily used, but thinking on it, it probably is, glass panes and similar nodes are very popular.
I like the look of what you are trying to do, but, anything that breaks texture packs and/or the API will be very disruptive and i think unacceptable. Many users may well prefer the previous appearence too.
The fixed 'ugly squares' are only minor visual glitches, that in a way make sense being on pane junctions.
So i am unfortunately :-1: for the various implementations so far.
> replace this node with a mesh [...] This will of course change the API even more

So is even less acceptable =)
> So to fix it; one can increase the texture size to 32x32

That would break texture packs.

Unless some simple, optional and backwards-compatible implementation is found i feel this project is best be moved to a mod that provides an alternative panes API.

TumeniNodes:
>Unless some simple, optional and backwards-compatible implementation is found i feel this project is best be moved to a mod that provides an alternative panes API.

I fully understand, and went into this realizing though a small fix... it comes with big impacts on current builds, and many users would not be pleased with such a visual change.
But felt it was worth an attempt before just offering it out as a mod.
Thank you to yourself and sfan5 for at least giving it a look

I am closing this, while at the same time opening https://github.com/minetest/minetest_game/pull/2614 to address another issue related to xpanes, I feel should be addressed and could have been done so a while ago.

PR #2611, opened 2020-03-01
Creative: Skip redundant refreshes, fix reset button

SmallJoker:
This PR filters skips redundant calls to `creative.update_creative_inventory` and fixes the 'Reset' button which double-refreshed the filter cache.

For example, searching for any inexistent item and pressing the Reset button resulted in the following formspec:
grafik

### DIFF
--- a/mods/creative/inventory.lua
+++ b/mods/creative/inventory.lua
@@ -100,7 +100,6 @@ function creative.update_creative_inventory(player_name, tab_content)
 	player_inv:set_size("main", #creative_list)
 	player_inv:set_list("main", creative_list)
 	inv.size = #creative_list
-	print("update", inv.size)
 end
 
 -- Create the trash field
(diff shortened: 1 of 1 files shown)


mods/creative/inventory.lua
```diff
@@ -25,7 +25,9 @@ function creative.init_creative_inventory(player)
 	player_inventory[player_name] = {
 		size = 0,
 		filter = "",
-		start_i = 0
+		start_i = 0,
+		_old_filter = nil,
+		_old_content = nil
```
sfan5: I get the idea but why not just `old_filter` and `old_content`

SmallJoker: To denote they're for caching only, and not thought to be changed anywhere else in the code.

sfan5: Sure but I don't see other code using `_` variable names. IMO putting a comment there suffices.

PR #2613, opened 2020-03-05
Dungeon loot: Warn on unknown registered loot items

sfan5:
Now that loot registration was sorted into the mods that provide those items (8d9aa077522135ef31af32579ffab7cf5ce7fd52), there is no reason an unknown item should find its way into the dungeon loot table. So print a warning if that happens.


SmallJoker:
How about either:
- Checking the item in `register()`
- Checking all items in a `register_on_mods_loaded` callback?

sfan5:
I kinda wanted to make this just warning which fits better in the place it is right now.

PR #2626, opened 2020-03-24
Add Rapeseed

### LINKED ISSUE #1679: Add rapeseed to farming
Dyes got too rare with recent changes in MTG. We also have nothing that provides oil.

Oilseed rape/rapeseed fields in full bloom are beautiful in RL. I'm sure a texture can be created that adds a plant that forms such nice fields.
Food mods usually require some oily/fat substance as that's usually part of the human diet in one way or another. Worlds with mobs can get butter from their cows. If there are no cows, there are sometimes crazy receipes like 2 sand -> one butter (probably because the sand has the right color - and there really isn't anything that fits any better).
Oil is also a base for oil colors. Admittedly those are diffrent from our dyes which occour in form of pulver. But then, getting dye as a pulver from a flower will require several steps instead of one crafting in RL as well, so there's no need to stay too realistic. Flower -> dye pulver and dye pulver + oil bottle -> 2 dye pulver would help builders a lot and offer a new way and incentive to do farming.

General idea:
Get the first oilseed seeds from dry shrub (it's there and not used for a plant yet) the same way junglegrass gives cotton seeds and grass gives wheat seeds.
The seeds then grow into oilseed rape with normal farming.
Harvesting fully grown oilseed rape gives more seeds - the seeds are the product. They can be used for planting more rape - or be cooked into oil (VanessaE has a nice texture of an oil bottle).
Dye pulver + oil would yield 2 dye pulver.

Desour:
# Rapeseed

Adds rapeseed, rapeseed oil, a way to make the oil and crafting changes.

Closes #1679.

## Screenshot:
screenshot_20200324_182730
(Plant stages 6 and 7 seem to be swapped in screenshot, oops.)

## How to get oil:
- Get rapeseed from dry shrub.
- Farm it (you'll only get seeds).
- Craft a raw mortar and a raw pestle with clay.
- Bake them in a furnace.
- Put rapeseed (max 3 at once) into a placed mortar with rightclick.
- Pound it with the pestle until you have oil.
- Craft a glass bottle from vessels.
- Leftclick a mortar with oil to get a glass bottle with oil.

## What you can do with the oil:
- ~~Craft oiled sticks for advanced tools (steel and above).~~
- Craft torches with cotton, stick and oil.
- Get more from dye recipes.
- Drink it.
- Use it as fuel.

## State of the PR and Todos:
- Pretty much everything is finished.
Creative is supported, protection is checked and co.
- Textures are improvable. They are also not optimized with optipng or whatever there exists.
- ~~Todo: Add lincenses for new textures.~~
- The farming api lacks documentation, I've added a note about this, todo: remove this note.
- ~~The added `rapeseed_oil.lua` file should be structured better.~~
- Drop chances, times that you have to use the pestle and fuel values should perhaps be adjusted.
- ~~Todo: Add `on_destruct` or similar to mortars to remove them from the pending poundes.~~
- ~~Todo:~~ Drop Rapeseed when mortar with rapeseed is dug or flushed away? Edit: I think, I'll leave this as-is. The seeds from the mortar are flushed away.
- The mortar should maybe be moved to another mod as it might want to be used for other stuff too in the future, like for mixing gunpowder, for producing dye or even as general new crafting type.


npmaier:
What's an impregnated stick?

game_api.txt
```diff
@@ -365,7 +365,9 @@ The farming API allows you to easily register plants and hoes.
 		steps = 8,                             -- How many steps the plant has to grow, until it can be harvested
 		-- ^ Always provide a plant texture for each step, format: modname_plantname_i.png (i = stepnumber)
 		minlight = 13,                         -- Minimum light to grow
-		maxlight = default.LIGHT_MAX           -- Maximum light to grow
+		maxlight = default.LIGHT_MAX,          -- Maximum light to grow
+		only_seed = false,                     -- If true, plant will only drop seeds. plant name is used as name for seed
```
paramat: Long line, 80-90 columns max with tab size 4.

mods/default/craftitems.lua
```diff
@@ -314,6 +314,12 @@ minetest.register_craftitem("default:stick", {
 	groups = {stick = 1, flammable = 2},
 })
 
+minetest.register_craftitem("default:impregnated_stick", {
```
paramat: As this depends on farming mod it seems it should be in that mod?

paramat:
> Get rapeseed from dry shrub.

See issue, please can we have a wild rapeseed plant instead? Although i know you disagree with this. However, i will not disapprove based on this.
> Craft impregnated sticks for advanced tools (steel and above).

I am not sure how this alters tools in the code. Do steel, mese and diamond tools now require an oiled stick?

If so:
:-1: that is disruptive and obviously not acceptable. The oiled stick depends on the farming mod, so half of tools would now depend on using the farming mod, finding dry shrub and and farming rapeseed oil. That challenge is reasonable in isolation or if it had been this way from the beginning, but this is too radical a change to the requirements for crafting the more powerful tools. What about the games that use default but not the farming mod?
I also find the idea of an 'oiled stick' rather silly and random, it seems like trying too hard to find a use for rapeseed oil.
The oiled stick being in default while dependent on the farming mod seems messy.
So unless i misunderstand, please remove the changes to tools and the impregnated stick.

The other features seem ok.

TumeniNodes:
I suggest only using the oil to repair or enhance tools. Meaning that applying the oil to the handles (tool) it only makes it last a certain time longer than normal.

I love the torches crafting change... I would however think adding a way to make `cotton batting` might make sense for torches? (or maybe overkill)
But I am also wondering if rapeseed oil and coal dust could combine to make coal oil for torches? (I get way too detailed)

Desour:
> What's an impregnated stick?

I'm bad at naming things.
It's a new item. Oil is used to strengthen sticks and increase the durability.
Feel free to suggest a better name. I'll probably rename it to "oiled".

---

> wild rapeseed plant

That can be done later if desired.

---

> Do steel, mese and diamond tools now require an oiled stick?

Yes.

I already thought that this feature might be controversial.

The idea behind this is that players can't easily get the best resources just by mining some minutes and feeling to have beaten the game. And bronze should be made useful (for the pre-oil game stage).

Using oil just to repair or to make tools live longer means that you will probably not use oil for steel tools. You can get the mese without having to invest time playing the different parts of the game.

While might think that it's a very disruptive change, players who have already crafted advanced tools are unaffected and those who haven't will be happy that there are changes to the game so that they have more to play with.
Never doing any change will leave the subgame stuck in its early state forever. Games should have some features that make it special. MTG is currently just some boring 08-15 thing. Don't we want to change this.

If someone insists on using the old crafting recipe, they can use a fork of mtg (many server owners probably, hopefully, do that already) or make a mod for it.

>rather silly and random

Splinters can hurt. Tool rods and other stuff out of wood have to be made smooth and using oil can increase the durability. (Oil is hydrophobic and water getting into wood makes it rot.)

>The oiled stick being in default while dependent on the farming mod seems messy.

The mods of a subgame are meant to be used together which is why they can be entangled.
The oiled stick is simply not very "farmy" which is why I placed it into default.

---

>cotton batting

A new craftitem would feel a bit over-kill.
I was not sure what item I should use for the new torch crafting recipe. Maybe string or paper would be better than cotton. I'm open for suggestions.

---

>coal oil

A fast glace into wikipedia told me that coal oil is the name of some processed fossil oil.

game_api.txt
Desour: You're right, the line is very long. So are other lines above.
I'll fix this for my new line. But I hope that there will be a commit that fixes this whole mess before this PR is merged.

Desour:
I've fixed some todos.

TumeniNodes:
cotton is definitely better than string or paper.

Cahrs:
Looks interesting, but you probably should change some names, am i the only one who sees the fishy **rape**seed + stick -> **impregnated**stick? Oiled stick would be much better.

SmallJoker:
`waterproof stick`

Desour:
I've renamed the stick to "oiled" (forgot to update first comment).

>make crafting dyes dependent on the oil

IIRC, I've tried this, but wool depends on dye and farming depends on wool, so dye can't depend on farming.

TumeniNodes:
> I've renamed the stick to "oiled" (forgot to update first comment).

enhanced_stick? (no, wait... that might be worse)
 
> but wool depends on dye and farming depends on wool, so dye can't depend on farming.

... think I'll have me that drink now...

Eventually, down the road, other uses for this oil will come along too. Possibly some healing properties?

TumeniNodes:
>  though as it is im against using it for tool upgrades.

@Cahrs  it should only make them last longer, but not to work faster. i.e. adding oil to a wooden pickaxe handle should make it last longer, but not suddenly break stone as fast as a stone pickaxe.

> Also if it isn't done already, perhaps make crafting dyes dependent on the oil? Nice to see vessels finally getting a use as well.

agreed

Desour:
>healing

You can already drink the oil.

TumeniNodes:
I did miss that, but noticed right after I commented. Apologies
This is really looking good

Cahrs:
Some thoughts when looking at the issue for adding rapeseed again:
-change name to coleseed?
-mix dyes with oil to create different paints (if the circular dependency can be worked around)
-you can dye blocks with group 'canvas' by right-clicking them with paints (using hardware coloring)
-paints come in glass bottles that are returned after the paint is used
-paints can be crafted by combining dye, oil, and a glass bottle (paint bottles that can be used multiple times?)

Oiled sticks still dont seem very useful to me tbh.

Desour:
>change name to coleseed?

I prefer rapeseed. One major reason is that there's no wikipedia article for coleseed, but for rapeseed there is one.

Arbitrary coloring of nodes (apart from wool) is out of scope for this PR. (It is probably also out of scope for MTG.) And imo weirdly colored wood i.e. looks bad, so I don't want to implement such a feature.

>oiled sticks [...] their only use would be to increase tool durability

That's something @TumeniNodes suggested. In the current state of PR you can't increase the durability. Oiled sticks are the only way you can craft advanced/high-level tools.
If you dig with a strong tool, the first thing that will break is not the tool head, but the handle. The chain is only as strong as its weakest link.

rubenwardy:
I feel like this complicates tool crafting to just give oil another use.

I'm not sure how I feel about the dyes issue, but I like that oil can be used as a fuel as well as a dye output increaser

paramat:
:-1: due to the change to tools and the mislocation of the oiled stick registration.
No objection to PR if these are sorted out.

The tools are a fundamental part of MTGame and of default. To suddenly, after 8 years, require dry shrubs (which require deserts), farming mod and farming rapeseed to obtain fundamental tools is crazy.
The tools should never be reliant on a different mod.
Tool crafting should remain simple to be possible in as many different games and game configurations as possible.
Users may be using default without the farming mod. They may not have deserts in their worlds.
We can find enough uses for rapeseed oil to justify it without this.
> Never doing any change will leave the subgame stuck in its early state forever. Games should have some features that make it special. MTG is currently just some boring 08-15 thing. Don't we want to change this.

This is a ridiculous invalid argument that can be used to try to justify anything. If we avoid this change to tools it is not as if MTGame will never change in any way. We can add other features.
I know you are more intelligent than this =)
> If someone insists on using the old crafting recipe, they can use a fork of mtg

This is also a ridiculous argument, the fact that a fork can be made does not justify a bad change.
> The mods of a subgame are meant to be used together which is why they can be entangled.

No. Mods of MTGame are often used independently, often some are left out. They always have to function as if they are independent mods, so need the correct dependencies and need logical structure.
> The oiled stick is simply not very "farmy" which is why I placed it into default.

Ok, but it is dependent on rapeseed so is dependent on farming mod, so should be in the farming mod. The current placement is messy and weird, it is registered in default but the crafting recipe is in farming mod, this is crazy.

Think about it:
Tools require oiled sticks to be crafted. Oiled stick is registered in default. So far so good.
But ...
The oiled stick crafting recipe is in the farming mod. Oiled sticks depend on the farming mod to exist.
But the farming mod is not a dependency of default, it cannot be because the dependency is the other way around.
However, since something as fundamental as tools are now completely dependent on oiled sticks, and therefore on the farming mod, the farming mod absolutely must be a dependency of default, but it cannot be.
This is nuts.

I cannot comprehend how you came to think these 2 aspects of the PR are acceptable, being as talented as you are.
The only solution to this is to remove the changes to tools and move the oiled stick to the farming mod. However if the oiled stick has no use then it should probably be removed from the PR.
Note rubenwardy's comment about tools.

paramat:
> wild rapeseed plant

> That can be done later if desired.

That is not how development works, you know this. If something is necessary it has to be done in the initial PR. PRs are not merged in an undesirable state because 'it can be fixed later'.
However, i do not expect other core devs to request a wild rapeseed plant, and i do not insist on it, so it is probably not being requested for the PR.

Desour:
Well, I already thought that the tool crafting change would encounter opposition. Hence I'll probably remove the oiled stick and the tool crafting changes, after feature freeze at the latest. Anyway, it was worth a try ; ).
Note that this doesn't mean that I agree with all your points, but this doesn't matter for this PR, so I won't go into detail.

Desour:
I've removed the oiled stick and added license information (IANAL and have probably made mistakes there).

paramat:
Keeping open as there seems to be 2 or 3 core devs supporting concept in the issue (although a little unclearly stated).

Desour:
Squashed and rebased.

sirrobzeroone:
Not sure if it helps but I saw some comments about name concern for the mod although correct "rapeseed" if it's more um, friendly downunder we call rapeseed canola - actually the bracketed name in the wiki article.  Also understand that rapeseed and canola in full sense can refer to industrial vs culinary versions...although does not appear to be a super big standard around that and fairly sure downunder at least its all just canola.

Second question with MTG being under a feature freeze will changes like this make there way into the base game? as they were submitted pre-freeze or are these also frozen out and placed on ice? (sorry lots of cold puns there)

sfan5:
You're entirely correct, MTG is frozen for features.

PR #2632, opened 2020-04-04
Fix bed player physics

appgurueu:
Title say it all: Before this PR, beds would always reset player physics to (1, 1, 1) - after this PR, the player physics will be preserved.


NathanSalapat:
This would be really nice when using 3d_armor and armor that overrides the players physics.

sfan5:
Hmm, not sure if this behaviour is really more correct.
In the long term MTG (or the engine?) really do need a solution akin to player_monoids.

mods/beds/functions.lua
```diff
@@ -60,8 +60,10 @@ local function lay_down(player, pos, bed_pos, state, skip)
 
 	-- stand up
 	if state ~= nil and not state then
+		if not beds.player[name] then
+			return
+		end
```
sfan5, line 65: Not sure if this is safe to do, if the overrides are missing it could just fall back to 1,1,1

mods/beds/functions.lua
```diff
@@ -75,15 +77,16 @@ local function lay_down(player, pos, bed_pos, state, skip)
...
 		hud_flags.wielditem = true
 		default.player_set_animation(player, "stand" , 30)
+		beds.player[name] = nil
 
 	-- lay down
 	else
 		beds.pos[name] = pos
 		beds.bed_position[name] = bed_pos
-		beds.player[name] = 1
+		beds.player[name] = player:get_physics_override()
```
sfan5, line 89: Also, is `beds.player` part of the official API? Could anyone be relying on it?

appgurueu: It still has the same falsy/truthy value, so it should be fine.

mods/beds/functions.lua
appgurueu: This is required, because when it's morning, players are kicked twice - the second kick being unneeded.

paramat:
`beds.player[name] = player:get_physics_override()`
Although this works, i think it is a bad implementation to use this for storing physics overrides.
`beds.player[name]` is a simple '1' / 'nil' flag for if a player is in bed, and i think it should remain that way.
Using it for 2 purposes reduces flexibility and may cause problems in future if the 2 purposes conflict and can no longer be combined.
:-1: for this implementation.

appgurueu:
> `beds.player[name]` is a simple '1' / 'nil' flag for if a player is in bed, and i think it should remain that way.

It doesn't matter, as we preserve truthiness. If it was a `true` / `false` flag, I'd get your point, but it pretty much just is an arbitrary truthy value.

Find me a single mod which relies on the value `1` for checking whether a player is in bed (as in, `if beds.player[name] == 1 then ... end` instead of `if beds.player[name] then ... end`), and I will happily lower my code quality for the sake of absolute backwards compatibility.

paramat:
> It doesn't matter, as we preserve truthiness.

That misses my point, i know truthiness is preserved and that is irrelevant to my point.

paramat:
> Find me a single mod which relies on the value 1 for checking whether a player is in bed [...] and I will happily lower my code quality

Simpler, clearer, more readable code is not 'lower code quality'. This is another indication that you have a problem with your approach and attitude to coding, thinking more unnecessary complexity and more 'cleverness' is necessarily better.

PR #2643, opened 2020-04-10
chests: Do not hardcode "default:".

### LINKED ISSUE #1936: Can't register new chest from 3rd party mod

orbea:
Fixes #1936.

This allows mods other than `default` to use `default.chest.register_chest()` to create their own chests.

The problem is that the function hard-codes `default:chest` ~~and `default:chest_locked`~~ leaving other mods unable to create new chests (e.g. `foo:new_chest`) while using the same function as opposed to reinventing the wheel and creating their own.

This changes the `default.chest.register_chest()` to require the mod name in addition to the item name (e.g. `default:chests` instead of `chests`), ~~but if this is not provided it will print a warning message and default to using `default:`.~~

~~This also requires locked chests to provide the `unlocked` definition and unlocked chests to provide the `locked` definition which contains the full name of the corresponding unlocked and locked chests. The `unlocked` definition is not need for unlocked chests and the `locked` definition is not needed for `locked` chests which is determined by if the `protected` definition is set as `true`. By default if these are not provided when needed it will default to `mod:name_locked` for locked chests and `mod:name` for unlocked chests. The caveat being that it depends on the locked chest name being appended with `_locked` when these definitions are missing.~~

Also here is a `testschests` mod that serves as a working test case.

__mod.conf__
```
name = testchests
description = Minetest Game mod: testchests
depends = default
```
__init.lua__
```
default.chest.register_chest("testchests:new_chest", {
	description = "Test Chest",
	tiles = {
		"default_chest_top.png",
		"default_chest_top.png",
		"default_chest_side.png",
		"default_chest_side.png",
		"default_chest_front.png",
		"default_chest_inside.png"
	},
	sound_open = "default_chest_open",
	sound_close = "default_chest_close",
	groups = {choppy = 2, oddly_breakable_by_hand = 2},
})

default.chest.register_chest("testchests:new_chest_protected", {
	description = "Test Locked Chest",
	tiles = {
		"default_chest_top.png",
		"default_chest_top.png",
		"default_chest_side.png",
		"default_chest_side.png",
		"default_chest_lock.png",
		"default_chest_inside.png"
	},
	sound_open = "default_chest_open",
	sound_close = "default_chest_close",
	groups = {choppy = 2, oddly_breakable_by_hand = 2},
	protected = true,
})

minetest.register_craft({
	output = "testchests:new_chest",
	recipe = {
		{"", "group:wood", "group:wood"},
		{"group:wood", "", "group:wood"},
		{"group:wood", "group:wood", "group:wood"},
	}
})

minetest.register_craft({
	output = "testchests:new_chest_protected",
	recipe = {
		{"", "group:wood", "group:wood"},
		{"group:wood", "default:steel_ingot", "group:wood"},
		{"group:wood", "group:wood", "group:wood"},
	}
})

minetest.register_craft( {
	type = "shapeless",
	output = "testchests:new_chest_protected",
	recipe = {"testchests:new_chest", "default:steel_ingot", "default:steel_ingot"},
})
```

### DIFF
--- a/mods/default/chests.lua
+++ b/mods/default/chests.lua
@@ -44,7 +44,7 @@ function default.chest.chest_lid_close(pn)
 	end
 
 	local node = minetest.get_node(pos)
-	minetest.after(0.2, minetest.swap_node, pos, { name = "default:" .. swap,
+	minetest.after(0.2, minetest.swap_node, pos, { name = swap,
 			param2 = node.param2 })
 	minetest.sound_play(sound, {gain = 0.3, pos = pos,
 		max_hear_distance = 10}, true)
@@ -132,7 +132,7 @@ function default.chest.register_chest(name, d)
 					pos = pos, max_hear_distance = 10}, true)
 			if not default.chest.chest_lid_obstructed(pos) then
 				minetest.swap_node(pos,
-						{ name = "default:" .. name .. "_open",
+						{ name = name .. "_open",
 						param2 = node.param2 })
 			end
 			minetest.after(0.2, minetest.show_formspec,
@@ -203,7 +203,7 @@ function default.chest.register_chest(name, d)
 					max_hear_distance = 10}, true)
 			if not default.chest.chest_lid_obstructed(pos) then
 				minetest.swap_node(pos, {
-						name = "default:" .. name .. "_open",
+						name = name .. "_open",
 						param2 = node.param2 })
 			end
 			minetest.after(0.2, minetest.show_formspec,
@@ -215,7 +215,7 @@ function default.chest.register_chest(name, d)
 		def.on_blast = function(pos)
 			local drops = {}
 			default.get_inventory_drops(pos, "main", drops)
-			drops[#drops+1] = "default:" .. name
+			drops[#drops+1] = name
 			minetest.remove_node(pos)
 			return drops
 		end
@@ -248,7 +248,7 @@ function default.chest.register_chest(name, d)
 			def_opened.tiles[i].backface_culling = true
 		end
 	end
-	def_opened.drop = "default:" .. name
+	def_opened.drop = name
 	def_opened.groups.not_in_creative_inventory = 1
 	def_opened.selection_box = {
 		type = "fixed",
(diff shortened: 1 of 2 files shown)


mods/default/chests.lua
```diff
@@ -203,19 +236,19 @@ function default.chest.register_chest(name, d)
...
 			if not default.chest.chest_lid_obstructed(pos) then
 				minetest.swap_node(pos, {
-						name = "default:" .. name .. "_open",
+						name = name .. "_open",
 						param2 = node.param2 })
 			end
 			minetest.after(0.2, minetest.show_formspec,
 					clicker:get_player_name(),
-					"default:chest", default.chest.get_chest_formspec(pos))
+					unlocked, default.chest.get_chest_formspec(pos))
```
sfan5, line 244: I think it's be better to leave the formspec name at `"default:chest"`

mods/default/chests.lua
```diff
@@ -265,29 +298,29 @@ function default.chest.register_chest(name, d)
...
-			local list = inv:get_list("default:chest")
+			local list = inv:get_list(unlocked)
 			if list then
 				inv:set_size("main", 8*4)
 				inv:set_list("main", list)
-				inv:set_list("default:chest", nil)
+				inv:set_list(unlocked, nil)
 			end
 		end
 	})
```
sfan5, line 320: this should be moved out here so it only applies to `default:chest`

mods/default/chests.lua
```diff
@@ -78,13 +75,49 @@ end)
...
 	def.is_ground_content = false
 
+	-- mod name defaults to 'default:'
+	if modname == nil then
+		minetest.log("warning", "default.chests.register_chest() is " ..
+			"missing the mod name, using 'default:'.")
+		modname = "default"
+		itemname = name
+		name = modname .. ":" .. name
+	end
```
sfan5, line 93: This backwards compatibility is not actually needed
For mods _not_ named "default" this function would cause an error anyway. So the only place that could have called this API is the default mod itself.

orbea:
Thanks for the review and suggestions. I kept the default formspec names, made the `register_lbm` call only for `default:` and removed the backwards compatibility as suggested. Everything seems to still work on my end.

mods/default/chests.lua
```diff
@@ -265,29 +266,31 @@ function default.chest.register_chest(name, d)
...
-		action = function(pos, node)
-			local meta = minetest.get_meta(pos)
-			meta:set_string("formspec", nil)
-			local inv = meta:get_inventory()
-			local list = inv:get_list("default:chest")
-			if list then
-				inv:set_size("main", 8*4)
-				inv:set_list("main", list)
-				inv:set_list("default:chest", nil)
+	if modname == "default" then
```
sfan5, line 273: this really just needs to apply to "default:chest" and "default:chest_locked"

mods/default/chests.lua
```diff
@@ -265,29 +266,31 @@ function default.chest.register_chest(name, d)
...
+			nodenames = {name},
+			action = function(pos, node)
+				local meta = minetest.get_meta(pos)
+				meta:set_string("formspec", nil)
+				local inv = meta:get_inventory()
+				local list = inv:get_list(modname .. ":chest")
+				if list then
+					inv:set_size("main", 8*4)
+					inv:set_list("main", list)
+					inv:set_list(modname .. ":chest", nil)
```
sfan5, line 286: this is always "default:chest"

mods/default/chests.lua
orbea: Thanks for making that clear, I changed it accordingly.

mods/default/chests.lua
orbea: I figured that even when its always the same its better to use the variable than a hardcoded value, but its fine either way. I dropped the `modname` variable now.

orbea:
Updated again, thanks for the review!

orbea:
This seems to be a duplicate of this old PR I didn't notice before. https://github.com/minetest/minetest_game/pull/2127

@sfan5 and @SmallJoker Can we get some action on these PRs? I don't care which one is merged, but its silly that a bug fixed in 2018 is still around to be fixed yet again.

sfan5:
Closing in favor of the old PR then.

PR #2644, opened 2020-04-10
furnace: Add default.get_furnace_formspec() to reduce code duplication.

orbea:
This adds `default.get_furnace_formspec()` and deprecates `default.get_furnace_active_formspec()` and `default.get_furnace_inactive_formspec()`.

The latter two functions are nearly duplicate and I think its easier to use and read when there is only one. By default if the first argument, `fuel_percent` is `nil` it will return an inactive furnace formspec and an active furnace formspec otherwise.

The main caveat is that the deprecation warnings are rather loud and spams the console, maybe it will encourage maintainers to update any mods that use this sooner rather than later.

### DIFF
--- a/game_api.txt
+++ b/game_api.txt
@@ -863,15 +863,20 @@ GUI and formspecs
 
  * Entire formspec for the survival inventory
 
+`default.get_furnace_formspec(fuel_percent, item_percent)`
+
+ * Get the furnace formspec using the defined GUI elements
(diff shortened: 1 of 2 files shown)


orbea:
It may also be helpful to expose the image names so that modders can use custom images?

SmallJoker:
I believe pipeworks would benefit quite a lot from this change. Currently it re-implements the entire furnace code to modify the formspec.

PR #2651, opened 2020-04-16
Small sign changes

HybridDog:
I've added a missing article to the sign action message,
removed a redundant nil check (player_name is an empty string for non-player objects),
and removed unused commented code, which may has been used for testing.


orbea:
@HybridDog Not to be too pedantic, but it'd be better if the commits were squashed so you have one small coherent commit instead of 3 less coherent commits.

PR #2694, opened 2020-06-01
Improve some texture

IFRFSX:
There are two considerations:
1. The original texture appearance texture is similar to the the texture in minecraft, which may cause unnecessary trouble.
2. The texture and color of the original are a little monotonous. In order to look good, I have improved the color and appearance.


IFRFSX:
screenshot_20200601_194915
screenshot_20200601_194921
screenshot_20200601_194932
screenshot_20200601_194936

IFRFSX:
screenshot_20200601_201117

I turned it slightly towards orange.

IFRFSX:
screenshot_20200601_201646

i changed it's color to orange.

IFRFSX:
After my inspection, I found some other texture similar to MC shape, so I will replace them as much as possible.

IFRFSX:
2020-06-01 22-04-19 创建的截图

new brick texture by me.

IFRFSX:
screenshot_20200601_234202

Remake two ladder's texture.

IFRFSX:
screenshot_20200601_235302

I remake two ladder's texture.

IFRFSX:
OK.
I've made some unique textures, and if you think it's good, you can merge them right away.
If there is a problem with the texture, please point it out, i will repair it.

SmallJoker:
These textures were changed quite often in the past, hence it'll be difficult to satisfy the contributors and players. You might also be interested in creating a new texture pack.

paramat:
A 'similar design' to MC is not a legal problem, it is only a legal problem if the textures are near-pixel-copies, i doubt our textures are.

The new TNT textures look good in the screenshots.

The new brick texture looks ok but may be controversial as it no longer has equal sized bricks, this will significantly alter the appearance of player's builds.

The steel ladder looks good in the screenshot.

The wooden ladder looks bright orange, it looks like copper, and no longer has any relation to our wood textures :-1: 

Will review more carefully later.

LoneWolfHT:
To add to the above review the bricks (or at least the brick mortar) look a bit pink

paramat:
default_brick128

Ugh, the pink botchy mortar is hideous :-1:

IFRFSX:
screenshot_20200602_113430

I improved 3 Texture, the brick texture has solved the shape problem, I adjusted its brick size to be roughly the same as the original, and the ladder(wood) texture, I have adjusted its color to be close to the original. : )

Andrey2470T:
Why are lines along the uprights of the wooden ladder so abruptly bright?

paramat:
:-1: for the wooden ladder, still very orange, inconsistent with our other wood textures, and has weird bright pixels on the inside of the verticals.

IFRFSX:
default_ladder_wood
screenshot_20200603_132159

I extracted the color of the original [default_wood.png] and used it as the color of this ladder after a little adjustment.

IFRFSX:
screenshot_20200604_212503

Andrey2470T:
Too dark lines on the uprights now. Bad look.

paramat:
compbrick

:-1: this looks worse than the current.

You have changed the locations of the bricks, there is no need to do this.
We recently added a commit that transformed one brick tile so that the tiles match up in a reasonable way at the edges, your changes probably breaks that.
One layer of bricks now no longer has brick centres that line up with the vertical mortar in the other layer of bricks.

The brick colour is a less rich, less textured and less pleasant red. The texture looks flatter.

compladder

:-1: for both, these are far worse than the current ones.

The comparison shows your textures are flat-looking, low contrast and lacking in fine detail.
The pixels on the inner edges of the verticals, in combination with the rungs, make no sense in terms of 3D structure.
The 3D-look shadowing of the current textures is not preserved.

The Steel ladder now has rungs projecting from the edges, looking like the wooden ladder, but this makes no sense for the structure of a steel ladder. The reason the wooden ladder does this is because it has a very simple and old-fashioned structure.

For both bricks and ladders, which you have changed during the PR already, it seems i will dislike anything you do to these, so i am not asking for further changes.
I disapprove of brick and ladder textures being changed in this PR. The current ones are very good and your replacements are of far lower quality.

comptnt

:-1: but no objection if improved.

I like the change to an orange-red, but the shading of the TNT sticks in the TNT block makes no sense because the 'hue' changes from red to orange across a single stick. The material that wraps a TNT stick would be a single hue, it would not have a hue gradient. Only 'value' should change across each stick.
You have done this correctly for the single TNT stick, all pixel colours have (almost) matching hue.
This also means that the single TNT stick is inconsistent with the TNT sticks in the TNT block.

If you colour the sticks in the TNT block using a single hue and with colours matching the TNT stick i will have no objections.
For the TNT stick, one of the orange colours used has a slightly different hue, please make all the hues match.

All your textures have an incorrect RGB 'Colour profile' that displays a warning when i open them, the images should be converted to 'Indexed colour'.

IFRFSX:
@paramat Should I change the texture of the ladder and bricks back to the original version, I only submit the improvement of TNT texture?

IFRFSX:
screenshot_20200605_092605

paramat:
The TNT looks much better in the screenshot. Will check more carefully soon. Thanks.

I am requesting that this PR only changes TNT and TNT stick. Personally i am :-1: for all your brick and ladder textures.
However, this is only my opinion, other core devs may have different opinions.

paramat:
TNT side and stick are now fine.
The TNT top, burning and bottom textures still have many pixels with hue 12 instead of a matching 9, which looks inconsistent.

For all TNT textures, the colour profiles are still incorrect, and the images need to be in 'indexed colour' mode, not RGB.

IFRFSX:
I added some noise to the brick texture to make it look slightly rougher, improved the ladder texture, and then switched to the indexed color mode.

IFRFSX:
screenshot_20200606_122414

IFRFSX:
screenshot_20200606_232901

IFRFSX:
screenshot_20200606_235809

IFRFSX:
screenshot_20200607_005714

An0n3m0us:
I'm sorry; but none of the new textures are to my liking. It's just I so much prefer the default textures how they are already. I suggest you create a texturepack instead.

You seem to have left out-of-place pixels in two places on the ladder:
image

IFRFSX:
> I'm sorry; but none of the new textures are to my liking. It's just I so much prefer the default textures how they are already. I suggest you create a texturepack instead.
> 
> You seem to have left an out-of-place pixel in the top-right part of the ladder texture:
> image
> 
> Also, while showing the new textures, please turn off the shaders; the brick texture looks like it continues at the bottom; probably because you have Parallax Occlusion on.

repaired.

paramat:
I am still :-1: for the ladders. 

TNT colouring is fixed, but the current textures seem slightly better because that white band around the centre is probably what holds the sticks together. Not opposed though.

The bricks do not have the depth and 3D effect of the current texture :-1: 

The colour profiles are still incorrect.

IFRFSX:
@paramat 
So, if you are not satisfied, I will restore the texture of brick and ladder, and only keep the improvement of TNT texture, is it okay?

IFRFSX:
@paramat 
and，the image's profile is Index mode, why incorrect?

paramat:
When i open your textures in GIMP i get this message box:

Screenshot from 2020-06-25 00-50-08

I have never seen that before when working with MT textures.
I do not know why this happens, it might not be a problem but i want to be sure first.

////////////////////////////////

Although i do not disapprove it, i cannot see the point of changing the TNT texture. The colour is now only a very slightly different hue of red, and the white band that held the sticks together has been broken into sections.
Textures should only be changed if there is a good reason to do so, or if the new texture is significantly higher quality (neither is the case here). Otherwise it is only a case of the author's personal taste, in which case they should just use a texture pack.

I will close this for the following reasons: 

* MTGame should only have essential changes made to it, we are already considering putting it into 'maintenance only' mode.
* The changed textures are already good and do not need changing. There is no legal problem concerning MC, and the current textures are not 'monotonous'.
* The new textures are very clearly of much lower quality, with less care and attention to detail, less shadow depth effects, and no improvement when new versions are made.
* The PR is a case of personal texture taste, not a case of textures that need improvement, so you should use a texture pack instead.
* SmallJoker has already hinted at these points in the comment above, which seems to me to be somewhat of a disapproval of this PR https://github.com/minetest/minetest_game/pull/2694#issuecomment-637040337

Sorry, i should have done this much earlier before you put extra effort into this.

PR #2698, opened 2020-06-09
Heavily optimize data usage with player_api

oilboi:
How this works:
1.) A blank control meta string is created for the player.

2.) The animation update loop will check if their input is changed.

3.) If changed it will serialize their control table for the next server step and update their model.


LoneWolfHT:
Intentional close?

sfan5:
Yes, it turned out that player_api was already doing a similar optimization.

PR #2700, opened 2020-06-11
Update template.txt

runsy:
Add Wild Cotton.


sfan5:
The translation templates will be updated in batch before release so this is not necessary.

runsy:
So I close it. :-)

PR #2715, opened 2020-07-12
Move keys to a separate mod. Make keys stackable if they contain the same secret

### LINKED ISSUE #1490: Make keys a separate mod

### LINKED ISSUE #2714: Keys are registered as tools instead of craftitem
Keys are registered using `minetest.register_tool` but in my opinion, they should be registered as `craftitem` instead because:
- keys have no `tool_capabilities`
- keys have no wear, and therefore the repair craft makes no sense (but group `disable_repair=1` is not present in current key definition!)
- this change already has been made for skeleton keys to allow stacking (we could allow stacking of keys sharing the same secret)

louisroyer:
- Keys is a separate module (it depends on default because of the crafting recipe) (Fixes #1490)
- Keys are no longer tools and can be stacked if containing the same secret (Fixes #2714)
- Registered aliases for compatibility


louisroyer:
Rebased because 87f6b941c42d649f09ae9f720645be4fbc39346f modified the code of skeleton key item.

louisroyer:
Rebased because conflict with 9d419bfac61afcf5e4a3ecf336e10fa2d5d3fa1c (edit in `mods/default/locale` files).

louisroyer:
Rebased because github is not able to perform the renaming of textures itself since 6bd059915cb43dbdae6f022aae27062fc52e1219.

louisroyer:
Rebased because new translations were added (b6377a7c0eba0b27e0730d4f25c9633cfd1544e8 and 2c9733ac4593e85cf27a44ad498f52b97499594e).

louisroyer:
Rebased because #2739 added new translation file

sfan5 (approved):
Seems to work but please change the commit messages to be more informative than just "Fix #9999"

louisroyer:
> Seems to work but please change the commit messages to be more informative,
> e.g. "Move keys to separate mod (fixes #9999)"

Ok, thanks for your feedback. I edited the commit messages, sorry for the inconvenience.

mods/keys/craftitems.lua
```diff
@@ -0,0 +1,101 @@
...
+
+		local node_reg = minetest.registered_nodes[node.name]
+		local on_skeleton_key_use = node_reg and node_reg.on_skeleton_key_use
+		if not on_skeleton_key_use then
+			return itemstack
+		end
+
+		-- make a new key secret in case the node callback needs it
+		local random = math.random
+		local newsecret = string.format(
```
appgurueu, line 31: This looks odd to me:

1. Why is this broken into 4 4-digit hex numbers instead of 2 8-digit ones?
2. Why is the one subtracted after the random? This might yield -1 if the random yields 0, which is likely to lead to very unexpected behavior:

```lua
assert(("%04x"):format(-1) == "ffffffffffffffff")
```

If you just want a random 8-byte string as a 16-byte hex string, why not use the following:

```lua
local newsecret = string.format("%08x%08x", random(0xFFFFFFFF), random(0xFFFFFFFF))
```

louisroyer: This has been introduced in e4b1c935120580b013a6f9ba2212a4cb2ef246ac, and I did only move this code to a separate mod. I don't have answers to these 2 questions yet. I will look into this in detail later this week.

louisroyer:
Rebased because 2361987421b9e00ff86fdcccca01538669f11690 changed a translation file.

sfan5 (approved):

appgurueu (approved):
Looks fine. Nitpick: I believe keys will only stack if both secret and description match.

PR #2718, opened 2020-07-13
Add icons for invisible nodes

Wuzzy2:
Problem: Some "special" nodes are invisible but have no icon. As a result, their inventory icon is completely invisible, making debugging really nasty, in case a player got one of these items somehow (e.g. by a bug).

I recommend for all games to make sure all their items (even `/giveme`-only items) have a visible icon. Invisible item icons are very hazardous.

Solution: I added/updated icons for these nodes (all of them are `/giveme` only):

* apple marker
* hidden butterflies/fireflies
* `tnt:boom`
* hidden door segment

This is neither a bug fix nor is it “really” a feature. It's for maintenance to help with debugging in the unlikely event a roge mod gives a player a hidden door segment for some reason.


paramat:
Textures are fine.

About texture naming:

I am thinking `default_invisible_node.png` , or better, `default_invisible_node_overlay.png` is better than `default_hidden_node.png`, as it is more descriptive and more specific.
'Hidden' does not necessarily mean 'invisible'.
I can see that some of the nodes attended to here have 'hidden' in their names/descriptions, which i guess is why you chose that word, but what is actually meant here, and the usage of the texture overlay is, 'the node is invisible'.
'Hidden' only refers to the node and its usage, and might be an unsuitable word for other nodes that need to use this texture overlay.

In `doors_item_hidden.png`, 'item' suggests an item, this is a node. I know that doors are an item when in inventory, but when placed they are certainly nodes.
Maybe 'doors_hidden_segment.png' to match the decription?

Wuzzy2:
Renamed textures.

paramat:
screenshot_20200805_233432

:+1:

paramat:
Bump. Fairly simple PR and a good improvement.

SmallJoker (approved):
LGTM

PR #2725, opened 2020-07-25
screwdriver: Allow protection bypass

tuedel:
Allow players with the `protection_bypass` privilege to rotate protected nodes


tenplus1:
Protector Redo already does this check inside minetest.is_protected and Areas mod defines it's own admin privelage.

SmallJoker:
https://github.com/minetest/minetest/blob/808fa5ecb3ddfd5d993000cc6b4c972257e182db/doc/lua_api.txt#L5554
Added in https://github.com/minetest/minetest/pull/8675

This is not the intended approach. Protection mods must check against this privilege.

PR #2733, opened 2020-08-22
Import translations from minetest-mods/craftguide

p-ouellette:


Wuzzy2:
Note the crafting guide is already translated into German, but this PR adds a German translation.

Wuzzy2:
Now the only new translation in this PR is Italian, the other languages are already present in MTG.

Please remove all languages from this PR except Italian and rename this PR to "Import Italian translation from minetest-mods/craftguide".

Also, since this PR is old, the Italian translation provided here is a little outdated (a few strings are missing), but it's still better than nothing.

sfan5:
Stale, closing.

PR #2735, opened 2020-08-27
Imgbot optimization

Adnan1091:
Reduced texture file sizes losslessly.


paramat:
As only 64 textures are affected, this suggests the PR compresses a small proportion of textures that have not yet been compressed. This is good, as already-compressed textures have not been unnecessarily messed with.

What is the exact configuration of the compression used, what does it do?
The most useful thing to do is remove useless metadata.
The leaves textures must remain as RGBA and not be converted to indexed colour, as the transparent pixels have necessary colour information. So leaves should be tested in all leaves rendering settings.

Adnan1091:
Lossless compression.Not aggresive.So you won't have problems with it.

paramat:
That is very vague. I will not approve a texture compression without details of the compression configuration used.
I know it is lossless, but lossless compression can still break textures.
I looked at the acacia leaves texture in Gimp, it has been converted frm RGBA to indexed colour so has been broken, i assume the other leaves textures have been too. This is why i am being careful with this.
I mentioned the leaves textures above so that you could confirm or check that RGBA textures have not been converted to indexed colour.

paramat:
Really, this task should be done soon before a release, otherwise it just needs doing again, duplicating work.
So thanks for contributing but the PR breaks textures and is done at the wrong time.
Keep an eye on #1831 i will bump that close to release, then you could make a PR if you want.

Adnan1091:
Ok,i'll be waiting.But,this pull request is open to be modified.You could discard leave textures.

paramat:
It is the responsibility of the PR author to make the PR suitable for merging.
You still have not detailed what the compression has exactly done, despite being asked, so i do not trust any of the changes here.
I explained that the leaves must remain RGBA, but you did not check, instead you assured us there would be no problems, but the leaves textures were indeed broken.
I appreciate your intent to contribute, but the approach so far has been sloppy and uncooperative. Currently you are not capable of this task, maybe in future you will be.

Adnan1091:
Screenshot_20200901-103314910
Imgbot's configuration setting is default.I didn't changed anything.I tested this and haven't experienced any issues but benefits.Slightly faster loading time and terrain generation.

PR #2761, opened 2020-10-23
more accurate russian translation.

CyberDonk3y:
Fixed tautology.
As an example, "Деревянные Доски Тропического Дерева" (Wooden Planks of Jungle Wood)
Changed to "Доски из Тропического Дерева" ( Jungle Wood Planks ) etc.

Bush Stem has been translated as "Саженца Яблочного Деревца" ("Apple Tree Sapling")
In this context, tree - "Деревцо" Tender form of word
Roughly speaking, it's also like saying"sweetie tree"
UPD.  *Is Diminutive form, "Деревцо" it something like "treelet"
Changed to "Саженец Куста" ( Bush Sapling )
Also removed the word "Apple" Which should not be there.
and etc.

Ingots of ore , had a name "Брусок Ore" ( Ore Bar )
Changed to Слиток ( Ingot ) - "Слиток" - Minecraft also use this variant.

color "Бурый" ( it's more dark-red than brown, like #320000 HEX ) Changed to "Коричневый" ( Brown )

In general, Minor improvements.


SmallJoker (approved):
> color "Бурый" ( it's more dark-red than brown, like #320000 HEX ) Changed to "Коричневый" ( Brown )

I'm impressed and a bit confused by this way of explaining colors.
The explanations make sense, so this should be good. :+1:

PR #2770, opened 2020-11-01
Fix deprecated get_player_velocity and set_physics_override calls

SmallJoker:
See https://github.com/minetest/minetest/issues/10584, https://github.com/minetest/minetest/pull/10585

PR #2771, opened 2020-11-01
Beds: Fix physical override bug

tenplus1:
This uses the group setting for player override to stop game crashing when using beds.


SmallJoker:
Ninja'd #2770

tenplus1:
lol :P

Issue #2786, opened 2020-11-30
Placing nodes in the air using butterflies


randomMesh:
You can use a butterfly (i guess fireflies too, but i didn't test) to place nodes in the air.
If you destroy the node afterwards, the butterfly is gone.

One should not use the words 'realism' and 'minetest' in the same sentence, but this just feels wrong.

![screenshot_20201130_090930](https://user-images.githubusercontent.com/32773707/100584309-394dc200-32ec-11eb-8a64-07c29a297772.png)

![screenshot_20201130_090941](https://user-images.githubusercontent.com/32773707/100584327-3f43a300-32ec-11eb-9e43-de984e04403e.png)

randomMesh:
Proposed fix (if approved, i'll post a pull request):

```diff
diff --git a/mods/butterflies/init.lua b/mods/butterflies/init.lua
index 49240ee..f2f97c4 100644
--- a/mods/butterflies/init.lua
+++ b/mods/butterflies/init.lua
@@ -31,7 +31,7 @@ for i in ipairs (butter_list) do
                waving = 1,
                paramtype = "light",
                sunlight_propagates = true,
-               buildable_to = true,
+               buildable_to = false,
                walkable = false,
                groups = {catchable = 1},
                selection_box = {
@@ -57,6 +57,9 @@ for i in ipairs (butter_list) do
                                minetest.set_node(pos, {name = "butterflies:hidden_butterfly_"..name})
                        end
                        minetest.get_node_timer(pos):start(30)
+               end,
+               on_rightclick = function(...)
+                       return false
                end
        })
```

NathanSalapat:
I agree that this seems odd, however the problem with this solution is you can't punch the butterflies or fireflies to break them, which means you would have to build a net and catch them if you wanted to build anything in a space where they are.

randomMesh:
> with this solution is you can't punch the butterflies or fireflies to break them

You're right, didn't think of that. I changed the code to take care of this:
```diff
diff --git a/mods/butterflies/init.lua b/mods/butterflies/init.lua
index 49240ee..3b57e1f 100644
--- a/mods/butterflies/init.lua
+++ b/mods/butterflies/init.lua
@@ -31,7 +31,7 @@ for i in ipairs (butter_list) do
                waving = 1,
                paramtype = "light",
                sunlight_propagates = true,
-               buildable_to = true,
+               buildable_to = false,
                walkable = false,
                groups = {catchable = 1},
                selection_box = {
[...]

PR #2787, opened 2020-12-02
Fix: Placing nodes in the air using butterflies (#2786)

randomMesh:
Fix for #2786

### DIFF
--- a/mods/butterflies/init.lua
+++ b/mods/butterflies/init.lua
@@ -59,7 +59,7 @@ for i in ipairs (butter_list) do
 			minetest.get_node_timer(pos):start(30)
 		end,
 		on_rightclick = function(...)
-			return false
+			return nil
 		end,
 		on_punch = function(pos, ...)
 			minetest.remove_node(pos)
--- a/mods/butterflies/init.lua
+++ b/mods/butterflies/init.lua
@@ -31,7 +31,7 @@ for i in ipairs (butter_list) do
 		waving = 1,
 		paramtype = "light",
 		sunlight_propagates = true,
-		buildable_to = true,
+		buildable_to = false,
 		walkable = false,
 		groups = {catchable = 1},
 		selection_box = {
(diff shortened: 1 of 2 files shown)


mods/butterflies/init.lua
```diff
@@ -57,6 +57,12 @@ for i in ipairs (butter_list) do
 				minetest.set_node(pos, {name = "butterflies:hidden_butterfly_"..name})
 			end
 			minetest.get_node_timer(pos):start(30)
+		end,
+		on_rightclick = function(...)
+			return false
```
SmallJoker, line 62: This should either return an itemstack or nil. Why `false`? Is this callback registration necessary at all?

randomMesh: The callback is necessary, otherwise you would still be able to place nodes using the butterfly. Sorry about the return value, i fixed it in another commit.

SmallJoker: Well, that's still possible using shift+place. Also a comment would be more appropriate to avoid this being reverted in a "cleanup PR".

randomMesh: Thanks for the hint. I didn't know about the Shift + RMB thing. I have searched lua_api.txt on how to intercept this combination but i had no luck. How would one make this work? Seems like there is no callback to use.

You're right about the comment, i'll add them once i solved the shift click problem.

SmallJoker: Nodes can be placed as long the butterfly has a selectionbox. I don't think there's any workaround yet, aside converting the butterfly to an object (which is far from ideal).

randomMesh: > I don't think there's any workaround yet

Mmh, too bad. Should i close this pull request or would it make sense to merge it in the current state nevertheless?

randomMesh:
After more than a year, i think no one will merge this.

appgurueu:
Wouldn't `buildable_to = false` suffice?

PR #2788, opened 2020-12-02
Fix boat floating in the air when water being drained away (#2778)

### LINKED ISSUE #2778: Floating carts and boats
If you place a boat, do not move the boat, and the water below it disappears / drains out, the boat keeps floating in mid-air. But the boat will fall as soon as a player starts to move the boat.

If you place a cart on a rail, then remove the block with the rail on it, the cart will float. You can not move it afterwards, unless you somehow find a way to place back the rail.

Version: 5.3.0

randomMesh:
Fixes part of #2778

### DIFF
--- a/mods/boats/init.lua
+++ b/mods/boats/init.lua
@@ -172,7 +172,8 @@ function boat.on_step(self, dtime)
 		end
 	end
 	local velo = self.object:get_velocity()
-	if not self.driver and self.v == 0 and velo.x == 0 and velo.y == 0 and velo.z == 0 then
+	if not self.driver and
+			self.v == 0 and velo.x == 0 and velo.y == 0 and velo.z == 0 then
 		self.object:set_pos(self.object:get_pos())
 		return
 	end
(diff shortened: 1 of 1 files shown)


paramat:
Thanks, will approve soon.

mods/boats/init.lua
```diff
@@ -172,7 +172,7 @@ function boat.on_step(self, dtime)
 		end
 	end
 	local velo = self.object:get_velocity()
-	if self.v == 0 and velo.x == 0 and velo.y == 0 and velo.z == 0 then
+	if not self.driver and self.v == 0 and velo.x == 0 and velo.y == 0 and velo.z == 0 then
```
MoNTE48, line 175:
```
	if not self.driver and
			self.v == 0 and velo.x == 0 and velo.y == 0 and velo.z == 0 then
```

paramat: I agree, as otherwise this goes beyond 90 characters (with tab size 4).

paramat:
Tested and works. Once that line break is done my approval can be assumed.

mods/boats/init.lua
paramat: Done.

paramat:
:+1:

PR #2789, opened 2020-12-06
Beds: Preserve player physics

appgurueu:
Successor of #2632

Preserves bed player physics by storing them as `beds.player[name].player_physics`.

Note:

* Keeps compatibility: Truthyness of `beds.player[name]` is the same. The only cases in which it won't work the same are `type(beds.player[name]) == "number"` or `beds.player[name] == 1`, as well as arithmetic operations (`-beds.player[name] * 1 / 1 + 0 - 0 % 0`). I highly doubt that any of these are out in the wild; if they are, that is on the modders, IMO. They first is a fairly bad check, the second is bad as well but thinkable, and the third might be used for convenience, but most likely isn't, due to the need of some branching or use of logical ops due to the possibility of `nil` anyways.
* Extendable, as it's only a field of `beds.player[name]`. More fields can be added for additional features.


paramat:
As i commented elsewhere:
> beds.player[name] is a simple '1' / 'nil' flag for if a player is in bed, and i think it should remain that way.

There is no harm in having a new table for physics overrides, and it is simpler, clearer and more readable too. There is no reason to avoid doing this.

`beds.player[name] = {physics_override = player:get_physics_override()}`
This is rather unclear and unreadable, i do not even know what value would be given to `beds.player[name]`, i have never seen this type of assignment used before in MTG or mods.

Again i think you are 'being too clever' to the point of forgetting about simplicity, clarity and readability, which are actually more important here and reduce maintainer headache and exhaustion in future.

paramat:
Thanks, this is better now.

SmallJoker (approved):
Works.

I observed in F7 that the player does not lie down. Is that a known problem?

appgurueu:
> I observed in F7 that the player does not lie down. Is that a known problem?

Yes. Probably an engine bug related to `player:set_local_animation(idle, walk, dig, walk_while_dig, frame_speed)`

PR #2790, opened 2020-12-07
Don't replace protected liquids with bones.

auouymous:
Griefers like to repeatedly die in protected water and lava. This destroys lava nodes or causes liquid to flow over the protection.


SmallJoker (approved):
LGTM

PR #2795, opened 2020-12-11
Use Minetest 5.3 'minetest.is_creative_enabled' API

MoNTE48:
Why are mods still dependent on the standard `creative` mod, which also adds inventory that is not used by almost any Minetest server?
The new `minetest.is_creative_enabled` API was added about half a year ago.
It should be used to reduce unnecessary mod dependency.


sfan5 (approved):

SmallJoker (approved):
LGTM

PR #2806, opened 2021-01-14
Do not crash when a skeleton key is used on an unknown node

Montandalar:


sofar:
Sure, yeah

mods/default/craftitems.lua
```diff
@@ -160,8 +160,10 @@ minetest.register_craftitem("default:skeleton_key", {
 			return itemstack
 		end
 
-		local on_skeleton_key_use = minetest.registered_nodes[node.name].on_skeleton_key_use
-		if not on_skeleton_key_use then
+		local on_skeleton_key_use
+		if minetest.registered_nodes[node.name] and minetest.registered_nodes[node.name].on_skeleton_key_use then
```
sfan5, line 164: It would be better to factor out `minetest.registered_nodes[node.name]` to a local variable here

Montandalar: Addressed

Montandalar: woops and once more for good measure (I accidentally used spaces instead of tabs)

SmallJoker:
Shortening it further is possible:

```Lua
		local node_reg = minetest.registered_nodes[node.name]
		local on_skeleton_key_use = node_reg and node_reg.on_skeleton_key_use
		if not on_skeleton_key_use then
			return itemstack
		end
```

Montandalar:
Golfed as per your request

PR #2807, opened 2021-01-22
Compress all .obj files

ExeVirus:
This PR uses lossless "compression" on all the .obj models in minetest_game. This is achieved by using a lua utility I wrote for this exact use case: https://github.com/exevirus/compress-obj

### What "compression" was performed?

It's best to give line-by-line examples, so I'll provide example lines from the current minetest_game master branch models. Specifically, let's use **chest_open.obj** as an example. 

I will show the current line, and the resulting line output, followed by an explanation:
##### Comments
```lua
--Line 1-original:
# Blender v2.78 (sub 0) OBJ File: 'chest-open.blend'

--Line 1-result:


--Note that no output is given for this line, as it is purely a blender default comment, and therefore wasteful
```
##### 'o' Objects
```lua
--Line 3-original:
o Top_Cube.002_None_Top_Cube.002_None_bottom
--Line 3-result:


--Minetest does not use _objects_ for representation of .obj, it uses _groups_. 'o' is therefore removed.
```
##### 'v' Vertexes
```lua
--Line 4-original:
v -0.500000 0.408471 0.720970
--Line 4-result:
v -0.5 0.408471 0.72097
--Note the trailing 0's have been removed, saving 6 characters on this line
```
##### 'vt' Vertex texture coordinates
```lua
--Line 18-original:
vt 0.0000 1.0000
--Line 18-result:
vt 0 1
--Note the trailing 0's have been removed, and 0.0 is just 0. Savings of 10 characters.
```
##### 'vn' Vertex normals
```lua
--Line 52-original:
vn -0.0000 -1.0000 -0.0000
--Line 52-result:
vt 0 -1 0
--Here -0 also becomes 0, saving even more characters: 17
```
##### 'g' Group names
```lua
--Line 61-original:
g Top_Cube.002_None_Top_Cube.002_None_bottom_Top_Cube.002_None_Top_Cube.002_None_bottom_Top
--Line 61-result:
g 1
-- Minetest does not use names, it just counts 'g's, this is far shorter
```
##### 's' Shading smoothness
```lua
--Line 62-original:
s off
--Line 62-result:
s off
-- Minetest does not use it, but if someone imported to blender again, this is helpful for viewing correctly.
```
##### 'f' Faces
```lua
--Line 63-original:
f 6/1/1 5/2/1 2/3/1 1/4/1
--Line 63-result:
f 6/1/1 5/2/1 2/3/1 1/4/1
-- These are always optimized, no work needed.
```
### All else
Deleted/not used, it doesn't matter to Minetest, or to blender. 

## Summary

- This PR removes useless information from .obj files in minetest_game.
- All models have been tested to match original output.
- The result shaves 7.06 KB off of media downloads and minetest distribution.
- Yes this is small, but every KB matters for some users.


MoNTE48:
Converting this model to b3d can give more payoffs as done in my game.
@An0n3m0us

ExeVirus:
@MoNTE48 

Without compression, that is true, but for most cases, this sort of removing needless data results in smaller files than .b3d 

For example, I converted my the original boats_boat.obj to .b3d using blender, and went from 8.89 KB to 11.1KB. 
My compressed version is only 7.66 KB, so .b3d is 44.9% larger than what I have in this PR. 

Here's the results for every file:
| File  | orig obj Size (KB) | b3d size (KB) | compressed obj (KB) |
| --- | --- | --- | --- |
| boats_boat.obj | 8.89 | 11.1 | 7.66 |
| chest_open.obj | 2.24 | 1.41 | 1.08 |
| torch_ceiling.obj | 1.35 | 1.03 | 1.04 |
| torch_floor.obj| 1.32 | 1.03 | 0.82 |
| torch_wall.obj | 1.74 | 1.24 | 1.27 |
| door_a.obj | 0.98 | 0.77 | 0.55 |
| door_a2.obj | 1.14 | 0.77 | 0.71 |
| door_b.obj | 1.14 | 0.77 | 0.69 |
| door_b2.obj | 1.14 | 0.77 | 0.69 |
| doors_fencegate_closed.obj | 2.73 | 2.19 | 1.90 |
| doors_fencegate_open.obj | 2.87 | 2.29 | 2.07 |

And this was done using the most up-to-date 2.8 b3d exporter from GreenXenith.

 In this case, OBJ is still the best option, just with removing useless information.

An0n3m0us:
It is indeed better than the B3D compression, but for animated/multi-texture models, B3D works a charm.

Note: Conflicts with #2804 (because of door model)

ExeVirus:
Thanks for pointing out the conflict @An0n3m0us. 
@paramat Should I just merge those changes from #2804 into this commit?

Helpfully, it will result in even smaller files.

ExeVirus:
Okay, I'll get to this before tomorrow, gotta deal with the conflicts

ExeVirus:
Good to Go, based on latest master (with #2804 changes applied)

An0n3m0us:
Sorry @ExeVirus; seems there was an issue with my PR :/
(#2812)

ExeVirus:
Bummer, alrighty I'll re-revert to the old 1.499 values when I get the chance.

An0n3m0us:
Well, things are a little 'up in the air' at the moment with regards to the door models; I'm waiting for sfan5s final decision on it based on this:
https://github.com/minetest/minetest_game/issues/2812#issuecomment-770785045

ExeVirus:
Okay, reverted #2812 changes, rebased on master, ready to merge.

SmallJoker (commented):
Overall good, but the fence gates now have Z-fighting issues when placed next to an adjacent node with exposed top texture.

appgurueu:
> Overall good, but the fence gates now have Z-fighting issues when placed next to an adjacent node with exposed top texture.

It looks like the lossy "compression" with a precision of only 3 digits was used (`0.720970` becoming `0.721`). This causes the Z-fighting reported by @SmallJoker. To me the compression is only fine if losslessly, which most likely diminishes the gain.

The compression tool should IMO be linked in the Readme for future compression.

I don't like the vertex group optimization. Groups with the same name should get the same index. Perhaps Minetest doesn't care, but Blender will.

With the lossy compression, the tradeoff seems highly questionable so far.

PR #2814, opened 2021-02-01
Stairs: fix rotation on place

alek13:
For example. slabs should placed vertically when places on a "wall".    
    
    use builtin \`rotate_node()\`:
    
    - so, use internal \`core.is_creative_enabled\`
    - fix rotation on stairs & slabs place
    - fix double \"take_item\" in creative mode

### DIFF
--- a/mods/stairs/init.lua
+++ b/mods/stairs/init.lua
@@ -22,32 +22,6 @@ minetest.register_alias("stairs:slab_pinewood", "stairs:slab_pine_wood")
 
 local replace = minetest.settings:get_bool("enable_stairs_replace_abm")
 
-local function rotate_and_place(itemstack, placer, pointed_thing)
-	local p0 = pointed_thing.under
-	local p1 = pointed_thing.above
-	local param2 = 0
-
-	if placer then
-		local placer_pos = placer:get_pos()
-		if placer_pos then
-			param2 = minetest.dir_to_facedir(vector.subtract(p1, placer_pos))
-		end
-
-		local finepos = minetest.pointed_thing_to_face_pos(placer, pointed_thing)
-		local fpos = finepos.y % 1
-
-		if p0.y - 1 == p1.y or (fpos > 0 and fpos < 0.5)
-				or (fpos < -0.5 and fpos > -0.999999999) then
-			param2 = param2 + 20
-			if param2 == 21 then
-				param2 = 23
-			elseif param2 == 23 then
-				param2 = 21
-			end
-		end
-	end
-	return minetest.item_place(itemstack, placer, pointed_thing, param2)
-end
 
 local function warn_if_exists(nodename)
 	if minetest.registered_nodes[nodename] then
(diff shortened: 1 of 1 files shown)


mods/stairs/init.lua
```diff
@@ -194,36 +168,11 @@ function stairs.register_slab(subname, recipeitem, groups, images, description,
...
-					itemstack:take_item()
-				end
-				return itemstack
-			else
-				return rotate_and_place(itemstack, placer, pointed_thing)
+			if pointed_thing.type ~= "node" then
+				return itemstack, nil
 			end
+
+			return minminetest.rotate_node(itemstack, placer, pointed_thing)
```
MoNTE48, line 175: O_o

alek13: oops :)  fixed

sfan5:
PR must merge into `master`, not `stable-5`.

alek13:
@sfan5 
If I understand correctly, `master` branch for now have changes for MT `5.4.0-dev` and possible uses new functionality, which it does not compatible with `5.3` (in general).

So if I want to fix this for `5.3` version, I need to PR into some corresponding branch (usually with same name - `5.3`).

Is there some way to fix this for `5.3` or I misunderstand the workflow ?

sfan5:
Any new changes are put into `master`, then when 5.4.0 is released the current state of `master` is pushed into `stable-5`.
The only exception to this are backport releases, which we don't do currently (there will be no 5.3.1).

You cannot retroactively fix this in 5.3, but what you *can* do is to fix the bug for the next version.

alek13:
@sfan5 
Sadly... :)
Maybe it would be better to create corresponding branches and when it have updates re-release MT with new MTG with patch up (`5.3.1`) ? Sounds crazy? :)
Or maybe some else solution ?

alek13:
@sfan5 @rubenwardy 
Maybe best of all
 - create corresponding branches (`5.1`, `5.2`, `5.3`)
 - release MT without MTG and on first launch ask to download it
 - and than everyone can update MTG if there is updates for theirs version (in corresponding branch)
?

alek13:
rebased on `master`

alek13:
> @sfan5
> Sadly... :)
> Maybe it would be better to create corresponding branches and when it have updates re-release MT with new MTG with patch up (5.3.1) ? Sounds crazy? :)
> Or maybe some else solution ?

> @sfan5 @rubenwardy
> Maybe best of all
> - create corresponding branches (`5.1`, `5.2`, `5.3`)
> - release MT without MTG and on first launch ask to download it
> - and than everyone can update MTG if there is updates for theirs version (in corresponding branch)
> 
> ?

@sfan5 , @rubenwardy 
any thoughts ?
does it make sense to put it in a separate thread (issue) for discussion with others ?

alek13:
@sfan5 can we merge this ?

sfan5:
I haven't got around to reviewing and testing this PR yet, after I (or someone else) has this can be merged.

Regarding review: I'm wondering what happens to the special handling for `stairs:slab_`, does that just disappear?

alek13:
not sure about what "special handling" you are meen, but all functionality works as before

SmallJoker (changes requested):
Side-place on stair

This code is (at least partially) required to allow placing same stairs in a column (place to its side face). This PR breaks a very convenient way of building rooftops because the stairs are now tilted as shown above.

sfan5:
Stale, closing.

PR #2815, opened 2021-02-02
Improvement texture for TNT and Beds

IFRFSX:
The texture of the original bed was a bit monotonous, so I made some adjustments to it.

For TNT texture, in order to make TNT look more beautiful and more distinctive, I have made certain improvements to it to increase the aesthetics（Mainly change red to black）.


IFRFSX:
screenshot_20210203_000141

SmallJoker:
I think those textures would better fit into a texture pack. The current textures are okay as-is.

PR #2816, opened 2021-02-03
Update translation templates

Wuzzy2:
Because new strings were added in `beds`, `fire`, `mtg_craftguide` and `default` (mese post light wood variants).

Templates were automatically updated via a script.


mods/mtg_craftguide/locale/template.txt
```diff
@@ -1,57 +1,53 @@
...
+Recipe @1 of @2=
+Previous recipe=
+Next recipe=
+Recipe is too big to be displayed.=
+Shapeless=
+Cooking time: @1=
+Search=
+Reset=
+Previous page=
 # Label for group ingredients
```
pauloue, line 46: Misplaced comment. Also in the other files

mods/stairs/locale/template.txt
```diff
@@ -1,4 +1,6 @@
 # textdomain: stairs
+Inner =
+Outer =
```
sfan5, line 3: These translations do not exist.

Wuzzy2:
OK, I have manually fixed those 2 template issues. The "Inner " and "Outer " were erronously collected by the string utility because they were in `S()` like so: `S("Inner" .. something_else)`. This confused the script. I have added a workaround for this by using a new function `T` for those which is equal to `S` but it will be ignored by the string util in future. :-)

German translation was removed.

Sorry for squash, I wasn't paying attention. :-(

sfan5:
What matters regarding squashing is that the commit that updates the templates (automated) and the commit that updates the German translation (by you) are separate.
Looking at the changes now you dropped the German translation changes.

Wuzzy2:
I will ASAP post the German strings in separate PR when this PR is done.

mods/mtg_craftguide/locale/template.txt
sfan5: Still here. @Wuzzy2 did you want to remove this?

Wuzzy2:
@sfan5

> Still here. @Wuzzy2 did you want to remove this?

The comment was about commenting the "G" string because its the "label" that marks group ingredients. This comment must be above the "G", but I already did this.

sfan5 (approved):

rubenwardy (approved):

PR #2822, opened 2021-02-06
[NO SQUASH] Maintenance PR

### LINKED ISSUE #2819: Warning spam about use_texture_alpha
The console is completely spammed with warnings like:

```
Texture "doors_door_glass.png" of doors:door_glass_d has transparency, assuming use_texture_alpha = "clip".
```

Can you please fix this? Is there a recommended way to fix this warning? Or would it be better to downgrade the log level of this to `action` or `info`?

Tested in 6a9cbcad827c9e04eee53321844b170dfa63e388, under Minetest 5.4.0-dev.

sfan5:
replaces #2729

fixes #2819

## What was done to the water textures?

Previously the `alpha` value in the nodedef indicated the transparency the water was actually rendered with. Now only the alpha in the texture is respected.
So this PR sets the alpha in there to match the value that was in the nodedef.

## How to test

Join a game and find that:
- there are no deprecation warnings (either `Field "use_texture_alpha": [...]` or `Texture "foo" of bar has transparency [...]`)
- glass stair still work
- glass panes still work
- water & river water have the color they should have


SmallJoker (approved):
Works. 
Checked nodes: mese post, bed, water, lava, glass doors&stairs,

PR #2826, opened 2021-02-08
Fix crash on exit of bed formspec if not sleeping

### LINKED ISSUE #2825: crash when sleep in bed and skip night
minetest version 5.4-dev
multiplayer (only in server)

minetest_game update at commit aa9cc39

In server, when players use skip night button, or leave bed button, server crash

appgurueu:
Closes #2825

### DIFF
--- a/mods/beds/functions.lua
+++ b/mods/beds/functions.lua
@@ -60,6 +60,7 @@ local function lay_down(player, pos, bed_pos, state, skip)
 
 	-- stand up
 	if state ~= nil and not state then
+		assert(beds.player[name], "player " .. name .. " not in bed")
 		local p = beds.pos[name] or nil
 		beds.bed_position[name] = nil
 		-- skip here to prevent sending player specific changes (used for leaving players)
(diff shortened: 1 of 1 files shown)


mods/beds/functions.lua
```diff
@@ -274,7 +275,10 @@ minetest.register_on_player_receive_fields(function(player, formname, fields)
 	local last_player_in_bed = get_player_in_bed_count()
 
 	if fields.quit or fields.leave then
-		lay_down(player, nil, nil, false)
+		if beds.player[player:get_player_name()] then
+			-- make player stand up only if they were in bed
+			lay_down(player, nil, nil, false)
+		end
```
SmallJoker, line 281: I'd prefer to see this check in `lay_down`. That would properly guard from other edge-cases as well. Throwing an error is not necessary the best idea for servers. If there's no need to perform a "stand up", just return silently.

appgurueu: How about keeping this guard, and logging a warning instead of throwing an error?

SmallJoker: I guess that might be okay too

appgurueu: Moved the check to `lay_down`. Haven't added a warning as it's only used internally anyways.

sfan5 (approved):
Looks good.

SmallJoker (approved):

PR #2832, opened 2021-02-17
added Bucket privilege to restrict bucket usage

codeandfix:
Griefers like the bucket, as it is not very safe if used near protected areas.

### DIFF
--- a/mods/bucket/init.lua
+++ b/mods/bucket/init.lua
@@ -4,6 +4,8 @@
 -- Load support for MT game translation.
 local S = minetest.get_translator("bucket")
 
+-- create Bucket privilege to allow Player use the bucket
+minetest.register_privilege("bucket",   {description = "Can use bucket" })
 
 minetest.register_alias("bucket", "bucket:bucket_empty")
 minetest.register_alias("bucket_water", "bucket:bucket_water")
@@ -30,6 +32,13 @@ local function check_protection(pos, name, text)
 		minetest.record_protection_violation(pos, name)
 		return true
 	end
+	
+	-- require Bucket privilege for using a Bucket, as bucket is known to damage protected areas if used by griefers
+	if not minetest.check_player_privs(user:get_player_name(), {bucket=true}) then
+		minetest.chat_send_player(user:get_player_name(), "You dont have the bucket privilege!")
+		return
+	end
+	
 	return false
 end
 


sfan5:
As stated by the pinned issue, Minetest Game is in maintenance mode and not accepting new features. This is unlikely to be accepted.

SmallJoker:
> As stated by the pinned issue, Minetest Game is in maintenance mode and not accepting new features. This is unlikely to be accepted.


As a workaround proposal: expose the function `check_protection` as `bucket.check_protection` so that it can be overwritten/extended in mods. I don't see a need for another privilege just for using buckets.

codeandfix:
smalljoker: you think about overwriting the check protection to check against a new priv bucket.
yes this could be a nice thing. maybe we should have a skeleton protection overwrite mod/api, where it is easier to customize more.
but in this case (bucket/fire) my problem is, that a player can start an ABM which will hurt protected areas.
so my proposal would be: check if an ABM runs from outside a protected area inside an protected area and deny this, if the first action was made by an unpriviliged player OR (if to complicate to check) just disallow any ABM acting from unprotected area into an protected area. otherwise you could ABM-grief (and thats what the bucket priv should prevent - ABM griefing)

SmallJoker:
@codeandfix I totally see why you'd like to have this in MTG, but the current state of the PR is not generic enough. If you would change `check_protection` to a public function, a custom mod of yours could append that  privilege check to this function.

S-S-X:
Add your own server customization mod to keep things clean, here's for example XP is used to check if buckets can be used: https://github.com/pandorabox-io/pandorabox_custom/blob/master/onplace_restriction.lua

codeandfix:
i have to figure out more how overwriting works fine. i fear it wont work in most cases.
i have to overwrite mod functions instead of core functions.
most time i quick hack mods and have problems while upgrading.
but its much easier.

S-S-X:
> i have to figure out more how overwriting works fine. i fear it wont work in most cases.
> i have to overwrite mod functions instead of core functions.

How well overrides work depends on how overrides are implemented. For example that liquid onplace_restriction example does not actually outright override any existing code but wraps it inside custom privilege checks.

Actually rewriting parts of mod code will likely fail some day but wrapping mod code often seems to work forever without issues. That example been unchanged since 2019, some similar things been unchanged without issues way longer.
These kinds of overrides (straight wrapping without changes) also support chaining so multiple mods can override same thing without issues.

Basic function wrapping can often be easily implemented without overriding any of the original code, if it seems you'd need to override actual original code and have no option to wrap it unchanged then it is way more likely to cause issues with updates.

Simply put for usual cases:
* if you're storing and calling original function then you're probably fine.
* if you're not storing and not calling original function then you're probably asking for troubles.

sfan5:
Stale, closing.

PR #2833, opened 2021-02-17
bucket should be loaded after protection mod

codeandfix:
maybe bucket should be loaded after the protection mod e.g. areas is loaded.
otherwise protection may not work correctly?

### DIFF
--- a/mods/bucket/mod.conf
+++ b/mods/bucket/mod.conf
@@ -1,4 +1,4 @@
 name = bucket
 description = Minetest Game mod: bucket
 depends = default
-optional_depends = dungeon_loot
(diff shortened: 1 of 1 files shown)


sfan5:
No? Why would that be the case?

We also cannot just hardcode "areas" in Minetest Game since that's not the only protection mod that exists.

codeandfix:
areas is one example. as its optionally depends, it wouldnt cause trouble.
what happens if bucket is loaded first before the protection mod is loaded?
will there be a short time window for griefers or will it cause problems with the protection?

sfan5:
All mods load completely before the first person can join, there is no need to change anything.

Issue #2836, opened 2021-02-26
Some issues with Minetest 5.4.0


kneekoo:
Some are inherited from 5.3.0 and maybe earlier. Sorry if there are duplicates, and if each item should've been an issue, but assuming owners/moderators can do it in this post, these can be checked if/when they're fixed. Also, I noticed this project is supposed to be in maintenance mode, so maybe one issue is better.

### Recipes
* [ ] The reset button only works after you've searched, not to also clear the initial input/typing
* [ ] book with text = book + book with text (?!)
* [ ] white wool = white dye + white wool (?!)
* [ ] the important recipe for blocks is actually the last; the first shows you can get it from slabs (but you can only make slabs from blocks), then that you can make blocks from stairs, and 3rd is the recipe that's supposed to be the first. Planks and glass are similar.

### Farming
* [ ] Crops [can't grow with torch light](https://i.imgur.com/r3bTHX9.jpg) (underground base - left has holes to the surface, right only has torches)
* [ ] You can place seeds on (growing or fully developed) crops; it won't replace them, but you can see how the seeds get planted and then the crops are almost instantly placed back
* [ ] Water source blocks only hydrate 3 blocks around them - intentional? (Minecraft does 4)
* [ ] Crops only grow if you're in proximity, afk becoming mandatory if you actually need crops

### Water
* [ ] Small water bodies (single blocks and tiny ponds) act like they're part of oceans/seas, looking [really weird for single water sources](https://i.imgur.com/F4bOpZk.jpg) meant to hydrate farming land, and weird enough for artificially made undergorund ponds of 2x2 or even 4x4.
* [ ] Placing solid blocks above water allows you to [see underwater](https://i.imgur.com/620OqBW.jpg)
* [ ] Flowing water has "holes" in [certain instances](https://i.imgur.com/S2Symem.jpg)

### Furnace
* [ ] Smelting/cooking/baking and/or fuel burning time are not based on full ticks: 1 coal burns enough for 13 ingots a few extra ticks - there shouldn't be wasted coal
* [ ] Baking bread is strangely expensive: 1 coal = 2 breads + about 55-60% bread worth of burning, while 2 coal = 5 breads +  about 20-25% of a bread; the least you could do is make sure there's no wasted coal, so either use 1 coal for 1.5 or 2 breads.
* [ ] The item counters need a shadow so you can see it on white background (flour is white, the counter text is white)
* [ ] The furnace sound doesn't stop once the fire stops, but continues for a few seconds

[...]

PR #2837, opened 2021-02-27
mtg_craftguide: Fix incorrect item tooltips

pauloue:
This happened when the same item was visible twice and one of them was a
group item. Fix it by ensuring that buttons that should have different
tooltips don't have the same element name.

Bug reported in #2836.


sfan5:
Is there any possibility that groups or item names contain characters that are not valid inside formspecs?

pauloue:
Item names can only be alphanumeric, underscore, and colon, but apparently group names are not required to follow naming conventions.

sfan5 (approved):
Looks good.

SmallJoker (approved):

PR #2855, opened 2021-03-25
Add sprint

MrDraxs:
i think that the base game could have sprint, its not somenthing to dificult to add in my opinion, and i believe that alot o potential players pull aside the game just because the base game dont have this feature.
in my opinion they should think in somenthing like "ahh this game dont even have sprint why would i play this game(dont even giving chance to the mods and custom games that this game has)".


sfan5:
You accidentally opened a pull request instead of an issue.
Regardless of that Minetest Game is frozen for new features (see pinned issue) and as such sprint will not be added.

PR #2862, opened 2021-04-10
Check for attachments before attaching player to bed

### LINKED ISSUE #2860: Sleeping on boats
You can sleep on boats by rightclicking a bed while on a boat:

Screenshot

When you leave the boat, your physics aren't reset and you can't move.

Possible solution: Check whether the player is attached. Don't allow them to enter a bed, or detach them.

ElCeejo:
Fixes #2860

Simply adds a check for attachments before attaching. That's it. 4 whole lines of code.

### DIFF
--- a/mods/beds/functions.lua
+++ b/mods/beds/functions.lua
@@ -101,6 +101,11 @@ local function lay_down(player, pos, bed_pos, state, skip)
 			minetest.chat_send_player(name, S("You have to stop moving before going to bed!"))
 			return false
 		end
+		
+		-- Check if player is attached to an object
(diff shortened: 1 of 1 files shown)


SmallJoker:
How about an error message?

Desour:
Suggestion for message: "This isn't the time to use that!"

ElCeejo:
> How about an error message?

I thought about it, but I feel like "You can't use beds while attached to an entity" seems out of place, and "You can't use beds while in a boat" is too specific and doesn't account for mods with horses and such.

SmallJoker (approved):

Issue #2864, opened 2021-04-11
A nasty misconception in boat mounting


Hume2:
When a player is attached to an entity and then he enters a boat, the boat tries to unmount the player from the original entity. However, it does in an ugly way. It expects that everything needed to unmount the player is setting the driver property to nil. However, it might not be enough and the original entity has no way to handle this correctly. It is possible to add the on_detach_child callback to the original entity but that is called after the entity is already broken by setting the driver property to nil.

To demonstrate this bug, place a cart next to a boat. Enter cart, then enter boat and then leave the boat. The player's camera is set wrongly, because the player wasn't unmounted from the cart properly.

The easiest way to solve this bug would be forbiding the player to mount the boat when he is attached to anything else. I think that the ability to mount an entity while being attached to another entity doesn't make sense anyway and it just makes things harder.

PR #2865, opened 2021-04-11
Attachments: Proper data cleanup in callbacks

SmallJoker:
Counterpart to https://github.com/minetest/minetest/pull/11181 to fix https://github.com/minetest/minetest_game/issues/2864

This avoids data being overwritten in `on_detach_child` callbacks after calling to `on_attach` from another entity/mod.

## To do

This PR is Ready for Review.

## How to test

1. Place a MTG cart
2. Place a MTG boat
3. Rightclick one and the other, switch between them and observe the eye and attachment positions

### DIFF
--- a/mods/boats/init.lua
+++ b/mods/boats/init.lua
@@ -53,10 +53,9 @@ function boat.on_rightclick(self, clicker)
 	end
 	local name = clicker:get_player_name()
 	if self.driver and name == self.driver then
-		self.driver = nil
-		self.auto = false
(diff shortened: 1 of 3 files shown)


sfan5:
Still buggy:
* Right-click boat
* Right-click  cart
* Right-click  different cart
* Now the following can be observed:
  * You have to right-click twice to detach
  * If you detach by destroying the cart (shift+click) the eye pos is not reset (= ends up too low)

SmallJoker:
Fixed; re-tested.

PR #2871, opened 2021-05-09
Allow to set use_texture_alpha when registering a door

### LINKED ISSUE #2866: Allow to set use_texture_alpha when registering a door
When you register a door, `use_texture_alpha` is always set to `clip`, even if you specify a different value in the door definition. That makes it impossible to register e.g. tinted glass doors.

### Solution
Change [this line](https://github.com/minetest/minetest_game/blob/c6171a8f29a0d6bea2566a74a8bc89121149df93/mods/doors/init.lua#L439) from
```lua
def.use_texture_alpha = "clip"
```
to
```lua
def.use_texture_alpha = def.use_texture_alpha or "clip"
```

Emojigit:
Close #2866 
Co-Authored-By: Gregor Parzefall <gregor.parzefall@posteo.de>

### DIFF
--- a/game_api.txt
+++ b/game_api.txt
@@ -226,7 +226,7 @@ The doors mod allows modders to register custom doors and trapdoors.
 	sound_open = sound play for open door, -- optional
 	sound_close = sound play for close door, -- optional
 	protected = false, -- If true, only placer can open the door (locked for others)
-	on_rightclick = function(pos, node, clicker, itemstack, pointed_thing),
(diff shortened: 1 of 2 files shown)


grorp:
Thank you :)

sfan5:
This needs to be documented in `game_api.txt` too.

Emojigit:
> This needs to be documented in `game_api.txt` too.

@sfan5 Done

PR #2885, opened 2021-06-01
improve zh_CN translation

ferrumcccp:
0. Constitency.
1. Add translations to `beds`.
2. “Emergent” is a forestry term. The last translator didn’t know that.
3. @1 will intersect protection on growth. It’s the tree that grows, not the protection.
4. Cobblestone: Although “圆石” is part of Minecraft’s translation convention(possibly due to early machine translation?), it is not a native expression. Replaced with 鹅卵石.


ferrumcccp:
Sorry, the network here is a bit laggy. If there’s a dup please close it.

sfan5 (approved):

PR #2892, opened 2021-07-30
beds: add option bed_night_skip_above_percent & enable_bed_in_the_daytime

nixnoxus:
adds option `bed_night_skip_above_percent` <0> .. <99> (default: 50 - previously fixed coded value)

With this it is now possible to configure the number of players (in percent) who have to lie in bed at night in order to be able to force skip the night (option `enable_bed_night_skip`)

adds option `enable_bed_in_the_daytime` {true | false} (default: false)

When this option is enabled
- Players can also go to bed in daylight.
- If a player is still in bed between 9:00 and 18:00, the prompt 'Wake Up!' appears. However, this can be ignored;)
- after waking up (in the morning) the player remains lying down and must explicitly 'get out of bed'
This option is useful on survival servers because you can lie down during AFK and possibly not block `enable_bed_night_skip`


SmallJoker:
@nixnoxus Rebase needed.

PR #2894, opened 2021-08-07
Fix the creative trash slot for player 'trash'

### LINKED ISSUE #2777: Creative inventory trash slot is broken for player named "trash"
If you're the player "trash", the trash slot in your Creative inventory is broken. The trash slot will be occupied by an item.

##### How to reproduce
1. Start MTG game server, activate Creative Mode
2. Join as player "trash"
3. Open Creative inventory

##### Analysis
This is becauce the detached creative inventory has the name "creative_[playername]", but the detached trash inventory has the name "creative_trash". So for the player "trash", there is a naming collision.

##### Possible fix
Rename the detached trash inventory "`creative_trash`" to something that can't collide with any player name.

Alternative name suggestions:
* `trash`
* `creative!trash`

Montandalar:
By using the inventory named just 'trash' instead of 'creative_trash' which overlaps when you name your player 'trash'.

Since nothing is meant to live in this slot, it doesn't matter if we rename it. In fact, unified_inventory uses the name trash for its own inventory.

Credit to neoh4x0r for this fix. Fixes #2777 

# How to test

1. Make a player called trash with creative priv.
2. Log in as trash on current master.
3. Notice how the trash slot is occupied by the first thing in the creative picker inventory.
4. Apply this patch.
5. Notice the issue is gone.

### DIFF
--- a/mods/creative/inventory.lua
+++ b/mods/creative/inventory.lua
@@ -125,7 +125,7 @@ function creative.update_creative_inventory(player_name, tab_content)
 end
 
 -- Create the trash field
-local trash = minetest.create_detached_inventory("creative_trash", {
+local trash = minetest.create_detached_inventory("trash", {
 	-- Allow the stack to be placed and remove it in on_put()
 	-- This allows the creative inventory to restore the stack
 	allow_put = function(inv, listname, index, stack, player)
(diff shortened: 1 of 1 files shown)


NathanSalapat:
I believe there was another issue with the player name being trash. which lead to the change to creative_trash

SmallJoker (approved):
LGTM

PR #2912, opened 2021-12-02
[NO SQUASH] Update locale files

Wuzzy2:
This PR updates the locale files of Minetest Game. New strings include "Read" and "Write" in `default`.
As always, this script was used: https://github.com/minetest-tools/update_translations

The 2nd commit updates the German translation.


sfan5 (approved):
It's somewhat bad that the script changes the order of existing translations but I don't really care.

Wuzzy2:
The output order is predictable, tho, and NOT random. It’s always in alphabetical order of Lua files, and within Lua files from first line to last.

Esperanto files were probably generated with a different script (or even manually?!). So later updates should be less "noisy".

PR #2914, opened 2022-01-02
Tim, where are you?

danil275487:
Why did you suddenly dissapear from Discord? What happened to you? Why did you give owner to those fuckers?


rollerozxa:
this is probably the worst way to contact someone...

PR #2915, opened 2022-01-05
Fix torch model rendering in `model[]` formspec element

OgelGames:
Fixes the torch models so they render correctly when used in the `model[]` formspec element. There is no change to rendering of the nodes.

Before:

After:


sfan5:
So what does this actually change about the model? What was the issue?

OgelGames:
The only thing I did to fix them was recalculate the normals, so there must have been something wrong with the normals.

sfan5:
I debugged this a bit and it was the material commands, no need to rewrite the whole file. It's fixed now.

OgelGames:
Ah, that makes more sense, thanks.

PR #2918, opened 2022-01-18
Remove eye offsetting hacks

appgurueu:
The eye height is now managed by player_api and automatically reduced when the sitting or laying animations are set. The resetting of eye height in beds was entirely unneeded and only had potential to break mods. For carts, this fixes the eye height being too low.

### DIFF
--- a/mods/beds/functions.lua
+++ b/mods/beds/functions.lua
@@ -71,7 +71,6 @@ local function lay_down(player, pos, bed_pos, state, skip)
 		end
 		player:set_pos(beds.pos[name])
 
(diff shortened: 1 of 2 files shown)


sfan5:
OK for beds. But for carts it looks like you're flying in 1st person view even though the model is positioned sanely in 3rd person.

appgurueu:
> OK for beds. But for carts it looks like you're flying in 1st person view even though the model is positioned sanely in 3rd person.

Hmm, carts are weird - while the model is standing, the eye offset is supposed to be that of a sitting player? I suspect this is to partially work around #1927.

PR #2921, opened 2022-01-27
Add ukrainian (uk) translation

baytuch:


appgurueu:
Not sure, but shouldn't the ISO language code for this be `ua`?

baytuch:
I think, must be UK

appgurueu:
> I think, must be UK

Yeah you're right.

baytuch:
I added a few more changes in translation to improve it.

sfan5 (approved):
Looks good.

mods/bones/locale/bones.uk.tr
```diff
@@ -0,0 +1,8 @@
+# textdomain: bones
+Bones=Кістки
+@1's old bones=Старі кістки @1
+@1 died at @2.=@1 помер в @2.
+@1 died at @2, and dropped their inventory.=@1 помер в @2 та втратив вміст свого інвентарю.
```
appgurueu, line 5: Google Translate translates this as "... and lost the contents of his inventory ...", which might be misleading (dropping = turned to items, lost = gone forever)?

appgurueu (approved):
Minor "nuance" question, but seems about right.

mods/bones/locale/bones.uk.tr
baytuch: I clarified the translation of this proposal not to create inaccuracies. You are right, gamer can interpret this  ambiguously.

PR #2931, opened 2022-02-13
Refinement translation for UK and RU langs (Cotton)

baytuch:


baytuch:
I am made an addition to translation, because the translation turned out to be incorrect. The fact is that the translation was performed as an indication of the name plant with the title of the final product. I mean that the plant and final products have different names - because it is not true. I corrected it.

appgurueu (approved):
Looks fine according to online translators. Your explanation makes sense.

PR #2936, opened 2022-03-18
Add support for dynamic shadows

x2048:
This PR adds control of shadow intensity and links it to cloud density. It uses `set_lighting` API that is introduced in minetest/minetest#11944.

Behavior is controlled by `weather.shadows_enabled` flag, so that downstream games can disable default behavior if it does not apply.

PR is also compatible with Minetest engine versions without `set_lighting` API.

## Testing

1. Switch to minetest/minetest#11944
2. Enable dynamic shadows in the client
3. Start MTG with any world
4. Observe the shadows
5. Shadow intensity is lowest in biomes with high humidity (and high loud density), e.g. mountains/snow, and shadows have the highest intensity in dry biomes with few clouds, e.g. desert.

### DIFF
--- a/mods/weather/init.lua
+++ b/mods/weather/init.lua
@@ -110,7 +110,7 @@ local function update_clouds()
 				), 2),
 			speed = {x = n_speedx * 4, z = n_speedz * 4},
 		})
-		if weather.shadows_enabled then
+		if weather.shadows_enabled and player.set_lighting ~= nil then
 			player:set_lighting({shadows = { intensity = 0.5 - density / 2.0} })
 		end
 	end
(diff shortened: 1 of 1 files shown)


x2048:
Tagging @sfan5 @appgurueu @lhofhansl 

Here is an example implementation in MTG proposed by @lhofhansl. Might become a good demo of dynamic shadows without having to code your own game or mod.

mods/weather/init.lua
```diff
@@ -6,6 +6,10 @@ if mg_name == "v6" or mg_name == "singlenode" or
 	return
 end
 
+-- Public behavior flags
```
sfan5, line 9: there's an if right above here that disables the mod for v6 or singlenode ... which would result in no shadows
not intended I bet

x2048:
Superseded by #2938.

PR #2937, opened 2022-03-21
default:ladder: Fix placing ladder on each other

bartoszkosiorek:
When the ladder is placed, it is ofter happened unnatural duplicate ladder (see attached screenshot
Wrong-ladder-placement

by enabling `buildable_to` it is no longer possible to break ladder in that way.

### DIFF
--- a/mods/default/nodes.lua
+++ b/mods/default/nodes.lua
@@ -2651,6 +2651,7 @@ minetest.register_node("default:ladder_wood", {
 	walkable = false,
 	climbable = true,
 	is_ground_content = false,
+	buildable_to = true,
 	selection_box = {
 		type = "wallmounted",
 		--wall_top = = <default>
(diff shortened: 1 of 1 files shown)


Desour:
`buildable_to` makes the old node disappear when placed to. This is unacceptable in survival mode.

bartoszkosiorek:
@Desour Do you mean that number of ladders will be decreasing, everytime when the old node will be replaced?

appgurueu:
> @Desour Do you mean that number of ladders will be decreasing, everytime when the old node will be replaced?

Yes. It means that you can simply replace a ladder but placing a node there. This is fine for nodes which are supposed to be "destroyable" that way (liquids for instance), but not for ladders. It also means that you will replace ladders with ladders.

Furthermore, ladder "towers" have been around for a while now.

This is to be implemented either using an engine change, which allows to specify nodes a node can not be mounted to, or an `on_placenode` callback that undoes the ladder placement if the ladder is mounted to another ladder.

pecksin:
This is not a bug. It is possible to place ANY node where you didn't mean to, when placing things fast.

PR #2938, opened 2022-03-26
Adjust shadow intensity based on cloud density.

lhofhansl:
Now that https://github.com/minetest/minetest/pull/11944 is merged, we can use that in minetest_game to set the shadow intensity automatically based on the cloud density.

Shadow intensity becomes as much a game emergence feature as clouds.

### DIFF
--- a/minetest.conf.example
+++ b/minetest.conf.example
@@ -72,6 +72,6 @@ default:torch 99,default:cobble 99
 # Helps rivers create more sound, especially on level sections.
 #river_source_sounds = false
 
-# Enable cloud variation by the 'weather' mod.
+# Enable cloud and shadow intensity variation by the 'weather' mod.
 # Non-functional in V6 or Singlenode mapgens.
 #enable_weather = true
--- a/mods/weather/init.lua
+++ b/mods/weather/init.lua
@@ -1,11 +1,18 @@
 -- Disable by mapgen or setting
 
 local mg_name = minetest.get_mapgen_setting("mg_name")
-if mg_name == "v6" or mg_name == "singlenode" or
-		minetest.settings:get_bool("enable_weather") == false then
+if minetest.settings:get_bool("enable_weather") == false then
 	return
 end
 
+if mg_name == "v6" or mg_name == "singlenode" then
+	-- set a default shadow intensity for mgv6 and singlenode
+	minetest.register_on_joinplayer(function(player)
+		player:set_lighting({ shadows = { intensity = 0.33 } })
+	end)
+
+	return
+end
 
 -- Parameters
 
@@ -96,15 +103,18 @@ local function update_clouds()
 		-- density_max = 0.8 at humid = 50.
 		-- density_max = 1.35 at humid = 100.
 		local density_max = 0.8 + ((humid - 50) / 50) * 0.55
+		local density = rangelim(density_max, 0.2, 1.0) * n_density
 		player:set_clouds({
 			-- Range limit density_max to always have occasional
 			-- small scattered clouds at extreme low humidity.
-			density = rangelim(density_max, 0.2, 1.0) * n_density,
+			density = density,
 			thickness = math.max(math.floor(
 				rangelim(32 * humid / 100, 8, 32) * n_thickness
 				), 2),
 			speed = {x = n_speedx * 4, z = n_speedz * 4},
 		})
+		-- now adjust the shadow intensity
+		player:set_lighting({ shadows = { intensity = 0.7 * (1 - density) } })
 	end
 end
 
--- a/settingtypes.txt
+++ b/settingtypes.txt
@@ -72,6 +72,6 @@ engine_spawn (Use engine spawn search) bool false
 #    Helps rivers create more sound, especially on level sections.
 river_source_sounds (River source node sounds) bool false
 
-#    Enable cloud variation by the 'weather' mod.
+#    Enable cloud and shadow intensity variation by the 'weather' mod.
 #    Non-functional in V6 or Singlenode mapgens.
 enable_weather (Enable weather) bool true


mods/weather/init.lua
```diff
@@ -96,15 +96,20 @@ local function update_clouds()
...
-			density = rangelim(density_max, 0.2, 1.0) * n_density,
+			density = density,
 			thickness = math.max(math.floor(
 				rangelim(32 * humid / 100, 8, 32) * n_thickness
 				), 2),
 			speed = {x = n_speedx * 4, z = n_speedz * 4},
 		})
+		-- now adjust the shadow intensity
+		local light = player:get_lighting()
+		light.shadows.intensity = 0.70 - density * 0.70
```
appgurueu, line 111: Why not make this `0.7 * (1 - density)`?

appgurueu (approved):
Looks good; the luacheck warnings are unrelated.

mods/weather/init.lua
lhofhansl: +1 that's better.

sfan5:
This has the same bug as https://github.com/minetest/minetest_game/pull/2936#discussion_r830377297 and cannot be merged in this state.

appgurueu:
> This has the same bug as [#2936 (comment)](https://github.com/minetest/minetest_game/pull/2936#discussion_r830377297) and cannot be merged in this state.

Why? It's reasonable for shadow intensity to be tied to cloud density (perhaps this should even be implemented at an engine level, but giving modders full control is also fine). I assume said `if` just exists because the assumptions used here to infer a reasonable, biome-based cloud density from heat and humidity just don't hold for v6 so this is better left disabled for v6. Why can't the feature be selectively applied to v6 only, like the rest of the weather mod?

lhofhansl:
@sfan5 Correct, it would not work with MapGen v6, but for other MapGens it does work, and people get shadows when they enabled them, and then they even make sense w.r.t. clouds. I'll take over nothing any day.

Alternatively could put in a default for MGv6 or derive the cloud density in some other way.

sfan5:
It is reasonable to tie cloud density to shadow density, it is *not* reasonable to have shadows just not work on singlenode or v6.
Also I'm not sure if tying this to `enable_weather` is a good idea. In either case you'll have to update [settingtypes.txt](https://github.com/minetest/minetest_game/blob/master/settingtypes.txt#L75) too.

appgurueu:
> It is reasonable to tie cloud density to shadow density, it is _not_ reasonable to have shadows just not work on singlenode or v6.

Why won't shadows work there? They should just have the default shadow intensity as it isn't explicitly turned off.

sfan5:
Also: 

> and then they even make sense w.r.t. clouds.

with the weather mod disabled there'd be a random (let's say 50%), why wouldn't there be shadows logically?

> I'll take over nothing any day.

This is not wrong however we can just do it properly from the start.

sfan5:
> They should just have the default shadow intensity as it isn't explicitly turned off.

In the respective PR I have been informed by @x2048 that the shadow API is opt-in. If it isn't then we have a problem.

lhofhansl:
> In the respective PR I have been informed by @x2048 that the shadow API is opt-in. If it isn't then we have a problem.

If the client has shadows enabled, this will indeed turn shadows on. Setting the intensity **is** the API to opt into shadows from the game side (again only if also enabled on the client). MT Game is a game, and yes it should opt into shadows. So, yes, MT-Game will opt into shadows with this.

I'm happy to make it work with MGv6, if we agree that that's right direction.

(If we're new discussing whether MT-Game can turn on shadows then ... thanks but no thanks, have that discussion without me please. In any case this will go into my ever-growing personal patchset/fork.)

sfan5:
> I'm happy to make it work with MGv6, if we agree that that's right direction.

Sure, here's my proposal:
* Add fallback code that sets shadow intensity to some okay default value (0.5)
* Have this code apply for unsupported mapgens (`v6`, `singlenode`)
* Also have this code apply if the user disables the weather mod via setting

lhofhansl:
How about this? No shadow intensity set when weather is disabled. Defaults to 0.5 for mgv6 and singlenode. (It is always required that the client enabled shadows of course)

x2048:
0.5 looks too dark. The original shadow intensity before the API was added was 0.2, which may feel a bit low-contrast. Something between 0.3-0.35 could be an optimal default value, tunable by gamma on the client side.

lhofhansl:
As you all know I liked it when it was a client-side setting. :)  Now games have to think about how it will look on the client.

0.3 is good. How about I set it to 1/3 (0.33)?

appgurueu:
> As you all know I liked it when it was a client-side setting. :) Now games have to think about how it will look on the client.
> 
> 0.3 is good. How about I set it to 1/3 (0.33)?
> 
> ... done

How about another MTG server-side setting ;)

lhofhansl:
> How about another MTG server-side setting ;)

No more settings if we can at all avoid it. :)
The client still has the shadow gamma to adjust.

lhofhansl:
Ping.

mods/weather/init.lua
```diff
@@ -1,11 +1,20 @@
...
+if minetest.settings:get_bool("enable_weather") == false then
 	return
 end
 
+if mg_name == "v6" or mg_name == "singlenode" then
+	-- set a default shadow intensity for mgv6 and singlenode
+	minetest.register_on_joinplayer(function(player)
+		local light = player:get_lighting()
+		light.shadows.intensity = 0.33
+		player:set_lighting(light)
```
x2048, line 13: it is simpler to just write player:set_lighting({ shadows = { intensity = 0.33 } }). The API call keeps values you don't pass.

lhofhansl: Didn't check that... Yeah, that's much cleaner.

appgurueu (approved):

lhofhansl:
I do not have merge rights on MTG (only on the Minetest core repos it seems), so we need a second approval and then someone to merge.

mods/weather/init.lua
```diff
@@ -1,11 +1,18 @@
 -- Disable by mapgen or setting
 
 local mg_name = minetest.get_mapgen_setting("mg_name")
-if mg_name == "v6" or mg_name == "singlenode" or
-		minetest.settings:get_bool("enable_weather") == false then
+if minetest.settings:get_bool("enable_weather") == false then
```
sfan5, line 4: Needs documentation in `minetest.conf.example` that enable_weather also controls shadows.

minetest.conf.example
```diff
@@ -72,6 +72,6 @@ default:torch 99,default:cobble 99
 # Helps rivers create more sound, especially on level sections.
 #river_source_sounds = false
 
-# Enable cloud variation by the 'weather' mod.
+# Enable cloud and shadow intensity variation by the 'weather' mod.
```
sfan5, line 75: Well technically it doesn't enable shadow *variation* it enables shadows *at all*

sfan5 (approved):
LGTM

minetest.conf.example
lhofhansl: Hmm... Fair. I'm not good at wording this.

How about "Enable cloud variation by the 'weather' mod and enable and adjust shadows accordingly" ?

sfan5: +1 for the latter

lhofhansl:
Wanna amend the last commit with the changed settings text?

appgurueu:
> Wanna amend the last commit with the changed settings text?

Just make a trivial follow-up PR which can be merged instantly. There's no need to rewrite history for this, is there?

PR #2941, opened 2022-03-30
Deduplicate player action logging, silence fake player actions

nixnoxus:
This PR adds the function default.log_action which only logs action of real players.
This is used in on_metadata_inventory_* for 'silent' item transfer between nodes.

Currently there is a log spam on the console for item transfer between nodes like https://github.com/minetest-mods/hopper/ and `default:chest`:

```
2022-03-30 23:07:35: ACTION[Server]: nixnoxus moves default:gravel to chest at (385,22,-119)
2022-03-30 23:07:36: ACTION[Server]: nixnoxus takes default:gravel from chest at (385,22,-119)
2022-03-30 23:07:37: ACTION[Server]: nixnoxus takes default:gravel from chest at (385,22,-119)
2022-03-30 23:07:38: ACTION[Server]: nixnoxus takes default:gravel from chest at (385,22,-119)
2022-03-30 23:07:39: ACTION[Server]: nixnoxus takes default:gravel from chest at (385,22,-119)
2022-03-30 23:07:40: ACTION[Server]: nixnoxus takes default:gravel from chest at (385,22,-119)
2022-03-30 23:07:41: ACTION[Server]: nixnoxus takes default:gravel from chest at (385,22,-119)
2022-03-30 23:07:42: ACTION[Server]: nixnoxus takes default:gravel from chest at (385,22,-119)
2022-03-30 23:07:43: ACTION[Server]: nixnoxus takes default:gravel from chest at (385,22,-119)
2022-03-30 23:07:44: ACTION[Server]: nixnoxus takes default:gravel from chest at (385,22,-119)
2022-03-30 23:07:45: ACTION[Server]: nixnoxus takes default:gravel from chest at (385,22,-119)
2022-03-30 23:07:46: ACTION[Server]: nixnoxus takes default:gravel from chest at (385,22,-119)
2022-03-30 23:07:47: ACTION[Server]: nixnoxus takes default:gravel from chest at (385,22,-119)
2022-03-30 23:07:48: ACTION[Server]: nixnoxus takes default:gravel from chest at (385,22,-119)
```

In the first entry, it was indeed the player. In the entries after that, however, it was the player's hopper.
This is because the hopper acts as a "fake_player".

With this PR, only the first action (that of the real player) would be logged. The item transfer between the nodes is no longer logged (as for example in pipeworks).

I think that would be the desired behavior.

### DIFF
--- a/game_api.txt
+++ b/game_api.txt
@@ -1127,6 +1127,9 @@ setting `log_non_player_actions` is enabled.
 A player is considered non-player if `player:is_player()` returns
 `false` or `player.is_fake_player` is truthy. The use of
 `is_fake_player` is an unofficial standard between mods.
+These non-players are marked by the content of `is_fake_player`
+(if it is a string) or a "*" in brackets after the player name in
+the log.
 
 `default.log_player_action(player, pos, message)`
 
--- a/game_api.txt
+++ b/game_api.txt
@@ -1132,7 +1132,7 @@ A player is considered non-player if `player:is_player()` returns
 
  * `player`  The player who performed the action
  * `pos`     Position of the node at which the action was performed
- * `message` A message describing the action
+ * `message` A message describing the action in 3rd person singular present tense.
 
 `default.set_inventory_action_loggers(def, name)`
 
--- a/game_api.txt
+++ b/game_api.txt
@@ -1123,7 +1123,7 @@ Log API
 Logs action of the player with a node at a certain position.
 By default only actions of real players are logged.
 Actions of non-players (usually machines) are logged only when
-setting `enable_non_player_action_log` is enabled.
+setting `log_non_player_actions` is enabled.
 A player is considered non-player if `player:is_player()` returns
 `false` or `player.is_fake_player` is truthy. The use of
 `is_fake_player` is an unofficial standard between mods.
--- a/game_api.txt
+++ b/game_api.txt
@@ -1142,6 +1142,6 @@ the log.
 
  * sets the callbacks `on_metadata_inventory_move`,
    `on_metadata_inventory_put` and `on_metadata_inventory_take`
-   log corresponding actions
+   that log corresponding actions
  * `def`     See [Node definition]
  * `name`    Description of the node in the log message
--- a/game_api.txt
+++ b/game_api.txt
@@ -1136,7 +1136,7 @@ the log.
  * `player`        The player who performed the action
  * `message_parts` Any mumber of message parts describing the action
                    in 3rd person singular present tense. It can also
-		   contain a `vector` which is logged as " at (X,Y,Z)"
+		   contain a `vector` which is logged as "(X,Y,Z)"
 
 `default.set_inventory_action_loggers(def, name)`
 
--- a/game_api.txt
+++ b/game_api.txt
@@ -1136,7 +1136,7 @@ the log.
  * `player`        The player who performed the action
  * `message_parts` Any mumber of message parts describing the action
                    in 3rd person singular present tense. It can also
-		   contain a `vector` which is logged as "(X,Y,Z)"
+                   contain a `vector` which is logged as "(X,Y,Z)"
 
 `default.set_inventory_action_loggers(def, name)`
 
--- a/game_api.txt
+++ b/game_api.txt
@@ -1127,7 +1127,7 @@ setting `enable_non_player_action_log` is enabled.
 A player is considered non-player if `player:is_player()` returns
 `false` or `player.is_fake_player` is truthy.
 
-`default.log_action(player, pos, message)`
+`default.log_player_action(player, pos, message)`
 
  * `player`  The player who performed the action
  * `pos`     Position of the node at which the action was performed
--- a/game_api.txt
+++ b/game_api.txt
@@ -1131,11 +1131,13 @@ These non-players are marked by the content of `is_fake_player`
 (if it is a string) or a "*" in brackets after the player name in
 the log.
 
-`default.log_player_action(player, message, pos)`
+`default.log_player_action(player, ...)`
 
- * `player`  The player who performed the action
- * `message` A message describing the action in 3rd person singular present tense.
- * `pos`     Position of the node at which the action was performed
+ * `player`        The player who performed the action
+ * `message_parts` Any mumber of message parts describing the action
+                   in 3rd person singular present tense. It can also
+		   contain a `position` / `vector` which is logged as
+		   " at (X,Y,Z)"
 
 `default.set_inventory_action_loggers(def, name)`
 
--- a/game_api.txt
+++ b/game_api.txt
@@ -1124,6 +1124,8 @@ Logs action of the player with a node at a certain position.
 By default only actions of real players are logged.
 Actions of non-players (usually machines) are logged only when
 setting `enable_non_player_action_log` is enabled.
+A player is considered non-player if `player.is_player()` returns
+`false` or `player.is_fake_player` is truthy.
 
 `default.log_action(player, pos, message)`
 
--- a/game_api.txt
+++ b/game_api.txt
@@ -1124,7 +1124,7 @@ Logs action of the player with a node at a certain position.
 By default only actions of real players are logged.
 Actions of non-players (usually machines) are logged only when
 setting `enable_non_player_action_log` is enabled.
-A player is considered non-player if `player.is_player()` returns
+A player is considered non-player if `player:is_player()` returns
 `false` or `player.is_fake_player` is truthy.
 
 `default.log_action(player, pos, message)`
--- a/game_api.txt
+++ b/game_api.txt
@@ -1125,7 +1125,8 @@ By default only actions of real players are logged.
 Actions of non-players (usually machines) are logged only when
 setting `enable_non_player_action_log` is enabled.
 A player is considered non-player if `player:is_player()` returns
-`false` or `player.is_fake_player` is truthy.
+`false` or `player.is_fake_player` is truthy. The use of
+`is_fake_player` is an unofficial standard between mods.
 
 `default.log_player_action(player, pos, message)`
 
--- a/game_api.txt
+++ b/game_api.txt
@@ -1127,7 +1127,7 @@ setting `enable_non_player_action_log` is enabled.
 
 `default.log_action(player, pos, message)`
 
- * `player`  The player who performed the action, may be `nil`
+ * `player`  The player who performed the action
  * `pos`     Position of the node at which the action was performed
  * `message` A message describing the action
 
--- a/game_api.txt
+++ b/game_api.txt
@@ -1136,7 +1136,7 @@ the log.
  * `player`        The player who performed the action
  * `message_parts` Any mumber of message parts describing the action
                    in 3rd person singular present tense. It can also
-                   contain a `vector` which is logged as "(X,Y,Z)"
+                   contain a `pos` which is logged as "(X,Y,Z)"
 
 `default.set_inventory_action_loggers(def, name)`
 
--- a/game_api.txt
+++ b/game_api.txt
@@ -1115,3 +1115,26 @@ This function registers a shapeless recipe that takes `ingredient`
 and `result` as input and outputs `result`.
 
 The metadata of the input `result` is copied to the output `result`.
+
+
+Log API
+-------
+
+Logs action of the player with a node at a certain position.
+By default only actions of real players are logged.
+Actions of non-players (usually machines) are logged only when
+setting `enable_non_player_action_log` is enabled.
+
+`default.log_action(player, pos, message)`
+
+ * `player`  The player who performed the action, may be `nil`
+ * `pos`     Position of the node at which the action was performed
+ * `message` A message describing the action
+
+`default.set_inventory_action_loggers(def, name)`
+
+ * sets the callbacks `on_metadata_inventory_move`,
+   `on_metadata_inventory_put` and `on_metadata_inventory_take`
+   log corresponding actions
+ * `def`     See [Node definition]
+ * `name`    Description of the node in the log message
--- a/game_api.txt
+++ b/game_api.txt
@@ -1131,11 +1131,11 @@ These non-players are marked by the content of `is_fake_player`
 (if it is a string) or a "*" in brackets after the player name in
 the log.
 
-`default.log_player_action(player, pos, message)`
+`default.log_player_action(player, message, pos)`
 
  * `player`  The player who performed the action
- * `pos`     Position of the node at which the action was performed
  * `message` A message describing the action in 3rd person singular present tense.
+ * `pos`     Position of the node at which the action was performed
 
 `default.set_inventory_action_loggers(def, name)`
 
--- a/game_api.txt
+++ b/game_api.txt
@@ -1136,8 +1136,7 @@ the log.
  * `player`        The player who performed the action
  * `message_parts` Any mumber of message parts describing the action
                    in 3rd person singular present tense. It can also
-		   contain a `position` / `vector` which is logged as
-		   " at (X,Y,Z)"
+		   contain a `vector` which is logged as " at (X,Y,Z)"
 
 `default.set_inventory_action_loggers(def, name)`
 
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -758,11 +758,21 @@ function default.can_interact_with_node(player, pos)
 	return false
 end
 
+local non_player_action_log = minetest.settings:get_bool("enable_non_player_action_log") or false
+
 function default.log_action(player, pos, message)
-	-- only log actions of real players
+	local who = player and player:get_player_name() or "(something)"
 	if player and not player.is_fake_player and player:is_player() then
-		minetest.log("action", player:get_player_name() ..
-			" " .. message .. " at " .. minetest.pos_to_string(pos))
+		-- log action of real player
+		minetest.log("action",  who .. " " .. message .. " at " .. minetest.pos_to_string(pos))
+	elseif non_player_action_log ~= false then
+		-- log action of non real player
+		if player and player.is_fake_player and type(player.is_fake_player) == "string" then
+			who = who .. "(" .. player.is_fake_player .. ")"
+		else
+			who = who .. "(*)"
+		end
+		minetest.log("action",  who .. " " .. message .. " at " .. minetest.pos_to_string(pos))
 	end
 end
 
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -719,14 +719,14 @@ end
 -- Log API / helpers
 --
 
-local non_player_action_log = minetest.settings:get_bool("enable_non_player_action_log") ~= false
+local log_non_player_actions = minetest.settings:get_bool("log_non_player_actions") ~= false
 
 function default.log_player_action(player, pos, message)
 	local who = player:get_player_name()
 	if not player.is_fake_player and player:is_player() then
 		-- log action of real player
 		minetest.log("action",  who .. " " .. message .. " at " .. minetest.pos_to_string(pos))
-	elseif non_player_action_log then
+	elseif log_non_player_actions then
 		-- log action of non real player
 		who = who .. "(" .. (type(player.is_fake_player) == "string"
 			and player.is_fake_player or "*").. ")"
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -719,7 +719,7 @@ end
 -- Log API / helpers
 --
 
-local log_non_player_actions = minetest.settings:get_bool("log_non_player_actions") ~= false
+local log_non_player_actions = minetest.settings:get_bool("log_non_player_actions") or false
 
 function default.log_player_action(player, message, pos)
 	local who = player:get_player_name()
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -737,7 +737,10 @@ function default.log_player_action(player, ...)
 			and player.is_fake_player or "*").. ")"
 	end
 	for _, v in ipairs({...}) do
-		msg = msg .. " " .. (is_vector(v) and minetest.pos_to_string(v) or v)
+		-- translate vectors
+		local part = is_vector(v) and minetest.pos_to_string(v) or v
+		-- no leading spaces before punctuation marks
+		msg = msg .. (string.match(part, "^[;,.]") and "" or " ") .. part
 	end
 	minetest.log("action",  msg)
 end
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -721,16 +721,27 @@ end
 
 local log_non_player_actions = minetest.settings:get_bool("log_non_player_actions", false)
 
-function default.log_player_action(player, message, pos)
-	local who = player:get_player_name()
+-- `vector.check` is available since minetest >= 5.5.0
+local is_vector = vector.check or function(v)
+	return v and type(v) == "table" and
+		v.x or type(v.x) == "number" and
+		v.y or type(v.y) == "number" and
+		v.z or type(v.z) == "number"
+end
+
+function default.log_player_action(player, ...)
+	local msg = player:get_player_name()
 	if player.is_fake_player or not player:is_player() then
 		if not log_non_player_actions then
 			return
 		end
-		who = who .. "(" .. (type(player.is_fake_player) == "string"
+		msg = msg .. "(" .. (type(player.is_fake_player) == "string"
 			and player.is_fake_player or "*").. ")"
 	end
-	minetest.log("action",  who .. " " .. message .. " at " .. minetest.pos_to_string(pos))
+	for _, v in ipairs({...}) do
+		msg = msg .. " " .. (is_vector(v) and ("at " .. minetest.pos_to_string(v)) or v)
+	end
+	minetest.log("action",  msg)
 end
 
 function default.set_inventory_action_loggers(def, name)
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -717,8 +717,8 @@ end
 
 
 --
--- NOTICE: This method is not an official part of the API yet.
--- This method may change in future.
+-- NOTICE: This methods are not an official part of the API yet.
+-- This methods may change in future.
 --
 
 function default.can_interact_with_node(player, pos)
@@ -756,4 +756,12 @@ function default.can_interact_with_node(player, pos)
 	end
 
 	return false
+end
+
+function default.log_action(player, pos, message)
+	-- only log actions of real players
+	if player and not player.is_fake_player and player:is_player() then
+		minetest.log("action", player:get_player_name() ..
+			" " .. message .. " at " .. minetest.pos_to_string(pos))
+	end
 end
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -723,10 +723,8 @@ local log_non_player_actions = minetest.settings:get_bool("log_non_player_action
 
 -- `vector.check` is available since minetest >= 5.5.0
 local is_vector = vector.check or function(v)
-	return v and type(v) == "table" and
-		v.x and type(v.x) == "number" and
-		v.y and type(v.y) == "number" and
-		v.z and type(v.z) == "number"
+	return type(v) == "table" and
+		type(v.x) == "number" and type(v.y) == "number" and type(v.z) == "number"
 end
 
 function default.log_player_action(player, ...)
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -719,7 +719,7 @@ end
 -- Log API / helpers
 --
 
-local log_non_player_actions = minetest.settings:get_bool("log_non_player_actions") or false
+local log_non_player_actions = minetest.settings:get_bool("log_non_player_actions", false)
 
 function default.log_player_action(player, message, pos)
 	local who = player:get_player_name()
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -748,7 +748,7 @@ function default.set_inventory_action_loggers(def, name)
 end
 
 --
--- NOTICE: This method are not an official part of the API yet.
+-- NOTICE: This method is not an official part of the API yet.
 -- This method may change in future.
 --
 
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -737,7 +737,7 @@ function default.log_player_action(player, ...)
 			and player.is_fake_player or "*").. ")"
 	end
 	for _, v in ipairs({...}) do
-		msg = msg .. " " .. (is_vector(v) and ("at " .. minetest.pos_to_string(v)) or v)
+		msg = msg .. " " .. (is_vector(v) and minetest.pos_to_string(v) or v)
 	end
 	minetest.log("action",  msg)
 end
@@ -745,13 +745,13 @@ end
 function default.set_inventory_action_loggers(def, name)
 	def.on_metadata_inventory_move = function(pos, from_list, from_index,
 			to_list, to_index, count, player)
-		default.log_player_action(player, "moves stuff in", name, pos)
+		default.log_player_action(player, "moves stuff in", name, "at", pos)
 	end
 	def.on_metadata_inventory_put = function(pos, listname, index, stack, player)
-		default.log_player_action(player, "moves", stack:get_name(), "to", name, pos)
+		default.log_player_action(player, "moves", stack:get_name(), "to", name, "at", pos)
 	end
 	def.on_metadata_inventory_take = function(pos, listname, index, stack, player)
-		default.log_player_action(player, "takes", stack:get_name(), "from", name, pos)
+		default.log_player_action(player, "takes", stack:get_name(), "from", name, "at", pos)
 	end
 end
 
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -748,8 +748,8 @@ function default.set_inventory_action_loggers(def, name)
 end
 
 --
--- NOTICE: This methods are not an official part of the API yet.
--- This methods may change in future.
+-- NOTICE: This method are not an official part of the API yet.
+-- This method may change in future.
 --
 
 function default.can_interact_with_node(player, pos)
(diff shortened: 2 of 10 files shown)


mods/default/functions.lua
```diff
@@ -717,8 +717,8 @@ end
 
 
 --
--- NOTICE: This method is not an official part of the API yet.
--- This method may change in future.
+-- NOTICE: This methods are not an official part of the API yet.
```
appgurueu, line 720: These*

appgurueu (approved):
looks mostly fine to me, but I feel this could be deduplicated way better by having a function that registers all the three `on_*` callbacks for a node

mods/default/functions.lua
nixnoxus: I think that log_actions() is not an official API function either. On the other hand, i would like to see that. :smiley:

nixnoxus: aha, sorry. I fixed the typo

nixnoxus:
> looks mostly fine to me, but I feel this could be deduplicated way better by having a function that registers all the three `on_*` callbacks for a node

I wanted it simple and flexible while keeping the original texts.

Also parameters of `on_metadata_inventory_move` are different to `on_metadata_inventory_put` and `on_metadata_inventory_take`.

So I have not found a more generic way

mods/default/functions.lua
```diff
@@ -757,3 +757,11 @@ function default.can_interact_with_node(player, pos)
 
 	return false
 end
+
+function default.log_action(player, pos, message)
+	-- only log actions of real players
+	if player and not player.is_fake_player and player:is_player() then
```
appgurueu, line 763: After some more consideration, I'm not quite happy with this (particularly if it's not documented). It relies on the `player.is_fake_player` field which is neither a MTE nor a MTG API and rather a convention a couple mods have stuck to. I wouldn't want MTG to add compatibility for random mods this way; rather, those mods should be adding compat for MTG, or where inconvenient, a decent MTG API should be created; designing the API to comply with mods seems odd to me though.

appgurueu (changes requested):
I'm starting to question the whether this can generally be considered the "desired" behavior.

Why would mods be calling these callbacks with a fake player object if not to trigger the callbacks (including logging)? What about griefing that happens "through" these mods?

As for the deduplication I had in mind: This set of callbacks is common across all 3 inventory nodes (bookshelves, chests and vessel shelves). I'd like it if there was a function `default.set_inventory_action_loggers(nodedef, name)` that set the three callbacks, perhaps even a more general form which would just use the nodenames (although this would be a somewhat "breaking change" as it changes log messages people might have come to rely on).

appgurueu:
> on the one hand the content of the message is wrong, because it is the hopper of the player and not the player himself. on the other hand this can lead to a lot of messages on the server.

That's not a Minetest Game issue but rather a mod issue.

> Mods like hoppers are machines that work independently (without any action from the player). These work timer or abm driven. That is why there is no player object. The callbacks allow_metadata_inventory_* and on_metadata_inventory_* are called if necessary to check protections by the node on the one hand and to inform the node about a inventory change on the other hand. This serves e.g. to activate the furnace when coal is put into the empty 'fuel' inventory slot.

Protection could be checked using `minetest.is_protected` (which isn't quite the same as ownership / inventory action permissions in all cases, I'll admit). It's reasonable that notifying the node about the change would require these callbacks. For other callbacks, MT just allows omitting the player; I suppose that wouldn't be useful here as it's still required for the permission checks.

> I think griefing mods would not call these callbacks at all.

It's not about griefing mods, but rather about griefers (or other malicious players) being able to evade logging through these machines while still doing the same "actions".

> Well, if it's okay to change the messages, then that makes sense. +1

It doesn't necessarily require changing the messages.

mods/default/functions.lua
nixnoxus: without `player.is_fake_player` we can not see if it is a real player. 
hopper and pipeworks mod set, for example, `is_player = function() return true end`:
- https://github.com/minetest-mods/hopper/blob/master/utility.lua#L61
- https://github.com/mt-mods/pipeworks/blob/master/common.lua#L183

This PR would then no longer make sense.

nixnoxus:
> > on the one hand the content of the message is wrong, because it is the hopper of the player and not the player himself. on the other hand this can lead to a lot of messages on the server.
> 
> That's not a Minetest Game issue but rather a mod issue.
> 
my intention is to find and establish a unified way to improve the interaction of the different mods.

minetest_game seems to me to be a good starting point for this, since many orient themselves on this and also start with code snippets from here.

> > Mods like hoppers are machines that work independently (without any action from the player). These work timer or abm driven. That is why there is no player object. The callbacks allow_metadata_inventory_* and on_metadata_inventory_* are called if necessary to check protections by the node on the one hand and to inform the node about a inventory change on the other hand. This serves e.g. to activate the furnace when coal is put into the empty 'fuel' inventory slot.
> 
> Protection could be checked using `minetest.is_protected` (which isn't quite the same as ownership / inventory action permissions in all cases, I'll admit). It's reasonable that notifying the node about the change would require these callbacks. For other callbacks, MT just allows omitting the player; I suppose that wouldn't be useful here as it's still required for the permission checks.
> 
I have not found an alternative callback for the 'fuel' problem. Even setting the node timer i think is not a good idea. This may work for the `default:furnace`, but for mods with more complex nodes this can lead to problems (i.e. `pipeworks:autocrafter`).

> > I think griefing mods would not call these callbacks at all.
> 
> It's not about griefing mods, but rather about griefers (or other malicious players) being able to evade logging through these machines while still doing the same "actions".
> 
I add a setting `enable_non_player_action_log`
If  enabled, the (non-)player name will be followed by "(*)" or the string of is_fake_player if applicable.

see https://github.com/minetest/minetest_game/pull/2941/commits/e5a13430485d67c58d6845c0b0f54ae72c2fb225

> > Well, if it's okay to change the messages, then that makes sense. +1
> 
> It doesn't necessarily require changing the messages.
- "move" vs. "put"
- "staff" vs.  stack:get_name()
see https://github.com/minetest/minetest_game/pull/2941/commits/8b8c693edd245182aa61062839288fa414c02806

appgurueu:
> my intention is to find and establish a unified way to improve the interaction of the different mods.

Then at the very least you ought to document this in `game_api.txt`.

mods/default/functions.lua
```diff
@@ -757,3 +757,34 @@ function default.can_interact_with_node(player, pos)
 
 	return false
 end
+
+local non_player_action_log = minetest.settings:get_bool("enable_non_player_action_log") or false
+
+function default.log_action(player, pos, message)
+	local who = player and player:get_player_name() or "(something)"
```
appgurueu, line 764: Please remove the `or "(something)"` as it wasn't there before either; a valid `player` object must have a name (or else it *should* throw the "attempt to concat a nil val" err). I'm also not sure whether `nil` players should be allowed here.

mods/default/functions.lua
appgurueu: the `if player` and `player:is_player()` still appear redundant to me: obviously `player:is_player()` is supposed to be a tautology - this callback does, as it's currently documented, not accept `nil` or invalid players and it's not on MTG to change that

mods/default/functions.lua
```diff
@@ -757,3 +757,34 @@ function default.can_interact_with_node(player, pos)
...
+	if player and not player.is_fake_player and player:is_player() then
+		-- log action of real player
+		minetest.log("action",  who .. " " .. message .. " at " .. minetest.pos_to_string(pos))
+	elseif non_player_action_log ~= false then
+		-- log action of non real player
+		if player and player.is_fake_player and type(player.is_fake_player) == "string" then
```
appgurueu, line 770: again `if player`

mods/default/functions.lua
```diff
@@ -757,3 +757,34 @@ function default.can_interact_with_node(player, pos)
...
+			who = who .. "(" .. player.is_fake_player .. ")"
+		else
+			who = who .. "(*)"
+		end
```
appgurueu, line 774: could shorten this to `who = who .. "(" .. (type(player.is_fake_player) == "string" and player.is_fake_player or "*") .. ")"

mods/default/functions.lua
```diff
@@ -757,3 +757,34 @@ function default.can_interact_with_node(player, pos)
 
 	return false
 end
+
+local non_player_action_log = minetest.settings:get_bool("enable_non_player_action_log") or false
```
appgurueu, line 761: You're ensuring that it must be a boolean here.

mods/default/functions.lua
```diff
@@ -757,3 +757,34 @@ function default.can_interact_with_node(player, pos)
...
+
+function default.log_action(player, pos, message)
+	local who = player and player:get_player_name() or "(something)"
+	if player and not player.is_fake_player and player:is_player() then
+		-- log action of real player
+		minetest.log("action",  who .. " " .. message .. " at " .. minetest.pos_to_string(pos))
+	elseif non_player_action_log ~= false then
```
appgurueu, line 768: Why this convoluted check when it must be a boolean - `true` or `false` and can't be `nil` anyways? Simply `elseif non_player_action_log` should suffice.

mods/default/functions.lua
nixnoxus: fixed in https://github.com/minetest/minetest_game/pull/2941/commits/eb162e6c2951c9e76f38b8157b40c3665fb5334a

nixnoxus:
thanks for the feedback :smile:

All fixed in https://github.com/minetest/minetest_game/pull/2941/commits/eb162e6c2951c9e76f38b8157b40c3665fb5334a

game_api.txt
```diff
@@ -1115,3 +1115,26 @@ This function registers a shapeless recipe that takes `ingredient`
 and `result` as input and outputs `result`.
 
 The metadata of the input `result` is copied to the output `result`.
+
+
+Log API
+-------
+
+Logs action of the player with a node at a certain position.
+By default only actions of real players are logged.
```
appgurueu, line 1124: Please document that players are considered "fake" only if `player.is_fake_player` is truthy.

game_api.txt
```diff
@@ -1124,6 +1124,8 @@ Logs action of the player with a node at a certain position.
 By default only actions of real players are logged.
 Actions of non-players (usually machines) are logged only when
 setting `enable_non_player_action_log` is enabled.
+A player is considered non-player if `player.is_player()` returns
```
appgurueu, line 1127: `player:is_player()`, the dot would trigger an error

mods/default/functions.lua
```diff
@@ -715,10 +715,41 @@ function default.register_craft_metadata_copy(ingredient, result)
...
+	end
+	def.on_metadata_inventory_take = function(pos, listname, index, stack, player)
+		default.log_action(player, pos, "takes " .. stack:get_name() .. " from " .. name)
+	end
+end
 
 --
--- NOTICE: These methods are not an official part of the API yet.
--- These methods may change in future.
+-- NOTICE: This methods are not an official part of the API yet.
```
appgurueu, line 751: This somehow went back to "this methods"?

nixnoxus:
I looked at the other `minetest.log("action", ...` messages where it is about player actions.

For consistency it would make sense to use the `default.log_player_action` function here as well.

However, there are some variations in the format such as:

in ./mods/bones/init.lua:

- player_name .. " dies at " .. pos_string .. ". No bones placed"
- player_name .. " dies at " .. pos_string .. ". Bones placed"
- player_name .. " dies at " .. pos_string .. ". Inventory dropped"

in ./mods/give_initial_stuff/init.lua:

- "Giving initial stuff to player " .. player:get_player_name()

in ./mods/creative/inventory.lua:

- player_name .. " takes " .. stack:get_name().. " from creative inventory"

Would it be okay to change the other messages from player actions to `default.log_player_action` as well and change the messages just mentioned as follows?

- player_name .. " " .. "died and left no bones"     .. " at " .. pos
- player_name .. " " .. "died and left bones"        .. " at " .. pos
- player_name .. " " .. "died and dropped inventory" .. " at " .. pos
- player_name .. " " .. "received initial stuff"     .. " at " .. pos
- player_name .. " " .. "takes " .. stack:get_name() .. " from creative inventory" .. " at " .. pos

game_api.txt
nixnoxus: done

game_api.txt
nixnoxus: thx, fixed

mods/default/functions.lua
nixnoxus: Back to the singular (and fixed). I moved my code over it because it is now documented in game_api.txt.

appgurueu:
> Would it be okay to change the other messages from player actions to `default.log_player_action` as well and change the messages just mentioned as follows?

I'm not sure about changing log messages because it makes grepping harder while the logs contain messages in both formats. Then again "human-readable" logs aren't exactly computer-friendly.

Two of the messages stick out to me though:

> player_name .. " " .. "received initial stuff" .. " at " .. pos
> player_name .. " " .. "takes " .. stack:get_name() .. " from creative inventory" .. " at " .. pos

Both messages don't need the pos. It might even be misleading as for the other logs messages, the pos is a node pos, whereas it has to be a player pos for these ones. I suggest just allowing the omission of pos.

nixnoxus:
Alternatively, the log function could be rewritten to `default.log_player_action(player, message, pos, message_addl)`, where the parameters `pos` and `message_add` would be optional.

Then most messages could use this function without modification.

Messages that do not fit into this scheme will then remain untouched.

Would this be a good solution?

game_api.txt
```diff
@@ -1115,3 +1115,28 @@ This function registers a shapeless recipe that takes `ingredient`
...
 The metadata of the input `result` is copied to the output `result`.
+
+
+Log API
+-------
+
+Logs action of the player with a node at a certain position.
+By default only actions of real players are logged.
+Actions of non-players (usually machines) are logged only when
+setting `enable_non_player_action_log` is enabled.
```
sfan5, line 1126: `log_non_player_actions` is a better name for this IMO.

game_api.txt
```diff
@@ -1115,3 +1115,28 @@ This function registers a shapeless recipe that takes `ingredient`
...
+A player is considered non-player if `player:is_player()` returns
+`false` or `player.is_fake_player` is truthy.
```
sfan5, line 1128:
```suggestion
`false` or `player.is_fake_player` is truthy. The use of
`is_fake_player` is an unofficial standard between mods.
```

mods/default/functions.lua
```diff
@@ -715,9 +715,40 @@ function default.register_craft_metadata_copy(ingredient, result)
...
+local non_player_action_log = minetest.settings:get_bool("enable_non_player_action_log") ~= false
+
+function default.log_player_action(player, pos, message)
+	local who = player:get_player_name()
+	if not player.is_fake_player and player:is_player() then
+		-- log action of real player
+		minetest.log("action",  who .. " " .. message .. " at " .. minetest.pos_to_string(pos))
+	elseif non_player_action_log then
+		-- log action of non real player
+		who = who .. "(" .. (type(player.is_fake_player) == "string"
```
sfan5, line 731: I can guess what this does but it's not documented...

mods/default/functions.lua
```diff
@@ -715,9 +715,40 @@ function default.register_craft_metadata_copy(ingredient, result)
...
+		default.log_player_action(player, pos, "moves " .. stack:get_name() .. " to " .. name)
+	end
+	def.on_metadata_inventory_take = function(pos, listname, index, stack, player)
+		default.log_player_action(player, pos, "takes " .. stack:get_name() .. " from " .. name)
+	end
+end
 
 --
--- NOTICE: This method is not an official part of the API yet.
+-- NOTICE: This method are not an official part of the API yet.
```
sfan5, line 751:
```suggestion
-- NOTICE: This method is not an official part of the API yet.
```

game_api.txt
```diff
@@ -1115,3 +1115,28 @@ This function registers a shapeless recipe that takes `ingredient`
...
+
+`default.log_player_action(player, pos, message)`
+
+ * `player`  The player who performed the action
+ * `pos`     Position of the node at which the action was performed
+ * `message` A message describing the action
```
sfan5, line 1134:
```suggestion
 * `message` A message describing the action in 3rd person singular present tense.
```

sfan5:
The messages in `bones` don't need to be changed, fake players can't die.
Same for `give_initial_stuff` or `creative`.

You missed a few others though:
<pre>
mods/farming/api.lua
181:	minetest.log("action", player_name .. " places node " .. plantname .. " at " ..
182-		minetest.pos_to_string(pt.above))
mods/default/trees.lua
575:	minetest.log("action", player_name .. " places node "
576-			.. sapling_name .. " at " .. minetest.pos_to_string(pos))
mods/tnt/init.lua
468:			minetest.log("action", puncher:get_player_name() ..
469-				" ignites tnt:gunpowder at " ..
470-				minetest.pos_to_string(pos))
--
638:					minetest.log("action", puncher:get_player_name() ..
639-						" ignites " .. node.name .. " at " ..
640-						minetest.pos_to_string(pos))
mods/default/nodes.lua
2618:			minetest.log("action", player_name .. " wrote \"" .. text ..
2619-				"\" to the sign at " .. minetest.pos_to_string(pos))
</pre>

Should more flexibility for log messages be needed don't forget that `string.format` exists or you could do [something like this](https://github.com/minetest/minetest/blob/5.5.0/builtin/game/chat.lua#L5-L42). No need for primitive solutions (append a string at the front, in the middle, at the end, ...).

> I'm not sure about changing log messages because it makes grepping harder while the logs contain messages in both formats 

By the way this is already the case with the current state of the PR:
<i>Old</i>: username puts stuff to bookshelf at (1,2,3)
<i>New</i> username moves example:item to bookshelf at (1,2,3)

mods/default/functions.lua
nixnoxus: gane_api.txt updated: https://github.com/minetest/minetest_game/pull/2941/commits/26553bfaee20f92deadcb748942010a2ea890ef8

game_api.txt
nixnoxus: I aggree.  Adapted in https://github.com/minetest/minetest_game/pull/2941/commits/90591c5a49972fac888da4fb2f0ceaf3237b8526

nixnoxus:
okay, if we hide `bones`, `give_initial_stuff` and `creative`, then most of the other messages will fit into the format:
`player .. message .. " at " .. pos`.

One exception is in ./mods/default/nodes.lua:
_old_: USERNAME tried to place ITEM at protected position (1,2,3)
_new_: USERNAME tried to place ITEM at protected position at (1,2,3)
See https://github.com/minetest/minetest_game/pull/2941/commits/86a4b7ccb710a87a6df33e7b707540b20d992659

I would like to avoid the complexity of string.format, because the function does not expect arbitrary arguments.

Another reason: I would like to use this function in other mods like this:
```lua
local L = minetest.global_exists("default.log_player_action") and
        default.log_player_action or
        function (player, message, pos)
                minetest.log("action", player:player_name
                        .. " " .. message
                        .. " at " minetest.pos_to_string(pos))
        end
...
L(player, "does something", pos)
```

appgurueu:
> One exception is in [...]

Pretty minor, probably fine.

> I would like to avoid the complexity of string.format, because the function does not expect arbitrary arguments.

Please elaborate. What would be wrong with passing a format string? That might allow dealing with the one exception.

> Another reason: I would like to use this function in other mods like this:
> 
> ```lua
> local L = minetest.global_exists("default.log_player_action") and
>         default.log_player_action or
>         function (player, message, pos)
>                 minetest.log("action", player:player_name
>                         .. " " .. message
>                         .. " at " minetest.pos_to_string(pos))
>         end
> ...
> L(player, "does something", pos)
> ```

And why can't you do that with a format string? Also note that this code is buggy as `global_exists` ignores the dot; it ought to be just `minetest.global_exists"default" and default.log_player_action`, but if you're duplicating the code anyways, there's little point in defaulting to the `default` implementation unless you want modders to be able to override it.

mods/default/functions.lua
```diff
@@ -715,6 +715,37 @@ function default.register_craft_metadata_copy(ingredient, result)
...
+function default.log_player_action(player, message, pos)
+	local who = player:get_player_name()
+	if not player.is_fake_player and player:is_player() then
+		-- log action of real player
+		minetest.log("action",  who .. " " .. message .. " at " .. minetest.pos_to_string(pos))
+	elseif log_non_player_actions then
+		-- log action of non real player
+		who = who .. "(" .. (type(player.is_fake_player) == "string"
+			and player.is_fake_player or "*").. ")"
+		minetest.log("action",  who .. " " .. message .. " at " .. minetest.pos_to_string(pos))
```
appgurueu, line 733: You may move the call to `minetest.log` to the bottom of the function to deduplicate it.

appgurueu (approved):

mods/default/functions.lua
```diff
@@ -719,7 +719,7 @@ end
 -- Log API / helpers
 --
 
-local log_non_player_actions = minetest.settings:get_bool("log_non_player_actions") ~= false
+local log_non_player_actions = minetest.settings:get_bool("log_non_player_actions") or false
```
appgurueu, line 722: Isn't this superfluous code? The docs for `get_bool` state the following:

> `default` is the value returned if `key` is not found

Considering that the default `false` is already set in `settingtypes.txt`, you don't need to specify it here.

nixnoxus:
> > One exception is in [...]
> 
> Pretty minor, probably fine.
> 
> > I would like to avoid the complexity of string.format, because the function does not expect arbitrary arguments.
> 
> Please elaborate. What would be wrong with passing a format string? That might allow dealing with the one exception.

We have 2 objects (player and pos) as argument and possibly further arguments.

1st possible version with placeholder
```lua
-- 1st argument: player object
-- 2nd argument: pos object
-- 3. argument: message format
-- 4 ... <N> argument(s) (optional): message arguments
default.log_player_action(player, pos, "@player places node @1 at @pos", sapling_name)
default.log_player_action(player, pos, "@player ignites tnt:gunpowder at @pos")
```
--> i find the argument list counterintuitive



2nd possible version with `string.format`
```lua
-- 1st argument: player object
-- 2nd argument: message format
-- 3 ... <N> argument(s) (optional): message arguments
default.log_player_action(player, "places node %s at %s", sapling_name, minetest.pos_to_string(pos))
default.log_player_action(player, "fires tnt:gunpowder at %s", minetest.pos_to_string(pos))
```
--> would be better



current version:
```lua
default.log_player_action(player, "places node " .. sapling_name, pos)
default.log_player_action(player, "ignites tnt:gunpowder", pos)
```
--> KISS version, which I like :heart: 

Do you maybe have another nice version?



> > Another reason: I would like to use this function in other mods like this:
> > ```lua
> > local L = minetest.global_exists("default.log_player_action") and
> >         default.log_player_action or
> >         function (player, message, pos)
> >                 minetest.log("action", player:player_name
> >                         .. " " .. message
> >                         .. " at " minetest.pos_to_string(pos))
> >         end
> > ...
> > L(player, "does something", pos)
> > ```
> 
> And why can't you do that with a format string? Also note that this code won't work as `global_exists` ignores the dot and would return `false`; it ought to be just `minetest.global_exists"default" and default.log_player_action`, but if you're duplicating the code anyways, there's little point in defaulting to the `default` implementation unless you want modders to be able to override it.

Yes sorry, the sample code was broken.
The idea is to use the `default` version when `minetest_game` is enabled. In some mods `minetest_game` is an optional dependency.

mods/default/functions.lua
nixnoxus: `minetest.settings:get_bool("log_non_player_actions")` returns `nil` if not set. So i want to make sure that the value is a bool

appgurueu:
> `default.log_player_action(player, "places node %s at %s", sapling_name, minetest.pos_to_string(pos))`

If `pos` is a vector with a metatable (as created by `vector.new`), you don't need to call `minetest.pos_to_string`:

Bildschirmfoto von 2022-04-22 17-18-39

mods/default/functions.lua
nixnoxus: ah sorry, now i see what you mean and fix this ....

nixnoxus: fixed in https://github.com/minetest/minetest_game/pull/2941/commits/fc6e5944a21a3ec4c36310cb3de8ef8caf4f8e91

nixnoxus:
> Your concatenation-based version is fine by me. There would be other options, such as passing just a list of items to be concatenated with spaces:

i like your suggestion. however, for this we need a function `is_vector`. I'll give it a try ...

appgurueu:
> > Your concatenation-based version is fine by me. There would be other options, such as passing just a list of items to be concatenated with spaces:
> 
> i like your suggestion. however, for this we need a function `is_vector`. I'll give it a try ...

You can use `vector.check(v)`.

nixnoxus:
> > > Your concatenation-based version is fine by me. There would be other options, such as passing just a list of items to be concatenated with spaces:
> > 
> > 
> > i like your suggestion. however, for this we need a function `is_vector`. I'll give it a try ...
> 
> You can use `vector.check(v)`; you don't need to use this though, as these vectors will have `__tostring` set in their metatable, hence `tostring(v)` will be roughly the same as `minetest.pos_to_string(v)`.

 `vector.check(v)` is available since minetest 5.5.0. I use these when available.
(not mentioned in 5.4.1: https://github.com/minetest/minetest/blob/5.4.1/doc/lua_api.txt#L7613)

Message parts was a great idea, Thanks!

mods/default/functions.lua
```diff
@@ -721,16 +721,27 @@ end
 
 local log_non_player_actions = minetest.settings:get_bool("log_non_player_actions", false)
 
-function default.log_player_action(player, message, pos)
-	local who = player:get_player_name()
+-- `vector.check` is available since minetest >= 5.5.0
```
appgurueu, line 724: Since the master of MTG always targets the most recent version (in fact even the dev version), you can rely on `vector.check` and don't need to duplicate it here AFAIK. @sfan5

mods/default/functions.lua
```diff
@@ -724,9 +724,9 @@ local log_non_player_actions = minetest.settings:get_bool("log_non_player_action
 -- `vector.check` is available since minetest >= 5.5.0
 local is_vector = vector.check or function(v)
 	return v and type(v) == "table" and
-		v.x or type(v.x) == "number" and
-		v.y or type(v.y) == "number" and
-		v.z or type(v.z) == "number"
+		v.x and type(v.x) == "number" and
```
appgurueu, line 727: You don't need to check for `v.x`; if `v.x` is `nil`, `type(v.x)` will be `"nil" ~= "number"`.

nixnoxus: Thanks, fixed in https://github.com/minetest/minetest_game/pull/2941/commits/76c67ac84eeedebcdb3b7dcc02c8965faaf21d65

mods/default/functions.lua
sfan5: Correct.

game_api.txt
```diff
@@ -1115,3 +1115,33 @@ This function registers a shapeless recipe that takes `ingredient`
...
+These non-players are marked by the content of `is_fake_player`
+(if it is a string) or a "*" in brackets after the player name in
+the log.
+
+`default.log_player_action(player, ...)`
+
+ * `player`        The player who performed the action
+ * `message_parts` Any mumber of message parts describing the action
+                   in 3rd person singular present tense. It can also
+		   contain a `vector` which is logged as " at (X,Y,Z)"
```
sfan5, line 1139: doing this is inflexible and error-prone

you even made a mistake in this PR:
<pre>
default.log_player_action(placer,
  "tried to place", itemstack:get_name(),
  "at protected position", pos_under)
</pre>
expands to "foo tried to place mod:item at protected position **at** (1,2,3)"

nixnoxus: not really a mistake (see https://github.com/minetest/minetest_game/pull/2941#issuecomment-1105221099 & https://github.com/minetest/minetest_game/pull/2941#issuecomment-1106347160) but fixed in https://github.com/minetest/minetest_game/pull/2941/commits/589bc4e5a4e34014efed7604d3f21e4fb862764f  :-)

Another possible error was a message like:
```lua
minetest.log("action", player_name .. " dies at " .. pos_string .. ". Inventory dropped")
--> "PLAYERNAME dies at (1,2,4). Inventory dropped"
```
vs.
```lua
default.log_player_action(player, "dies at ", pos, ". Inventory dropped")
--> "PLAYERNAME dies at (1,2,4) . Inventory dropped"
```
fixed in https://github.com/minetest/minetest_game/pull/2941/commits/238a93fcb353627b507315960b36a6152a7e93e9
```lua
default.log_player_action(player, "dies at ", pos, ". Inventory dropped")
--> "PLAYERNAME dies at (1,2,4). Inventory dropped"
```

game_api.txt
```diff
@@ -1115,3 +1115,33 @@ This function registers a shapeless recipe that takes `ingredient`
...
+These non-players are marked by the content of `is_fake_player`
+(if it is a string) or a "*" in brackets after the player name in
+the log.
+
+`default.log_player_action(player, ...)`
+
+ * `player`        The player who performed the action
+ * `message_parts` Any mumber of message parts describing the action
+                   in 3rd person singular present tense. It can also
+		   contain a `vector` which is logged as "(X,Y,Z)"
```
sfan5, line 1139:
```suggestion
                   contain a `vector` which is logged as "(X,Y,Z)"
```

game_api.txt
```diff
@@ -1115,3 +1115,33 @@ This function registers a shapeless recipe that takes `ingredient`
...
+
+`default.set_inventory_action_loggers(def, name)`
+
+ * sets the callbacks `on_metadata_inventory_move`,
+   `on_metadata_inventory_put` and `on_metadata_inventory_take`
+   log corresponding actions
```
sfan5, line 1145:
```suggestion
   that log corresponding actions
```

mods/default/functions.lua
```diff
@@ -715,6 +715,48 @@ function default.register_craft_metadata_copy(ingredient, result)
...
 end
 
+--
+-- Log API / helpers
+--
+
+local log_non_player_actions = minetest.settings:get_bool("log_non_player_actions", false)
+
+-- `vector.check` is available since minetest >= 5.5.0
+local is_vector = vector.check or function(v)
```
sfan5, line 725: You can directly use `vector.check` everywhere, backwards compatibility is not necessary.

mods/default/functions.lua
```diff
@@ -715,6 +715,48 @@ function default.register_craft_metadata_copy(ingredient, result)
...
+end
+
+function default.log_player_action(player, ...)
+	local msg = player:get_player_name()
+	if player.is_fake_player or not player:is_player() then
+		if not log_non_player_actions then
+			return
+		end
+		msg = msg .. "(" .. (type(player.is_fake_player) == "string"
+			and player.is_fake_player or "*").. ")"
```
sfan5, line 737:
```suggestion
			and player.is_fake_player or "*") .. ")"
```

mods/default/functions.lua
```diff
@@ -715,6 +715,48 @@ function default.register_craft_metadata_copy(ingredient, result)
...
+	end
+	for _, v in ipairs({...}) do
+		-- translate vectors
+		local part = is_vector(v) and minetest.pos_to_string(v) or v
+		-- no leading spaces before punctuation marks
+		msg = msg .. (string.match(part, "^[;,.]") and "" or " ") .. part
```
sfan5, line 743: I don't insist on removing this but since it's not needed inside MTG removing the string.match would simplify this

sfan5:
:+1 Works fine other than that.

mods/default/functions.lua
nixnoxus: okay, then i have to update my (test-) minetest server from 5.4.1 to 5.5.0 first ...

nixnoxus: In many places in minetest_game we still work with pos as a table (as described in https://github.com/minetest/minetest_game/blob/edd033b708488472c6bd688b14d8a5f59ef5b218/game_api.txt#L16). 

You can see it with `grep -rF '{x ' .`.

The use of `vector.check` breaks the previous compatibility to minetest 5.4.1.  In the README.md it should be pointed out that at least minetest 5.5.0 is needed.

Under these circumstances, I would find it more consistent and cleaner to do without `vector.check` and rename the function `is_vector(v)` to `is_pos(pos)`.

I think a general change from "pos" to "vector" should be done separately.

However, if we stick to `vector.check`, the use of `vector.to_string` instead of `minetest.pos_to_string` would be obvious. However, the messages would change here (space after comma):
`minetest.pos_to_string` returns `(X,Y,Z)`
`vector.to_string` returns `(X,Y,Z)`

Do you still insist on `vector.check` and what do you think about `vector.to_string`?

mods/default/functions.lua
nixnoxus: If possible, i would like to keep it that way since i want to use this function in other mods in the future.

mods/default/functions.lua
sfan5: > In many places in minetest_game we still work with pos as a table

Does `vector.check` not work on them?

> The use of vector.check breaks the previous compatibility to minetest 5.4.1. In the README.md it should be pointed out that at least minetest 5.5.0 is needed.

The readme is already correct on this. Commits merged into MTG *git* do not need to be compatible with MT *stable*.

> what do you think about vector.to_string?

I think it's better to keep the logging format without spaces.

nixnoxus: >>  In many places in minetest_game we still work with pos as a table

> Does `vector.check` not work on them?

An error would only occur if you pass pos as table like:
```lua
default.log_player_action(player, "does something", {x = 1, y = 2, z = 3})
```
But we don't have such a constellation in minetest_game at the moment.

I just find it, as mentioned above, a bit inconsistent that the function expects a vector instead of pos as a table.
But i have now adjusted it as desired: https://github.com/minetest/minetest_game/pull/2941/commits/1ba876893dc9f9c37e8c436cddda4c56bce1a1ab ;-)

nixnoxus: I am very happy with this https://github.com/minetest/minetest_game/pull/2941/commits/a6d940abda11e09342685893ec28b1acce20cf45 :sweat_smile:

PR #2942, opened 2022-04-02
Don't block fluid (water, lava) flow by flowers, mushrooms, grass and saplings

bartoszkosiorek:
In previous implementation, little flower or mushroom was barieer
by which water was not able to pass.

With this implementation water is able to pass through flowers
by destroying it and also is allowing to pick up these flowers.

### DIFF
--- a/mods/flowers/init.lua
+++ b/mods/flowers/init.lua
@@ -53,6 +53,8 @@ local function add_simple_flower(name, desc, box, f_groups)
 		buildable_to = true,
 		groups = f_groups,
 		sounds = default.node_sound_leaves_defaults(),
+		floodable = true,
+		on_flood = default.drop_on_flood,
 		selection_box = {
 			type = "fixed",
 			fixed = box
@@ -197,6 +199,8 @@ minetest.register_node("flowers:mushroom_red", {
 	groups = {mushroom = 1, snappy = 3, attached_node = 1, flammable = 1},
 	sounds = default.node_sound_leaves_defaults(),
 	on_use = minetest.item_eat(-5),
+	floodable = true,
+	on_flood = default.drop_on_flood,
 	selection_box = {
 		type = "fixed",
 		fixed = {-4 / 16, -0.5, -4 / 16, 4 / 16, -1 / 16, 4 / 16},
@@ -216,6 +220,8 @@ minetest.register_node("flowers:mushroom_brown", {
 	groups = {mushroom = 1, food_mushroom = 1, snappy = 3, attached_node = 1, flammable = 1},
 	sounds = default.node_sound_leaves_defaults(),
 	on_use = minetest.item_eat(1),
+	floodable = true,
+	on_flood = default.drop_on_flood,
 	selection_box = {
 		type = "fixed",
 		fixed = {-3 / 16, -0.5, -3 / 16, 3 / 16, -2 / 16, 3 / 16},
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -176,6 +176,15 @@ if minetest.settings:get_bool("enable_lavacooling") ~= false then
 	})
 end
 
+--
+-- Drop items on flooding
+--
+
+function default.drop_on_flood(pos, oldnode, newnode)
+    for _, item in ipairs(minetest.get_node_drops(oldnode)) do
+        minetest.add_item(pos, item)
+    end
+end
 
 --
 -- Optimized helper to put all items in an inventory into a drops list
--- a/mods/default/nodes.lua
+++ b/mods/default/nodes.lua
@@ -718,6 +718,8 @@ minetest.register_node("default:sapling", {
 	sunlight_propagates = true,
 	walkable = false,
 	on_timer = grow_sapling,
+	floodable = true,
+	on_flood = default.drop_on_flood,
 	selection_box = {
 		type = "fixed",
 		fixed = {-4 / 16, -0.5, -4 / 16, 4 / 16, 7 / 16, 4 / 16}
@@ -881,6 +883,8 @@ minetest.register_node("default:junglesapling", {
 	sunlight_propagates = true,
 	walkable = false,
 	on_timer = grow_sapling,
+	floodable = true,
+	on_flood = default.drop_on_flood,
 	selection_box = {
 		type = "fixed",
 		fixed = {-4 / 16, -0.5, -4 / 16, 4 / 16, 7 / 16, 4 / 16}
@@ -917,6 +921,8 @@ minetest.register_node("default:emergent_jungle_sapling", {
 	sunlight_propagates = true,
 	walkable = false,
 	on_timer = grow_sapling,
+	floodable = true,
+	on_flood = default.drop_on_flood,
 	selection_box = {
 		type = "fixed",
 		fixed = {-4 / 16, -0.5, -4 / 16, 4 / 16, 7 / 16, 4 / 16}
@@ -995,6 +1001,8 @@ minetest.register_node("default:pine_sapling", {
 	sunlight_propagates = true,
 	walkable = false,
 	on_timer = grow_sapling,
+	floodable = true,
+	on_flood = default.drop_on_flood,
 	selection_box = {
 		type = "fixed",
 		fixed = {-4 / 16, -0.5, -4 / 16, 4 / 16, 7 / 16, 4 / 16}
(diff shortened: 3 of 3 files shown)


appgurueu:
Why are you removing the `flammable` group? Setting flowers on fire should be left unchanged.

bartoszkosiorek:
I didn't removed flammable group, I just make it default to 1.

appgurueu:
> I didn't removed flammable group, I just make it default to 1.

Not quite a "default" as it overrides whatever is in the def but yes, I overlooked that. Sorry.

mods/flowers/init.lua
```diff
@@ -31,6 +31,12 @@ minetest.register_alias("flowers:flower_viola", "flowers:viola")
 minetest.register_alias("flowers:flower_dandelion_white", "flowers:dandelion_white")
 
 
+local function on_flood(pos, oldnode, newnode)
+	minetest.add_item(pos, ItemStack(oldnode.name))
+	-- Remove flower
+	return false
```
appgurueu, line 37: According to the docs you don't need to `return false` here.

bartoszkosiorek: True fixed

appgurueu (approved):
Looks fine. PR title is misleading as the flowers are dropped, not destroyed.

Desour:
* For consistency, this should also be done for grass and other shrubs, or not at all, imo.
* You might want to drop the drops and not the nodes themselves.
* It would be nice if the drop pos was randomized a bit.

bartoszkosiorek:
> * For consistency, this should also be done for grass and other shrubs, or not at all, imo.
done
> * You might want to drop the drops and not the nodes themselves.
all flooded things producing drops.
> * It would be nice if the drop pos was randomized a bit.
I would like to keep alghoritm simple and fast. Maybe it is not worth to implement that. Do you have example how it could be done?

appgurueu:
> I would like to keep alghoritm simple and fast. Maybe it is not worth to implement that. Do you have example how it could be done?

Performance impact is almost negligible. It might looks as follows:
```lua
local cube_size = 0.6
local offset = (1 - cube_size) / 2
local function randomize_drop_pos(pos)
    return vector.add(vector.add(vector.multiply(vector.new(math.random(), math.random(), math.random()), cube_size), offset), pos)
end
```
or using signed random (perhaps more readable?):
```lua
local cube_size = 0.6
local function sigrand()
    return 2*math.random() - 1
end
local function randomize_drop_pos(pos)
    return vector.add(vector.multiply(vector.new(sigrand(), sigrand(), sigrand()), cube_size), pos)
end
```

appgurueu:
> All flooded things are producing drops.

What @Desour was saying is that when you generalize this to all MTG plant nodes, you should be determining the dropped item over the node def "drop" field using `minetest.get_node_drops(oldnode)`.

Which leaves me with an API suggestion: What about a `default.def.floodable_drop` function which extends & returns the node def by setting `floodable = true` and the appropriate `on_flood` func:

```lua
default.def = {}
local function drop_on_flood(pos, oldnode, newnode)
    for _, item in ipairs(minetest.get_node_drops(oldnode)) do
        minetest.add_item(randomize_drop_pos(pos), item)
    end
end
function default.def.floodable_drop(def)
    def.floodable = true
    def.on_flood = drop_on_flood
end
```

which you can then simply use in `register_flower` etc. if you want all flowers to be dropped on flooding.

bartoszkosiorek:
Unfortunately after changing implementation to:
```
local function on_flood_drop_item(pos, oldnode, newnode)
	minetest.get_node_drops(oldnode)
end
```
or
```
local function on_flood_drop_item(pos, oldnode, newnode)
	minetest.get_node_drops(oldnode.name, "")
end
```
there is no more drops.

It seems that it is not initialized here.

appgurueu:
> Unfortunately after changing implementation to:
> 
> ```
> local function on_flood_drop_item(pos, oldnode, newnode)
> 	minetest.get_node_drops(oldnode)
> end
> ```
> 
> or
> 
> ```
> local function on_flood_drop_item(pos, oldnode, newnode)
> 	minetest.get_node_drops(oldnode.name, "")
> end
> ```
> 
> there is no more drops.
> 
> It seems that it is not initialized here.

You need to add the drops using `add_item` as I have shown in my second snippet.

bartoszkosiorek:
Could you please test it with Lava and Water?
It seems that after ading randomization, sometimes the drops disapear under ground.

appgurueu:
> Could you please test it with Lava and Water?

Lava is interesting as it should both flood the crops *and* set them on fire; this might lead to a race condition if the ABM runs before `on_flood` I suppose. Perhaps destroy (and don't drop) the plant if the new node has the `igniter` group set? Shouldn't matter as the drops will be destroyed by the lava though.

> It seems that after ading randomization, sometimes the drops disapear under ground.

It is possible that `0.4` extent (`0.8`³) is too large a size for the cube. Try smaller values (say, `0.25` (`0.4`³ cube)).

bartoszkosiorek:
> Lava is interesting as it should both flood the crops _and_ set them on fire; this might lead to a race condition if the ABM runs before `on_flood` I suppose. Perhaps destroy (and don't drop) the plant if the new node has the `igniter` group set? Shouldn't matter as the drops will be destroyed by the lava though.

From my testing it seems that crops are destroyed/dropped instantly, and there is no time for burn.
It also looks very realistic, as after few seconds, the drops are burned.

appgurueu:
What about griefers? Should there be a `minetest.is_protected` check in the `on_flood` callback?

bartoszkosiorek:
> What about griefers? Should there be a `minetest.is_protected` check in the `on_flood` callback?

Unfortunately I am completely new with griefers. Could you please explain a little more?

appgurueu:
> Unfortunately I am completely unexperienced with griefers. Could you please explain a little more how we should improve it?

Say you have a protected area including a flower bed / mushroom farm / grass / tree farm (saplings). Griefers might now use liquids to destroy (flood) your beautiful farm as the liquids don't care about protection.

A simple fix might be checking for protection, but the problem here is that liquids carry no owner/source information, which would make crops entirely unfloodable in protected areas...

bartoszkosiorek:
Ok, I get it. 
The floodable items are waterlilly, butterflies, torch, fire, fireflies and snow. There is no any protection for it.
Is that mean that currently someone could easily destroy above by using water?
Maybe water should not flow in protected areas?

bartoszkosiorek:
I have disabled randomization, as it looks weird. Instead I would like to implement moving dropping by water flow. I will do that in next Pull Request.

appgurueu:
> Instead I would like to implement moving dropping by water flow. I will do that in next Pull Request.

Please think it through first! A clean implementation definitely requires an engine change (to obtain liquid flow direction), a hacky one requires a significant amount of code checking all flowingliquid neighbors for each "flowing" item every step, which I imagine might eventually be performance critical.

Desour:
Protection checks seem unnecessary (and also weird).
Griefers can also let lava and water flow into your area to fill it with stone (if the server allows placing lava), or they can build up into the sky and let sand drop or build platforms to make shadow. If you want to protect your flowers, can build a trench or similar at the border of your protected area.

bartoszkosiorek:
Is it possible to merge this PR?

I have reached my personal goals with this implementation:
- simple implementation (easy to understand)
- easy to extend (it is an example for mods, how it could be implemented for other nodes)
- minor impact on performance

bartoszkosiorek:
Fixed duplication of `on_flood_drop` function.

How many approvals need to be get to be able to merge PR?

appgurueu:
> How many approvals need to be get to be able to merge PR?

You need two approvals.

sfan5:
It's unfortunate to have to ask this question now that the PR is finished but: Since Minetest Game is not accepting new features, shouldn't this be rejected? (#2710 for more detail)
Worth noting that I think this would be a good addition, but no features means no features.

bartoszkosiorek:
@sfan5 I think it is not a feature, as the flooding is working for example for torch and fire. It is just align the same behaviour with the rest of `minetest_game`.

bartoszkosiorek:
I renamed the title to show what is the real issue with water/lava flow.

Could you please change the PR to Bug Fix?

sfan5:
This was discussed in a coredev meeting we just had and the we weren't too sure either but the conclusion was that it is too close to being feature.
Consider making a mod, sorry.

PR #2947, opened 2022-04-24
Mapgen: Don't clear biomes, decorations and ores

appgurueu:
I might be unaware of the intention behind this clearing, but for *vanilla MTG it should be a no-op* as nothing is registered before. For modded MTG however this had the potential to *silently unregister the biomes, decorations and ores of mods that don't depend on default and are thus loaded before it*.
For decorations and ores in particular this should most likely not be done; for biomes I don't see the point either, but there *might be a point I'm missing*.

`git blame` yields https://github.com/minetest/minetest_game/commit/2e950ac6 which provides no reason for the clearings either. @paramat please explain.

### DIFF
--- a/mods/default/mapgen.lua
+++ b/mods/default/mapgen.lua
@@ -2476,9 +2476,6 @@ end
 -- Detect mapgen to select functions
 --
 
-minetest.clear_registered_biomes()
-minetest.clear_registered_ores()
-minetest.clear_registered_decorations()
 
 local mg_name = minetest.get_mapgen_setting("mg_name")
 


sfan5:
Does this change the map generated for a given seed in any way?
Does the engine have any defaults here?

appgurueu:
> Does this change the map generated for a given seed in any way?

According to very brief testing it does not.

> Does the engine have any defaults here?

I don't think so. Inserting
```lua
assert(next(minetest.registered_biomes) == nil)
assert(next(minetest.registered_ores) == nil)
assert(next(minetest.registered_decorations) == nil)
```
throws no error.

Montandalar:
I feel like changing this would help more mapgen mod authors than it could possibly harm. With no biomes built into the engine, I don't know why you would clear them in MTG.

appgurueu:
@sfan5 @SmallJoker objections to merging this?

SmallJoker (approved):
There is the possibility that a mod that registers biomes/ores/decorations does neither directly nor indirectly depend on `default`. In that case, it could be loaded before `default`, thus silently overwritten. After this PR, this would generate new, differently looking mapblocks.
The overall chance for that to happen is however very small. The introduces consequences in that case do however look reasonable and acceptable to me.

PR #2951, opened 2022-05-06
Carts: Improve movement behaviour

### LINKED ISSUE #2948: Minecart Problem
The minecart passes through the railway when it goes up. It can not go up correctly.

SmallJoker:
Backported various changes from boost_cart
Improved direction handling and code cleanup
Smoothed-out 'end of rail' animation

Partial fix for #2948 - the cart is now stopping in a horizontal direction. The problem here is that the next node (after downhill turn-back) is on the same Y level as the cart, hence there's straight rail friction which makes the cart stop immediately.

This far from perfect but aimed to provide at least a little bit better experience until a bullet-proof solution is found.

### DIFF
--- a/mods/carts/cart_entity.lua
+++ b/mods/carts/cart_entity.lua
@@ -47,18 +47,12 @@ function cart_entity:on_activate(staticdata, dtime_s)
 	end
 	self.railtype = data.railtype
 	self.old_dir = data.old_dir or self.old_dir
-	self.old_pos = data.old_pos or self.old_pos
-	-- Correct the position when the cart drives further after the last 'step()'
-	if self.old_pos and carts:is_rail(self.old_pos, self.railtype) then
-		self.object:set_pos(self.old_pos)
-	end
 end
 
 function cart_entity:get_staticdata()
 	return minetest.serialize({
 		railtype = self.railtype,
-		old_dir = self.old_dir,
-		old_pos = self.old_pos
+		old_dir = self.old_dir
 	})
 end
 
--- a/mods/carts/cart_entity.lua
+++ b/mods/carts/cart_entity.lua
@@ -46,15 +46,19 @@ function cart_entity:on_activate(staticdata, dtime_s)
 		return
 	end
 	self.railtype = data.railtype
-	if data.old_dir then
-		self.old_dir = data.old_dir
+	self.old_dir = data.old_dir or self.old_dir
+	self.old_pos = data.old_pos or self.old_pos
+	-- Correct the position when the cart drives further after the last 'step()'
+	if self.old_pos and carts:is_rail(self.old_pos, self.railtype) then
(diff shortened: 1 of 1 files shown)


mods/carts/cart_entity.lua
```diff
@@ -46,15 +46,19 @@ function cart_entity:on_activate(staticdata, dtime_s)
...
+		self.object:set_pos(self.old_pos)
 	end
 end
 
 function cart_entity:get_staticdata()
 	return minetest.serialize({
 		railtype = self.railtype,
-		old_dir = self.old_dir
+		old_dir = self.old_dir,
+		old_pos = self.old_pos
```
appgurueu, line 61: I don't feel quite comfortable with serializing old cart positions here... If data is lost (due to a hard crash), the staticdata will be outdated, potentially causing the cart to teleport. Serializing the direction isn't an issue, because that can be properly updated based on the rails.

mods/carts/cart_entity.lua
```diff
@@ -192,11 +196,11 @@ local function rail_on_step(self, dtime)
 	end
 
 	local pos = self.object:get_pos()
-	local cart_dir = carts:velocity_to_dir(vel)
-	local same_dir = vector.equals(cart_dir, self.old_dir)
+	local dir = carts:velocity_to_dir(vel)
+	local dir_changed = not vector.equals(dir, self.old_dir)
```
appgurueu, line 200: I found `dir_same` more readable. What about `dir_unchanged`?

mods/carts/cart_entity.lua
SmallJoker: Good point. How could rail-jumping be avoided then, if not by teleporting to the last safe position? The rail type underneath the cart may chance (or there's no rail at all), resulting in out-of-place carts.

appgurueu: > Good point. How could rail-jumping be avoided then, if not by teleporting to the last safe position? The rail type underneath the cart may chance (or there's no rail at all), resulting in out-of-place carts.
> 
> EDIT: I suppose the engine could be blamed in this case.

Well, while the entity step is running, the cart may teleport to the latest rail pos in order to not overshoot. This may only go for a single entity step though and *may not be persisted* in particular to avoid larger teleportation.

SmallJoker: Whatever. I removed the code.

mods/carts/cart_entity.lua
SmallJoker: Subjective matter. I'd like to avoid negations within variable names to introduce another lever of confusion. At the end of the day it does not matter at all.

appgurueu (approved):
From a quick glance the code looks fine to me. Cart movement works fine.

appgurueu:
Deleted the branch.

PR #2958, opened 2022-06-01
Carts: Get rid of animation hack

### LINKED ISSUE #2950: Cart: Get rid of animation hack
Carts currently use a B3D model and an animation for non-yaw (pitch) rotation. This is an old remnant from 0.4 times when only `set_yaw` existed, but since 5.0 `set_rotation` is available and should be used; the cart should be re-exported as Wavefront `.obj`.

Cleanly reimplementing cart rotation will presumably also allow fixing #2948.

appgurueu:
Fixes #2950; also fixes the visual aspect of #1927 - the player is now rotated:

Rotated player

- [ ] offset the entity by +0.5 if rotated; as a bonus, this will also fix the previous incorrectly offset selection box


sfan5:
I _just_ wanted to give this my approal :thinking: 

Unrelated some smooth animation would be nice in the future, poor Sam must have a headache after riding such a track:
one

SmallJoker:
After the offset is added this will look definitely better than the current one. The patch is surprising simple.

@sfan5 The rotation is actually interpolated, just rather fast. It can also be observed when punching a cart into the other direction due to the direction <---> yaw connection. https://github.com/minetest/minetest/blob/9fc018ded10225589d2559d24a5db739e891fb31/src/client/content_cao.cpp#L1133

sfan5:
What's the status here?

appgurueu:
> What's the status here?

Not finished. I remember this not being as easy as applying the appropriate offset.

PR #2959, opened 2022-06-03
Use add_wear_by_uses to fix incorrect uses counts

### LINKED ISSUE #2929: Hoe usage count is off
The hoe usage count is off.

* Steel Hoe: Expected: 500 uses. Actual: 501 uses
* (legacy) Mese Hoe: Expected: 350 uses. Actual: 351 uses
* (legacy) Diamond Hoe: Expected: 500 uses. Actual: 501 uses
* Wood/Stone Hoe: Are fine (due to coincidence)
* All custom hoes (those registered with the game API) with a high uses count are also affected

The error increases the higher the number of uses (up to tens of thausands).

## Reason for this bug

This is basically https://github.com/minetest/minetest/issues/4900, but in Minetest Game. The same mistake is done as was done in the engine: The naive calculation `add_wear(65535/(uses-1))` that is currently used won't always get in the number of uses you want due to rounding errors. The larger the number of desired uses, the larger the rounding error is (see #11110 for an in-depth analysis).

## Possible fix

Merge https://github.com/minetest/minetest/pull/12047 into Minetest, then replace the naive wear calculation with `item:add_wear_by_uses(number_of_uses_that_you_want_the_tool_to_have_before_it_breaks)`. ;-)

This is a helper function made specifically for this use case.

Wuzzy2:
Fixes #2929.

This replaces the `add_wear` calls by the more modern `add_wear_by_uses` calls which allows to specify the uses count directly. Since this method ensures the correct uses count, it will fix the hoe uses count being slightly off.

Note: The effective uses count of bug net, screwdriver and flint and steel will not change, this is more for consistency.
For example: For flint and steel, adding 1000 wear every time (the old method) is equivalent to 66 uses because with the 66th use, wear would be 66000, which is above 65535. You can do similar calculations for bug net and screwdriver.

### DIFF
--- a/mods/farming/api.lua
+++ b/mods/farming/api.lua
@@ -66,7 +66,7 @@ farming.hoe_on_use = function(itemstack, user, pointed_thing, uses)
 	if not minetest.is_creative_enabled(player_name) then
 		-- wear tool
 		local wdef = itemstack:get_definition()
-		itemstack:add_wear(65535/(uses-1))
+		itemstack:add_wear_by_uses(uses)
 		-- tool break sound
 		if itemstack:get_count() == 0 and wdef.sound and wdef.sound.breaks then
 			minetest.sound_play(wdef.sound.breaks, {pos = pt.above,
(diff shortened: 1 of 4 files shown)


appgurueu (approved):

sfan5 (approved):
Thanks.

PR #2964, opened 2022-06-23
Don't check protection of air when placing bones

OgelGames:
It doesn't make sense to check the protection of air nodes, and it makes bone placement fail more often.

### DIFF
--- a/mods/bones/init.lua
+++ b/mods/bones/init.lua
@@ -224,7 +224,7 @@ minetest.register_on_dieplayer(function(player)
 	-- check if it's possible to place bones, if not find space near player
 	if bones_mode == "bones" and not may_replace(pos, player) then
 		local air = minetest.find_node_near(pos, 1, {"air"})
-		if air and not minetest.is_protected(air, player_name) then
+		if air then
 			pos = air
 		else
 			bones_mode = "drop"


OgelGames:
> Won't this lead people to being able to effectively grief areas which are protected? Filling up a protected spawn with bones by repeatedly killing yourself, or the like.

That was already possible, this just changes it to match the logic in `may_replace()`:
https://github.com/minetest/minetest_game/blob/986b65775f3d7a89959f569c516af8c44b4bc7e3/mods/bones/init.lua#L144-L152

appgurueu:
> That is already possible, this just changes the secondary check to match the logic in may_replace() (which is only called for the initial position)

Why don't you switch the "find nearby pos" to `may_replace()` entirely? Ideally there should IMO be a hierarchy:

1. Try placing bones where the player died;
2. Try nearby positions, checking `may_replace()` - prefer air over buildable nodes, protected over unprotected etc.

OgelGames:
> Why don't you switch the "find nearby pos" to `may_replace()` entirely?

Because that would be more complicated and I just wanted to fix this small mistake without changing too much.

appgurueu (approved):
Probably fixes 99% of cases frequently encountered on servers

PR #2970, opened 2022-07-20
More paper

MidnightPhantom:
It will add more paper from crafting. The paper has only one crafting recipe that is only through papyrus. Plus papyrus spawns only in one biome hence I just changed the crafting count from one to 3

### DIFF
--- a/mods/default/craftitems.lua
+++ b/mods/default/craftitems.lua
@@ -443,7 +443,7 @@ minetest.register_craft({
 })
 
 minetest.register_craft({
-	output = "default:paper",
+	output = "default:paper 3",
 	recipe = {
 		{"default:papyrus", "default:papyrus", "default:papyrus"},
 	}


MidnightPhantom:
Man, I had to do three changes to MTG for the gameplay improvement that I cannot do now as it is only for maintenance. Many creators like me would like to create underground homes with grass and grow crops too, but the light level down is so low we cannot hence we have to bring dozens of meselamps which are hard to craft also hence I just wanted that light levels for growing crops to be changed to 10 (minimum torchlight) and for growing grass to be 4(just like mc). If you allow me I can fork these two things for maintenance as It is totally absurd for the natural growing stuff to require so much of light.

appgurueu:
Consider making a MTG fork with your features.

MidnightPhantom:
So do you know how can I do that

IrrlichtMt4:
By adding or trying to make a file in minetest_game it will automatically make a copy of the same for you to change and add your own features or you can go to https://content.minetest.net and make your own package.

PR #2971, opened 2022-07-22
Update mod translations

Wuzzy2:
This PR does the following:

* Manually move the Japanese translations for the key items from `default` to the `keys` mod (it seems it was accidentally overlooked) (the translations themselves were NOT changed)
* Update the translation files (by using a script). This adds 1 string in `default`
* Update German translation


Wuzzy2:
Sorry, this PR is bad. Forget this one ...

PR #2984, opened 2022-09-09
creative: Override Hand After Mods Loaded

AntumDeluge:
Prevents other mods from overwriting changes & removes dependence on `default`. This allows `creative` to be more of a standalone mod.

### DIFF
--- a/mods/creative/init.lua
+++ b/mods/creative/init.lua
@@ -46,34 +46,36 @@ end
 dofile(minetest.get_modpath("creative") .. "/inventory.lua")
 
 if minetest.is_creative_enabled("") then
-	-- Dig time is modified according to difference (leveldiff) between tool
-	-- 'maxlevel' and node 'level'. Digtime is divided by the larger of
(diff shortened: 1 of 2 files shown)


mods/creative/init.lua
```diff
@@ -55,25 +55,27 @@ if minetest.is_creative_enabled("") then
 	local digtime = 42
 	local caps = {times = {digtime, digtime, digtime}, uses = 0, maxlevel = 256}
```
appgurueu, line 56: Why not move these inside the function to further limit their scope (along with the comment) rather than having them as upvalues?

appgurueu (approved):
Decent solution which will allow the creative mod to be more game-independent. My only concern would be that modders might have their hands overridden, but this is already the case, and this.

mods/creative/init.lua
AntumDeluge: Done. Thanks for pointing that out.

sfan5 (approved):
Looks good.

PR #2994, opened 2022-10-22
Japanese Translation update

SakuraRiu:
Fixed an obvious mistranslation. (one)
As Japanese, I changed the notation to make it easier to see. (many)


mods/bucket/locale/bucket.ja.tr
```diff
@@ -1,5 +1,5 @@
 # textdomain: bucket
-Empty Bucket=空のバケツ
-Water Bucket=水入りバケツ
-River Water Bucket=川の水入りバケツ
-Lava Bucket=溶岩入りバケツ
+Empty Bucket=バケツ（空）
+Water Bucket=バケツ（水）
+River Water Bucket=バケツ（河川水）
+Lava Bucket=バケツ（溶岩）
```
SmallJoker, line 5: Is this notation with parentheses common in Japanese? Is the previous translation wrong?

SakuraRiu: It may not be common. As a matter of fact, I feel that the revision of the translation of Minetest_Game was premature, because I am currently working on the translation of Mineclone2 and have learned that bracketed names do not suit it well. Once the translation of Mineclone2 is completed, I will fix the translation of Minetest_Game better.

PR #3000, opened 2022-11-19
sethome: Fix error if `/home` is executed from irc, discord, etc

zmv7:
Running `/home` by any player in the remote chat utilities(for example IRC, discordmt) was crashing the server if the player with the same name is not on the server.
This bug can be abused on a popular servers if the owners are not experienced enough to fix it themselves

### DIFF
--- a/mods/sethome/init.lua
+++ b/mods/sethome/init.lua
@@ -52,6 +52,9 @@ end
 
 sethome.get = function(name)
 	local player = minetest.get_player_by_name(name)
+	if not player then
+		return nil
+	end
 	local player_meta = player:get_meta()
 	local pos = minetest.string_to_pos(player_meta:get_string("sethome:home"))
 	if pos then
--- a/mods/sethome/init.lua
+++ b/mods/sethome/init.lua
@@ -89,6 +89,10 @@ minetest.register_chatcommand("home", {
 	description = S("Teleport you to your home point"),
 	privs = {home = true},
 	func = function(name)
+		local player = minetest.get_player_by_name(name)
+		if not player then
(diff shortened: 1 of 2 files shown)


mods/sethome/init.lua
```diff
@@ -52,6 +52,9 @@ end
 
 sethome.get = function(name)
 	local player = minetest.get_player_by_name(name)
+	if not player then
+		return nil
```
appgurueu, line 56: `nil`? Consider returning something more helpful like `false, "This command can only be used from in-game!"`

appgurueu (commented):
Is this the only command that fails to do this check?

mods/sethome/init.lua
appgurueu: I don't see the issue with returning a helpful error message.

mods/sethome/locale/sethome.ru.tr
```diff
@@ -6,3 +6,4 @@ Set a home using /sethome=Установите домашнюю точку, ис
 Set your home point=Установите вашу домашнюю точку
 Home set!=Домашняя точка установлена!
 Player not found!=Игрок не обнаружен!
+This command can be executed in-game only!=Эта команда может быть использована только в игре!
```
appgurueu, line 9: This is one of the few strings I would *not* translate, as its purpose is to solely be visible *outside of the game*. Discord / IRC chat bridges don't handle translations.

SmallJoker (approved):
LGTM

sfan5 (approved):

MrRar:
Don't forget to update the translation template.

appgurueu (approved):

PR #3004, opened 2022-11-30
fix registration of craft

alek13:
Walls have no crafts, When register new wall via API from another mod, but in the same namespace (`walls`):
```lua
walls.register(":walls:sandstone_brick", S("Sandstone Brick Wall"), "default_sandstone_brick.png",
		"default:sandstonebrick", default.node_sound_stone_defaults())
```
(for example like in [this mod](https://content.minetest.net/packages/Don/mywalls/))

We should remove `":"` at the beginning of the name.

### DIFF
--- a/mods/walls/init.lua
+++ b/mods/walls/init.lua
@@ -45,6 +45,12 @@ walls.register = function(wall_name, wall_desc, wall_texture_table, wall_mat, wa
 	})
 
 	-- crafting recipe
+	-- HACK:
+	--   Walls have no crafts, when register new wall via API from another mod, but in the same namespace (`walls`).
+	--   So we should remove `":"` at the beginning of the name.
+	if wall_name:sub(1, 1) == ":" then
+		wall_name = wall_name:sub(2)
+	end
 	minetest.register_craft({
 		output = wall_name .. " 6",
 		recipe = {


alek13:
thnx for review.

> ... . If you want to keep the explicit `if`, I'd still apply some minor tweaks to make it more readable:
> 
> ```lua
> if wall_name:sub(1, 1) == ":" then
>     wall_name = wall_name:sub(2)
> end
> ```
fixed to this variant

alek13:
> Could still use a comment though.

Not sure what you mean, add this comment: [changes with comment](https://github.com/minetest/minetest_game/compare/4dd0e2ead697b08db245f2765d00c840bb6e07ac..c629cc3c351c1f781331f1bbe013bb2b3873fd3d)

appgurueu (approved):
Yeah, this is what I meant.

mods/walls/init.lua
```diff
@@ -45,6 +45,12 @@ walls.register = function(wall_name, wall_desc, wall_texture_table, wall_mat, wa
 	})
 
 	-- crafting recipe
+	-- HACK:
```
SmallJoker, line 48: FYI. It's not a hack. It's a node registration convention enforced by `minetest.register_item`, so this change is totally reasonable.

SmallJoker (approved):
LGTM

mods/walls/init.lua
appgurueu: It is inconsistent that `minetest.register_craft` does not enforce the same convention, thus this is a workaround / "hack".

PR #3005, opened 2022-12-28
No more buggy entities

MidnightPhantom:
I have done this so that the falling nodes gravel and silver sand do not fall into the ignore when a player mistakenly punches them at the bottommost limit.

This is what happened when I mistakenly punched it.
screenshot_20221228_124920

Please do not take this as a feature but rather as a maintenance.

### DIFF
--- a/mods/default/mapgen.lua
+++ b/mods/default/mapgen.lua
@@ -542,7 +542,7 @@ function default.register_ores()
 		clust_scarcity  = 16 * 16 * 16,
 		clust_size      = 5,
 		y_max           = 31000,
-		y_min           = -31000,
+		y_min           = -30000,
 		noise_threshold = 0.0,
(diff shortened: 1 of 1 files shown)


appgurueu:
This is an ugly hack. The proper fix is to make falling entities properly deal with the world border. See https://github.com/minetest/minetest/issues/11343.

MidnightPhantom:
I don't understand the words "ugly hack", but I did limit the generation of gravel and silver sand till y level -30000 disabling them to generate near ignore nodes causing this. And also I have done this on the game which I have and then made this pull request.

Well there was a way to remove these unruly entities from the void but it requires mods.

SmallJoker:
This is the wrong solution. Please fix it in minetest/minetest (`__builtin:falling_node`).

Issue #3002, opened 2022-11-25
Door handle of glass door inventory image is on the wrong side


Wuzzy2:
On the inventory image of the glass door and obsidian glass door, the door handle is on the wrong side. The handle is to the left but if you place the door (without any neighboring door), the handle will be at the right.
Interestingly the handle for the steel bar door is correct.

Suggested bugfix: Edit the inventory images to put the handle at the other side.

Loppansson:
Has this been resolved as of "Flip textures of glass door"?:)

appgurueu:
> Has this been resolved as of "Flip textures of glass door"?:)

Yes, thanks for the reminder.

PR #3009, opened 2023-01-23
Flip item textures of glass doors

CrackedMatter:
See #3002


appgurueu (approved):
Minor note: You flipped the textures (rather than only the handles). Not sure if this might cause an inconsistency where the "stripes" go in a different direction :smile:.

CrackedMatter:
> Minor note: You flipped the textures (rather than only the handles). Not sure if this might cause an inconsistency where the "stripes" go in a different direction smile.

The stripes are actually more consistent than before:
screenshot_20230124_091318

sfan5 (approved):

PR #3011, opened 2023-01-31
Clear animation data when appearance is updated

nocturnalwizard:
Discord/#modding

*nil — Today at 8:33 AM
```lua
    if player_data.animation == anim_name and player_data.animation_speed == speed then
        return
    end
```
this check doesn't check whether the model has changed
so you get an early return because it's the same anim with the same speed, but the model has changed

*nil — Today at 8:35 AM
player_data.animation and player_data.animation_speed ought to be cleared in set_model
I'll patch this after lunch

nocwiz — Today at 8:41 AM
Confirmed working

### DIFF
--- a/mods/player_api/api.lua
+++ b/mods/player_api/api.lua
@@ -71,6 +71,10 @@ function player_api.set_model(player, model_name)
 
 	local model = models[model_name]
 	if model then
+		if (model.animations) then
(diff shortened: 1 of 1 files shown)


appgurueu:
Closing in favor of #3012.

PR #3012, opened 2023-01-31
Fix `player_api.set_model` not updating the animation

appgurueu:
Bug reported by @nocturnalwizard. The previous code exhibited the following bug:

1. `character.b3d` gets set as model, the stand animation is played
2. `custom.b3d` gets set; `player_api` tries to set the `stand` animation, but since the animation name & speed are the same this is a no-op and the early return is triggered


sfan5 (approved):
lgtm

Issue #3001, opened 2022-11-24
Ugly opaque leaves: blueberry bush with blueberries, and aspen


Desour:
Screenshot of all leaves in opaque leaves mode (forgot pine needles, but they look fine):
![screenshot_20221124_143821](https://user-images.githubusercontent.com/7613443/203798099-f14cebf8-7086-41f3-bbaf-a9e38fb59b48.png)
Blueberry bush leaves with blueberries (bottom left) and aspen leaves (bottom right) are ugly.
The corresponding textures (`default_blueberry_overlay.png` and `default_aspen_leaves.png`) need better transparent colors.

(Btw. aspen could also use a thicker texture for simple leaves mode.)

Loppansson:
Aspen tree suggestion:
![image](https://github.com/minetest/minetest_game/assets/61879422/8fa5c83d-0914-453b-a247-d2f6b31e4be5)

16x16 resolution:
![image](https://github.com/minetest/minetest_game/assets/61879422/85c2db49-d76d-4551-9d94-fc7b30942785)

Loppansson:
As these has been merged, it seams appropriate to close it?:)

PR #3013, opened 2023-02-01
Improve opaque leaves textures

Wuzzy2:
Partial fix of #3001.

I updated all leaves textures so they look fine in Opaque Leaves mode. I edited the color of the transparent pixels but didn't change any other pixel. So the leaves in Fancy Leaves and Simple Leaves mode should **NOT** change at all. My technique was simply using the base texture and then rotating, copying and shifting it around until I filled the whole image and adding it to the background and then combine the whole thing. This works surprisingly well. :-)

~~There is one limitation: The blueberry bush with berries node still is bad. I swear I fixed the textures, but in Minetest, black pixels show up although both textures have no black pixels (even in the transparent ones). So my guess is Minetest has a problem with the texture overlay because this node uses an overlay. What should I do?~~

EDIT: Blueberry bush with berries fixed.

Screenshot: https://i.imgur.com/JYi3MTi.png

How to test: Test all leaves in Fancy, Simple and Opaque Leaves mode. Please also compare it with the previous texture and judge if it looks better than before.

**Warning: Be careful when optimizing the leaves textures in future, don't accidentally throw away the transparent pixel color.**


Wuzzy2:
OK, I managed to find a workaround for the blueberry bush with berries node: For the `default_blueberry_overlay.png`, I used colors from the new `default_blueberry_leaves.png` for the transparent pixels. It would be better if leaves don't use any overlay at all but I didn't want to introduce new textures in order to not break texture packs.

Wuzzy2:
Hmmm, I just noticed the bush leaves and apple leaves now look pretty similar in Opaque mode. Should I revert those back to the previous one?

appgurueu:
Thanks for this. It's definitely a massive improvement:

Screenshot

> Hmmm, I just noticed the bush leaves and apple leaves now look pretty similar in Opaque mode. Should I revert those back to the previous one?

I don't think that's an issue. I've always thought of bush and apple leaves as closely related (differing mostly in sparseness), similar to how acacia bush and acacia tree leaves are very similar (duh, they could even be the same node IMO).

Either way, this can be bikeshedded later on. No need to delay this improvement (or to go with uglier textures for better distinguishability). Why would you want to distinguish bush and apple leaves anways? Most of the time it is clear anyways from whether the leaves are, well, up on a tree or down on a bush.

sfan5 (approved):

PR #3020, opened 2023-03-10
Implement API to disable weather

### LINKED ISSUE #2913: Weather mod needs API to disable it
Quoting from https://github.com/minetest/minetest/issues/11915, talking about `weather` mod:
> I had no idea such mod existed. I looked at the code and it's just all about cycling clouds, so "weather" is a bit of a stretch

Since the introduction of `set_sky()` etc. functions, having a mod that only cycles clouds feels useless in my opinion. There are no atmospheric agents implemented and I doubt any mod depends from something like this. Right now is only something that messes with the `set_clouds()` function, and it's not like it'll be expanded anyway (as MTG is in mainteinance mode only)

t-affeldt:
I implemented an API for the weather mod. This resolves #2913 .

This is not a new feature per se, it is designed to enable compatibility between MTG and weather mods that add other effects to the sky or lighting. See the respective issue for the discussion.

This also reduces the amount of flashing due to clouds updating. When a new player joins, the sky will no longer be immediately recalculated for everyone but only for the new player.

The new API methods are exposed through a global `weather` object and are as follows:
```lua
-- returns bool for whether weather is active
weather.is_enabled()

-- enables / disables weather globally.
-- Has no effect if enable_weather == false in minetest.conf
-- If disabled, completely stops update cycle and reset every player to default values
weather.set_enabled(enable)

-- Most important part of this PR
-- Allows for overriding this function
-- Passes player object and tables that are about to be updated
-- Return values determine actual update to player object
-- Returning nil values for a table prevents it from getting updated
weather.on_update = function(player, { clouds, lighting })
```

Edit: Updated this post in line with redesign. Previous iteration suggested a `set_enable` function per-player instead of the `on_update`.

Tested on Minetest 5.6.1 and 5.7.0 RC1

### DIFF
--- a/game_api.txt
+++ b/game_api.txt
@@ -1147,3 +1147,36 @@ the log.
    that log corresponding actions
  * `def`     See [Node definition]
  * `name`    Description of the node in the log message
+
+
+Weather API
+-------
+
+The weather API allows you to enable / disable weather effects programmatically
+and also to modify weather effects before they are applied to the player.
+
+`weather.is_enabled()`
+
+* Returns true if weather mod is enabled or false if disabled
+* Will return false regardless of whether the mod has been disabled via the settings or via `set_enabled`.
+* Will return true even if selected worldgen doesn't support cloud updates. This is the case for v6 and singlenode types.
+
+`weather.set_enabled(enable)`
+
+* `enable` Boolean. Set to `true` to enable weather globally or `false` to disable it.
+* Setting this to `false` is only advised if you want to disable the weather effects __permanently__ and for every player.
+  Re-enabling effects later on will work but might conflict with other mods doing the same.
+  Override `weather.on_update` if you want to disable temporarily or need a higher level of control.
+* If the `enable_weather` setting is set to `false` in minetest.conf then this function will __not__ do nothing.
+
+`weather.on_update(player, overrides)`
+
+* Override this function to manipulate weather effects before they are applied to the player.
+* Calling this function on its own will do nothing. You have to override it in order to modify its behavior.
+* `player` The player entity that is about to be updated
+* `overrides` Defines which changes are about to be made. It is a table consisting of the following values:
+  * `overrides.clouds` A table (or `nil`) with cloud data following the same format as used for `player:set_clouds()`.
+  * `overrides.lighting` A table (or `nil`) with lighting data following the same format as used for `player:set_lighting()`.
+* This function is expected to return a table in the same format as `overrides`.
+  When overriding this function, the return value will be applied to the player instead of the original values.
+  Setting `clouds` or `lighting` to `nil` will __prevent__ that table from getting updated.
--- a/game_api.txt
+++ b/game_api.txt
@@ -1167,9 +1167,9 @@ and also to modify weather effects before they are applied to the player.
 * Setting this to `false` is only advised if you want to disable the weather effects __permanently__ and for every player.
   Re-enabling effects later on will work but might conflict with other mods doing the same.
   Override `weather.on_update` if you want to disable temporarily or need a higher level of control.
-* If the `enable_weather` setting is set to `false` in minetest.conf then this function will __not__ do nothing.
+* If the `enable_weather` setting is set to `false` in minetest.conf then this function will not do anything.
 
-`weather.on_update(player, overrides)`
+`weather.on_update = function(player, overrides)`
 
 * Override this function to manipulate weather effects before they are applied to the player.
 * Calling this function on its own will do nothing. You have to override it in order to modify its behavior.
--- a/mods/weather/init.lua
+++ b/mods/weather/init.lua
@@ -69,8 +69,8 @@ local os_time_0 = os.time()
 local t_offset = math.random(0, 300000)
 
 -- set a default shadow intensity for mgv6 and singlenode
-local function set_default_lighting(playerset)
-	for playername, _ in pairs(playerset) do
+local function set_default_lighting(players)
+	for playername, _ in pairs(players) do
 		local player = minetest.get_player_by_name(playername)
 		player:set_lighting({
 			shadows = { intensity = 0.33 }
@@ -78,10 +78,10 @@ local function set_default_lighting(playerset)
 	end
 end
 
-local function update_weather(playerset)
+local function update_weather(players)
 	-- skip cloud randomization if worldgen not supported
 	if not randomize_clouds then
-		set_default_lighting(playerset)
+		set_default_lighting(players)
 		return
 	end
 	-- Time in seconds.
@@ -98,7 +98,7 @@ local function update_weather(playerset)
 	local n_speedx = nobj_speedx:get_2d({x = time, y = 0}) -- -1 to 1
 	local n_speedz = nobj_speedz:get_2d({x = time, y = 0}) -- -1 to 1
 
-	for playername, _ in pairs(playerset) do
+	for playername, _ in pairs(players) do
 		local player = minetest.get_player_by_name(playername)
 		-- Fallback to mid-value 50 for very old worlds
 		local humid = minetest.get_humidity(player:get_pos()) or 50
--- a/mods/weather/init.lua
+++ b/mods/weather/init.lua
@@ -74,7 +74,7 @@ local function set_default_lighting(players)
 		local lighting = {
 			shadows = { intensity = 0.33 }
 		}
-		local overrides = weather.on_update(player, { lighting })
+		local overrides = weather.on_update(player, { lighting = lighting })
 		if overrides.lighting ~= nil then
 			player:set_lighting(overrides.lighting)
 		end
--- a/mods/weather/init.lua
+++ b/mods/weather/init.lua
@@ -4,7 +4,7 @@ local mod_enabled = minetest.settings:get_bool("enable_weather", true)
 local randomize_clouds = mod_enabled and mg_name ~= "v6" and mg_name ~= "singlenode"
 
 weather = {}
-local playerlist = {}
+local playerset = {}
 
 -- Parameters
 local TSCALE = 600 -- Time scale of noise variation in seconds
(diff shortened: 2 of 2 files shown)


mods/weather/init.lua
```diff
@@ -1,21 +1,12 @@
...
-if mg_name == "v6" or mg_name == "singlenode" then
-	-- set a default shadow intensity for mgv6 and singlenode
-	minetest.register_on_joinplayer(function(player)
-		player:set_lighting({ shadows = { intensity = 0.33 } })
-	end)
-
-	return
-end
+weather = {}
+local playerlist = {}
```
appgurueu, line 7: I'd rename this to `playerset` since it conceptually is a set of players who have weather enabled. `*list` definitely is misleading since this is not a list in the Lua sense.

mods/weather/init.lua
```diff
@@ -79,7 +68,7 @@ end
 local os_time_0 = os.time()
 local t_offset = math.random(0, 300000)
 
-local function update_clouds()
+local function update_clouds(players)
```
appgurueu, line 71: Why pass a list of players? ~~I'd just use the `playerset` upvalue.~~ Consider passing a set of players instead such that you may directly pass `playerset`. Note: This requires some further refactoring: You have to remove players from the set rather than setting them to `false`.

mods/weather/init.lua
```diff
@@ -94,7 +83,7 @@ local function update_clouds()
 	local n_speedx = nobj_speedx:get_2d({x = time, y = 0}) -- -1 to 1
 	local n_speedz = nobj_speedz:get_2d({x = time, y = 0}) -- -1 to 1
 
-	for _, player in ipairs(minetest.get_connected_players()) do
+	for _, player in ipairs(players) do
```
appgurueu, line 86: You'd have to change this to `for player in pairs(players) do` ...

mods/weather/init.lua
```diff
@@ -114,22 +103,93 @@ local function update_clouds()
...
+		player:set_lighting({
+			shadows = { intensity = 0.7 * (1 - density) }
+		})
 	end
 end
 
+local function purge_effects(player)
+	-- reset potentially touched values to their defaults
+	if randomize_clouds then
+		player:set_clouds({
```
appgurueu, line 115: Just pass no arguments, which resets the clouds to their default values.

mods/weather/init.lua
```diff
@@ -114,22 +103,93 @@ local function update_clouds()
...
+			density = 0.4,
+			thickness = 16,
+			speed = { x = 0, z = -2 }
+		})
+	end
+	player:set_lighting({
```
appgurueu, line 121: Since all fields are optional according to the docs, you *might* be able to reset this using just `{}`.

mods/weather/init.lua
```diff
@@ -114,22 +103,93 @@ local function update_clouds()
...
-	minetest.after(CYCLE, cyclic_update)
+-- Define API hooks
+
+-- override to set state for newly joined players
+weather.enable_on_join = true
+
+-- returns bool for whether weather is active
+-- returns nil if player is offline or weather is disabled
+weather.get_enabled = function(player)
+	return playerlist[player]
```
appgurueu, line 134: This API would change to returning `nil` (rather than `false`) if weather is disabled. It could be fixed using an "offline" check though.

mods/weather/init.lua
```diff
@@ -114,22 +103,93 @@ local function update_clouds()
...
 end
 
+-- override weather generation for individual player
+weather.set_enabled = function(player, enable)
+	if enable == playerlist[player] then
+		return
+	end
+	playerlist[player] = enable
```
appgurueu, line 142: You'd have to use `enable or nil` here to convert `false` to `nil`.

mods/weather/init.lua
```diff
@@ -114,22 +103,93 @@ local function update_clouds()
...
+		else
+			-- set a default shadow intensity for mgv6 and singlenode
+			player:set_lighting({
+				shadows = {
+					intensity = 0.33
+				}
+			})
+		end
+	else
+		playerlist[player] = false
```
appgurueu, line 189: You can just omit this if you convert to an actual set.

mods/weather/init.lua
```diff
@@ -114,22 +103,93 @@ local function update_clouds()
...
+-- override weather generation for individual player
+weather.set_enabled = function(player, enable)
+	if enable == playerlist[player] then
+		return
+	end
+	playerlist[player] = enable
+	if enable then
+		if randomize_clouds then
+			update_clouds({player})
+		end
```
appgurueu, line 146: Seems to be missing the `else` for a default shadow intensity for mgv6 and singlenode? If you were to add the `else`, this would be a code duplication, so I suggest making a new function `add_clouds(player)`.

I also now see the purpose of passing a list of players to `update_clouds`. You could consider passing a set instead (conceptually the more reasonable data structure since players may not appear multiple times). You'd have to change your calls for single players to `update_clouds({[player] = true})`. You could just pass the `playerset` inside the cyclic update.

mods/weather/init.lua
```diff
@@ -114,22 +103,93 @@ local function update_clouds()
...
+
+if update_clouds then
+	local function cyclic_update()
+		local players = {}
+		for player, state in pairs(playerlist) do
+			if state then
+				table.insert(players, player)
+			end
+		end
+		update_clouds(players)
```
appgurueu, line 167: This would change to `update_clouds(playerset)`; the for-loop above would be obsolete.

appgurueu (commented):
Should work in its current state (only concern is that if weather for a player is enabled later on, the default shadow intensity may not be set for mgv6 and singlenode). However, I have an overarching reliability concern: You consistently use player refs as indices. This works fine in practice since player refs *currently* don't change during the session; [in theory there is no such API guarantee though](https://github.com/minetest/minetest/issues/12271), which is why I like to err on the side of caution, using player names as keys instead.

mods/weather/init.lua
t-affeldt: My idea here was to only reset values actually touched by the mod instead of the whole table. So if another mod would have changed the cloud height then that change doesn't get overridden needlessly. The drawback of this implementation is of course that it needs to get updated if the default values ever change.

mods/weather/init.lua
t-affeldt: You would think so but that actually crashes the game. The lighting table is a bit inconsistent in that regard.

t-affeldt:
I have now updated the code to use a proper set and to store players by name. Shadows should now get correctly updated in v6 and singlenode.

mods/weather/init.lua
appgurueu: gah, why do the API docs never explicitly state whether absent values lead to overwriting with defaults or keeping the old values...

appgurueu (approved):
Seems fine to me. Luacheck will probably complain because you're shadowing `playerset` though (even though IMO this shadowing is valid)...

t-affeldt:
> Seems fine to me. Luacheck will probably complain because you're shadowing `playerset` though (even though IMO this shadowing is valid)...

Yes, I just got the email about the failed check. I thought shadowing is valid here because it's semantically the same. Just renamed it.

sfan5:
The need to have `weather.enable_on_join` in addition to the API functions is pretty inconvenient.
What about defining a public `weather.update = function(player)` that mods can override?
This gives:
* clean way for mods to disable weather, no need for synchronization on join
* multiple mods can use this mechanism without interference (mostly)

It does not provide ability to "reset" the weather changes but that's a lost cause anyway since values set by mods cannot stack.

appgurueu (changes requested):
I forgot to raise that you should document this in [`game_api.txt`](https://github.com/minetest/minetest_game/blob/master/game_api.txt) as sfan5 has pointed out on IRC.

t-affeldt:
> What about defining a public `weather.update = function(player)` that mods can override? This gives:

This would require a complete rewrite, so we should probably have this discussion _before_ I do that.

The current implementation offers, in my opinion, the easiest way to toggle weather effects per player. The `enable_on_join` is sadly necessary to prevent new players from ever getting any effects in the first place.

In an ideal scenario, mod authors would not only be able to toggle the weather but also adjust it slightly without a rewrite.
In order to do so, they would need to be handed the cloud and lighting tables before they get written to the player object. However, if disabling the mod is the only purpose then generating those tables would be unnecessary overhead.

So, an alternative suggestion would be the following:
```lua
weather.enabled = true
weather.on_update(player, { clouds, lighting })
```

The `weather.enabled` would simply toggle the weather globally.

The `weather.on_update` would be called after the tables have been generated and they would be passed as an argument. You could then override individual values and return the modified tables before they get written to the player object. Returning `nil` would instead prevent a table from getting written at all. I `weather.enabled == false` then this function would never be called.

t-affeldt:
I have redesigned the API  again. This iteration should be as flexible as possible.
Overriding the `on_update` method can be used to disable functionality, filter by players, add additional effects, apply modifiers, etc. with not much of a hassle.

The `set_enabled` function can be used to completely disable the update cycle, improving performance when weather is unused.

I have also added the respective explanations to the game_api.txt

appgurueu (approved):
Should be fine code-wise now. Note that

```lua
local a, b = 42, 101
local t = { a, b }
```

is not equivalent to

```js
let a = 42, b = 101;
let t = { a, b };
```

in JS this will create an object `{a: 42, b: 101}`, whereas in Lua it will simply create a "list" `{[1] = 42, [2] = 101}`. I've fixed this.

game_api.txt
```diff
@@ -1147,3 +1147,36 @@ the log.
...
+* Will return false regardless of whether the mod has been disabled via the settings or via `set_enabled`.
+* Will return true even if selected worldgen doesn't support cloud updates. This is the case for v6 and singlenode types.
+
+`weather.set_enabled(enable)`
+
+* `enable` Boolean. Set to `true` to enable weather globally or `false` to disable it.
+* Setting this to `false` is only advised if you want to disable the weather effects __permanently__ and for every player.
+  Re-enabling effects later on will work but might conflict with other mods doing the same.
+  Override `weather.on_update` if you want to disable temporarily or need a higher level of control.
+* If the `enable_weather` setting is set to `false` in minetest.conf then this function will not do anything.
```
sfan5, line 1170: I can't think of a way how more than one mod could correctly use `set_enabled` without stepping on the toes of the other mod.

An alternative would be to have just `is_enabled` and tell modders to override it.

mods/weather/init.lua
```diff
@@ -79,7 +68,25 @@ end
...
-local function update_clouds()
+-- set a default shadow intensity for mgv6 and singlenode
+local function set_default_lighting(players)
+	for _, player in ipairs(players) do
+		local lighting = {
+			shadows = { intensity = 0.33 }
+		}
+		local overrides = weather.on_update(player, { lighting = lighting })
+		if overrides.lighting ~= nil then
+			player:set_lighting(overrides.lighting)
```
sfan5, line 79: this will constantly reset the lighting parameters for everyone.
I suppose it's not bad because that's what this mod does anyway (on supported mapgens), but is this intended?
Asking because it didn't do so before.

PR #3021, opened 2023-03-13
Fix syntax errors in bones.sv.tr

Wuzzy2:
I've found technical syntax errors in the translation file `bones.sv.tr`. The `@2` symbol accidentally became an `@a` in the translations.

Disclaimer: I do not speak Swedish but since I am only touching symbols, this should be fine.


Wuzzy2:
FYI: I've found this with a script I wrote (warning, it's still rough around the edges!):

https://codeberg.org/Wuzzy/minetest_check_translations

appgurueu (approved):

PR #3029, opened 2023-04-29
Improved bucket texture

Extex101:
screenshot_20230429_104115

My goal was to maintain a sense of familiarity for players, so I kept it in roughly the same shape.

Would need to be squashed if merged.


Desour:
Imo the current texture fits the current style of mtg well. This change is not needed.

appgurueu:
I like the change. Yes, it is not strictly "needed", but I just merging it comes at a low cost considering the work has already been done. IMO MTG lacks a uniform texture style anyways so I don't think there is any style to harm; besides, the changes are not too drastic.

I'd wait for more opinions on it.

TheOnlyJoeEnderman:
I like it. I also made a texture too, but I don't know if people would like it as much.
bucket
image
image
image
image

sfan5:
Given the lack of discussion/interest happening and the fact that this is quite clearly not a maintenance change, I'll close this.

PR #3030, opened 2023-05-05
150 bones inventory slots and register_transfer_inventory_to_bones_on_player_death callback

imre84:
# in this PR there are two changes

## first, the size of the bones inventory were changed to 15*10

rationale of these changes:

there might be a scenario when a bones inventory would need to hold up to 4\*8+6+4\*3\*8+4+3\*3 distinct items:
* 4\*8 for the main inventory
* 6 for the 3d_armor
* (at most) 4\*3\*8 for 4 backpack worth of items (unified inventory)
* 4 more for the actual backpacks
* 3\*3 more for the crafting grid

that adds up to 147, so 150 slots would be sufficient

## secondly, I made a public interface in bones to provide a way for other mods to register callbacks to transfer items to bones on player death.

rationale of these changes:

3d armor for instance tries to search for a bones block in the vicinity of the death:
https://github.com/minetest-mods/3d_armor/blob/master/3d_armor/init.lua#L371
given all the precautions there might be an ambiguity (the same player dying on the same spot in the past 20 minutes)
also a scenario is possible when the main inventory and the crafting grid of the player was empty, in this case 3d armor will just drop its items disregarding the bones_mode configuration

also, please consider unified_inventory where backpack related items are stored in detached inventories, making the pre-existing api unable to handle these inventories


appgurueu:
As I already wrote on IRC, I don't think a fixed "upper bound" of 150 slots is a decent solution.

First of all, a few more items and 150 slots are too few again; this needs to be variable.

Second, most of the time most of these slots will be unused, especially in vanilla MTG, worsening UX.
I think the size of the bones should depend on how many items they contain, and 4*8 should be a lower bound to not break hacks such as 3d_armor's.

Third, all items making their way to the bones might break game mechanics: Currently, 3d_armor will just drop the armor if it doesn't fit in the bones (correct me if I'm wrong). With this change, they'd always be in the bones, thus there'd be no point in quick retrieval of them, and no point in keeping slots free in your inventory; really, if you think of it, the size limitation of the bones may be considered a gameplay feature. Not sure what backpacks etc. do. 

I definitely support the addition of an API. That way, mods which want everything to always go in the bones should be able to make it happen. Here's my proposal: Allow adding items to the bones via a callback which is called on player death and returns the items to be added. If the items exceed 4*8, expand the inventory accordingly. Otherwise, leave it untouched.

mods/bones/init.lua
```diff
@@ -190,15 +204,68 @@ local function is_all_empty(player_inv)
 	return true
 end
 
+--functions registered this way won't becalled if bones_mode is keep
+function bones.public.register_transfer_inventory_to_bones_on_player_death(func)
+	bones.private.dead_player_callbacks[#(bones.private.dead_player_callbacks)]=func
```
appgurueu, line 209: Use `table.insert` instead. This is completely broken in its current form: It just sets `dead_player_callbacks[0] = func`.

game_api.txt
```diff
@@ -83,6 +83,16 @@ in `bones.player_inventory_lists`.
 
 e.g. `table.insert(bones.player_inventory_lists, "backpack")`
 
+Additionally, callbacks can be registered to transfer items into the bones on death:
+
+`bones.public.register_transfer_inventory_to_bones_on_player_death(function(player){})`
```
appgurueu, line 88: We don't do that here. Just `bones.name`. And please give it a shorter name.

game_api.txt
```diff
@@ -83,6 +83,16 @@ in `bones.player_inventory_lists`.
...
+
+In the above functions this routine should be used to add itmes to the bones inventory:
+
+`bones.public.transfer_stack_to_bones(stack)`
```
appgurueu, line 92: Same here. I wonder whether the slightly "functional" returning a list to be added might be cleaner as opposed to this "imperative" style, especially considering the ugly global context in which this function operates: What happens if this is called when no bones are currently being built?

Get rid of this function in its current form, either by providing the callback with a closure parameter (`function(name, add_item_to_bones)`) such that it can then call its local `add_item_to_bones` to well, add items to bones. Or return a list.

mods/bones/init.lua
```diff
@@ -5,8 +5,22 @@
 
 -- Load support for MT game translation.
 local S = minetest.get_translator("bones")
-
-bones = {}
+-- bones are supposed to hold up to 4*8+6+4*3*8+4+3*3 item slots:
```
appgurueu, line 8: This is MTG. No external-mod-specific calculations here. As said, make this variable.

mods/bones/init.lua
```diff
@@ -5,8 +5,22 @@
...
+-- (at most) 4*3*8 for 4 backpack worth of items (unified inventory)
+-- 4 more for the actual backpacks
+-- 3*3 more for the crafting grid
+-- that adds up to 147, so 150 slots would be sufficient
+local cols=15
+local rows=10
+
+bones = {
+	private={
+		dead_player_callbacks={}
```
appgurueu, line 20: Just use a local variable. This will also make the code considerably shorter and slightly faster.

mods/bones/init.lua
```diff
@@ -5,8 +5,22 @@
...
+	},
+	public={}
```
appgurueu, line 22: As said, eliminate this. Use `local` variables for private stuff and global variables (under the `bones` table) for public stuff.

appgurueu (changes requested):
* You don't need parenthesis around conditions.
* You don't need the entire private/public stuff. This isn't Java.
* Get rid of the global variables applying to the current bones. They should instead be local variables (local to the function placing the bones, not local to the file).

mods/bones/init.lua
```diff
@@ -211,67 +278,43 @@ minetest.register_on_dieplayer(function(player)
...
-			". Inventory dropped")
-		if bones_position_message then
-			minetest.chat_send_player(player_name, S("@1 died at @2, and dropped their inventory.", player_name, pos_string))
+	else
+		if(not(bones.private.current_dead_player.dropped))then
+			bones_conclusion="Bones placed"
+			public_conclusion="bones were placed"
+		else
+			bones_conclusion="Inventory partially dropped"
+			public_conclusion="partially dropped their inventory"
```
sfan5, line 303: the text won't be properly translatable like this.

game_api.txt
imre84: > And please give it a shorter name.
I can't come up with a short but descriptive name
would bones.register_dead_player_inv_management be short enough?

game_api.txt
imre84: > What happens if this is called when no bones are currently being built?
it would crash the whole server due to indexing a nil. That was one of the reasons it was done this way: I wanted to discourage everyone from using this routine outside of this context (however as per other requests the context will now be stored in a function local variable and passed as function argument)

mods/bones/init.lua
```diff
@@ -211,81 +275,73 @@ minetest.register_on_dieplayer(function(player)
...
+		drop(current_dead_player.player_pos, ItemStack("bones:bones"))
+		if not ( current_dead_player.dropped ) then
+			bones_conclusion="No bones placed"
+			if bones_position_message then
+				minetest.chat_send_player(player_name, S("@1 died at @2.", player_name, pos_string))
+			end
+		else
+			bones_conclusion="Inventory dropped"
+			if bones_position_message then
+				minetest.chat_send_player(player_name, S("@1 died at @2, and dropped their inventory.", player_name, pos_string, public_conclusion))
```
appgurueu, line 299: This won't work. Three arguments but only two `@` params in the translation string.

mods/bones/init.lua
```diff
@@ -211,81 +275,73 @@ minetest.register_on_dieplayer(function(player)
...
-		drop(pos, ItemStack("bones:bones"))
-		minetest.log("action", player_name .. " dies at " .. pos_string ..
-			". Inventory dropped")
-		if bones_position_message then
-			minetest.chat_send_player(player_name, S("@1 died at @2, and dropped their inventory.", player_name, pos_string))
+	else
+		if not ( current_dead_player.dropped ) then
+			bones_conclusion="Bones placed"
+			if bones_position_message then
+				minetest.chat_send_player(player_name, S("@1 died at @2, and bones were placed.", player_name, pos_string, public_conclusion))
```
appgurueu, line 306: Same here.

mods/bones/init.lua
```diff
@@ -211,81 +275,73 @@ minetest.register_on_dieplayer(function(player)
...
+			end
+		else
+			bones_conclusion="Inventory partially dropped"
+			if bones_position_message then
+				minetest.chat_send_player(player_name, S("@1 died at @2, and partially dropped their inventory.", player_name, pos_string, public_conclusion))
```
appgurueu, line 311: And here.

mods/bones/init.lua
```diff
@@ -211,81 +275,73 @@ minetest.register_on_dieplayer(function(player)
...
 
-	-- check if it's possible to place bones, if not find space near player
-	if bones_mode == "bones" and not may_replace(pos, player) then
-		local air = minetest.find_node_near(pos, 1, {"air"})
-		if air and not minetest.is_protected(air, player_name) then
-			pos = air
-		else
-			bones_mode = "drop"
-		end
+	for i=1,#dead_player_callbacks do
```
appgurueu, line 282: Use `ipairs` here.

mods/bones/init.lua
```diff
@@ -181,24 +199,70 @@ end
...
+bones.register_dead_player_inv_management(player_dies_transfer_inventory)
+
 minetest.register_on_dieplayer(function(player)
+	local pos = vector.round(player:get_pos())
 	local bones_mode = minetest.settings:get("bones_mode") or "bones"
 	if bones_mode ~= "bones" and bones_mode ~= "drop" and bones_mode ~= "keep" then
 		bones_mode = "bones"
 	end
+	local player_name = player:get_player_name()
+	local current_dead_player={player=player, player_name=player_name, bones_inv=nil, bones_pos=nil,
```
appgurueu, line 262: Stuffing all variables into this table isn't clean at all. You could easily get rid of this by just inlining all the code here as it was before, and inserting the appropriate callback calls.

If you want to split this into functions, do it properly, with well-defined functions and proper arguments rather than opaque tables.

appgurueu:
* Fix the Luacheck warnings. This includes eliminating global variables.
* Your code style still doesn't match the rest of the code. You use excessive spacing around parenthesis and parenthesis where none are needed.
* My point about a variable size was not that there should be a configurable max, but rather that after the callbacks have been run, the size should be determined based on how many items were added; the minimum size should be 4*8 to support 3d_armor's hack.
* What happened to the logging?
* Your usage of translations is wrong (see review comments).

mods/bones/init.lua
imre84: >     * What happened to the logging?

https://github.com/imre84/minetest_game/blob/master/mods/bones/init.lua#L316
here's the logging. The code was restructured but I cannot remember any log entries being removed by me.

mods/bones/init.lua
imre84: > * the size should be determined based on how many items were added; the minimum size should be 4*8 to support 3d_armor's hack.

so do you want the armor to be dropped if there are no free slots in the main inventory? you can set the bones max slots to 64 to do so. or maybe I don't understand what do you mean.

imre84:
>     * My point about a variable size was not that there should be a configurable max, but rather that after the callbacks have been run, the size should be determined based on how many items were added; the minimum size should be 4*8 to support 3d_armor's hack.

also, please take a look at https://github.com/imre84/minetest_game/blob/master/mods/bones/init.lua#L33
the displayed size of the grid depends on the actual number of items inside, and has nothing to do with the configurable max (other than serving as an upper bound)

imre84:
> * Fix the Luacheck warnings. This includes eliminating global variables.

can I run luachecks somehow?

appgurueu (commented):
Some variable names and some minor formatting could use improvement.

Can you try to clean up the code and in particular the highly imperative loop where you lazily add bones if there is an item to be placed a bit?

mods/bones/init.lua
```diff
@@ -211,81 +243,99 @@ minetest.register_on_dieplayer(function(player)
...
-			end
-			player_inv:set_list(list_name, {})
+	local bones_conclusion
+	local public_conclusion
+
+	if not bones_pos then
+		drop(player_pos, ItemStack("bones:bones"))
+		if not dropped then
+			bones_conclusion = "No bones placed"
+			public_conclusion = S("@1 died at @2.", player_name, pos_string)
```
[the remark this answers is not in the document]
imre84: > The `S(x, player_name, pos_string)` seems to be common to all four cases. There's no need to duplicate it. I'd eliminate it and move it to after the cases. Why did you call this `public_conclusion`?

thanks for supporting me btw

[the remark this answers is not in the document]
appgurueu: How about `log_message` and `chat_message`?

mods/bones/init.lua
```diff
@@ -211,81 +243,99 @@ minetest.register_on_dieplayer(function(player)
...
+		for i = 1, theoretical_max_slots do
+			local stack = bones_inv:get_stack("main", i)
+			if stack:get_count() == 0 then
+				inv_size = i - 1
+				break
 			end
 		end
-		player_inv:set_list(list_name, {})
-	end
+		if inv_size <= 4 * 8 then
```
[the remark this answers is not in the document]
imre84: given the fact that 4*8 was used all over the places I added
```lua
-- we're going to display no less than 4*8 slots, we'll also provide at least 4*8 slots in bones
local min_inv_size = 4 * 8
```
locally to the file

mods/bones/init.lua
```diff
@@ -6,6 +6,9 @@
 -- Load support for MT game translation.
 local S = minetest.get_translator("bones")
 
+local theoretical_max_slots = minetest.settings:get("bones_max_slots") or 15 * 10
```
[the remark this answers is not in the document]
imre84: renamed to `bones_max_slots`

imre84:
> where you lazily add bones

the original functionality also drops a bone if the bone block cannot be placed, I wanted to retain this functionality. Is there a better way to do this?

mods/bones/init.lua
```diff
@@ -211,81 +245,100 @@ minetest.register_on_dieplayer(function(player)
...
-	local player_inv = player:get_inventory()
-	if is_all_empty(player_inv) then
-		minetest.log("action", player_name .. " dies at " .. pos_string ..
-			". No bones placed")
-		if bones_position_message then
-			minetest.chat_send_player(player_name, S("@1 died at @2.", player_name, pos_string))
-		end
-		return
-	end
+	for _, fun in ipairs(dead_player_callbacks) do
```
appgurueu, line 248: I feel it should be possible to refactor this loop by splitting it up into functions:

* A function to collect the items to be added to the inventory in a list. This function would run the callbacks and append all the items to a list.
* A function to find a suitable position for a bones node.
* A function to set and populate the inventory of set node.

Perhaps the latter two functions should be only one. You might also want to inline these, but the important thing is that I'd like to have this as a clear-cut three stage process:

1. Collect items to be placed in the inventory
2. Find a position to place the block
3. Place the block and populate the inventory

in order to get rid of the very stateful loop with high cyclomatic complexity

appgurueu (approved):
Sorry for the late review. The logic checks out, but is harder to follow than it needs to be. Consider refactoring as suggested in my comment.

There are also some minor typos here and there that I'll fix in one go before merge when this gets a second approval.

appgurueu (approved):
Refactored.

fluxionary:
not sure if this is super relevant, but someone suggested that i link my bones fork here: 

https://content.minetest.net/packages/rheo/bones/

it's got an API for adding additional inventories, and comes w/ support for 3d_armor, unified_inventory, and i3. it also has a lot of other additional features, and due to #2710, i didn't think it was appropriate to try to get them integrated here.

appgurueu:
It's difficult to draw the line between features and maintenance. I see dehardcoding and adding APIs as maintenance rather than "adding features"; it is necessary in this case such that mods like 3d_armor don't have to rely on unreliable hacks anymore. Effectively there will be bugs (or at least very hacky code) in 3d armor and many other mods which add player inventories which are induced by us not offering a proper API.

Note however that these dependencies have go the other way around: MTG will not make itself dependant on 3d armor, i3, unified inventory, etc. These mods (and any mods that may come in the future) have to use the game APIs we offer to interact with MTG.

imre84:
hello, have anything happened to this PR? is it incorporated into mtg5.8.0?

fluxionary:
> hello, have anything happened to this PR? is it incorporated into mtg5.8.0?

if the core devs don't respond, feel free to PR any fixes to https://github.com/fluxionary/minetest-bones_redo. it's already in use on a couple popular servers, and i'm happy to maintain it.

SmallJoker:
Rebase needed. I support this minor feature and will review it. The "maintenance-only" state that MTG currently has should not be the reason for requiring mods to provide helpful functionality of MTG mods.

imre84:
can somebody kindly please point me towards an RTFM regarding how to solve this conflict? I cannot think of an easy way for git mergetool to work given the fact that the conflict is between repositories and not inside a single repository

imre84:
> can somebody kindly please point me towards an RTFM regarding how to solve this conflict? I cannot think of an easy way for git mergetool to work given the fact that the conflict is between repositories and not inside a single repository

For my future reference:

```
git remote add upstream <original-repository-URL>
git fetch upstream
git checkout your-branch-name
git merge upstream/master
git add .
git commit -m "Resolved conflicts and merged upstream/master"
git push origin your-branch-name
```

SmallJoker:
I tried to rebase but force-pushing to master was incomplete. Sorry for the fuss. Here's a rebased version: https://pastebin.com/raw/kx9uGyew

imre84:
I merged in the meantime, as far as I can tell the only conflict was safe to be resolved by using my version of the code

imre84:
> In the future I would recommend you to not use the `master` branch for PRs.

Yes, I came to the same conclusion

correct me if I'm wrong but does all this mean that this PR will find its way to the next release of MTG? am I required to do anything else?

SmallJoker:
> correct me if I'm wrong but does all this mean that this PR will find its way to the next release of MTG?

This will only happen when this PR is approved >= 2 times and then merged into MTG. For reference: https://dev.minetest.net/Git_Guidelines#Upstream_commit_rules

mods/bones/init.lua
```diff
@@ -182,41 +196,107 @@ local drop = function(pos, itemstack)
...
+-- Try to find the closest space near the player to place bones
+local function find_bones_pos(player)
+	local rounded_player_pos = vector.round(player:get_pos())
+	local bones_pos
+	if may_replace(rounded_player_pos, player) then
+		bones_pos = rounded_player_pos
+	else
+		bones_pos = minetest.find_node_near(rounded_player_pos, 1, {"air"})
+	end
+	if minetest.is_protected(bones_pos, player:get_player_name()) then
```
SmallJoker, line 234: This indirectly reverts 110c23560 (the conflicting commit).

When air was found, there should be no protection check.

imre84: > This indirectly reverts [110c235](https://github.com/minetest/minetest_game/commit/110c23560edf5ddc7ed3ffd0d0c288a96e0343e8) (the conflicting commit).
> 
> When air was found, there should be no protection check.

may_replace also check for protections, that might mean duplicate code. I think the if in 234 is safe to be deleted altogether then

mods/bones/init.lua
```diff
@@ -221,71 +295,52 @@ minetest.register_on_dieplayer(function(player)
...
-			else -- no space left
-				drop(pos, stack)
-			end
+	if bones_placed then
+		if drop_bones then
+			log_message = "Inventory partially dropped"
+			chat_message = "@1 died at @2, and partially dropped their inventory."
+		else
+			log_message = "Bones placed"
+			chat_message = "@1 died at @2, and bones were placed."
```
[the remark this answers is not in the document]
imre84: log_messages are not to be translated, they are going into the log (which is never translated afaik)
S is already applied to chat_message at line 352, I'm working on the rest

mods/bones/init.lua
```diff
@@ -6,8 +6,15 @@
 -- Load support for MT game translation.
 local S = minetest.get_translator("bones")
 
+local bones_max_slots = minetest.settings:get("bones_max_slots") or 15 * 10
+local min_inv_size = 4 * 8 -- display and provide at least this many slots
+
 bones = {}
 
+local function TRANSLATEMEPLEASE(s)
```
appgurueu, line 14: This no-op function needs to be named `NS`. See the [translation updater script](https://github.com/minetest/minetest/blob/fbec378869691444e354425073a7210555a9b31d/util/mod_translation_updater.py#L125).

SmallJoker (commented):
(wrong button was focused ..?)

Dying with the creative privilege:
✔ No bones were placed
✔ Log: "No bones placed"
✘ Inventory is cleared (it should not)

Dying without creative (non-empty inventory):
✔ Bones were placed
✔ Log: "Bones placed"
✔ Inventory is cleared


Dying without creative (empty inventory):
✘ Bones were placed
✘ Log: "Bones placed"
✔ Inventory is cleared (it already was empty)

imre84:
> (wrong button was focused ..?)

Hello, I don't understand this sentence, if there's nothing I need to do about it, then nevermind.

SmallJoker:
@imre84 That's not relevant. What needs fixing are the test cases that are marked with ✘.

imre84:
> @imre84 That's not relevant. What needs fixing are the test cases that are marked with ✘.

ok, fixed, will rebase soon

imre84:
let me know if something's off I think that the rebase was successful

SmallJoker (commented):
Dying with the creative privilege:
✔ No bones are placed
✔ Log: "No bones placed"
✔ Inventory is kept

Dying without creative (non-empty inventory):
✔ Bones are placed
✔ Log: "Bones placed"
✔ Inventory is cleared

Dying without creative (empty inventory):
✔ No bones are placed
✔ Log: "Bones no placed"
✔ Inventory is cleared (it already was empty)

imre84:
all done, formspec vs scrollbars was a bit fidgety, I hope you like it

mods/bones/init.lua
```diff
@@ -6,8 +6,15 @@
 -- Load support for MT game translation.
 local S = minetest.get_translator("bones")
 
+local bones_max_slots = minetest.settings:get("bones_max_slots") or 15 * 10
```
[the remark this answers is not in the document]
imre84: fixed in e27ad580

SmallJoker:
@imre84 Thank you for the scroll container. This should scale much better on mobile platforms with limited screen space. Does the scroll container fit well on your side? Asking because I get the following situation when inserting about 9 additional stacks:

scroll Y max

imre84:
Képernyőkép ekkor: 2024-03-30 18-14-03
this is how it looks like for me

mods/bones/init.lua
```diff
@@ -23,24 +23,38 @@ local function is_owner(pos, name)
...
-		cols, rows = math.ceil(numitems / 4), 4
 	else
-		cols, rows = 15, math.ceil(numitems / 15)
+		cols, rows = 8, math.ceil(numitems / 8)
+		scroll=true
+	end
+	local output={}
+	appendmulti(output, "size[", 8.5, ",", 9, "]")
+	if scroll  then
+		appendmulti(output, "scrollbaroptions[max=",rows*9.3,"]")
```
imre84, line 44: the problem with the scrollbar might be lying here, maybe the proper formula would be something like (rows-delta)*multiplier

mods/bones/init.lua
```diff
@@ -235,6 +235,10 @@ bones.register_collect_items(function(player)
 
 		player_inv:set_list(list_name, {})
 	end
+	-- debug code, needs to be reverted:
```
imre84, line 238: needs to be reverted but maybe makes testing easier

imre84:
is it any better now?

imre84:
also current behaviour is to display the first 4*8 slots even if they're empty, which is a nightmare from UX perspective, some form of slot reordering would be beneficial

imre84:
ok, I'm not sure if bones inventory reordering is over the top, maybe it should be nerfed somehow, idk

mods/bones/init.lua
```diff
@@ -6,8 +6,15 @@
 -- Load support for MT game translation.
 local S = minetest.get_translator("bones")
 
+local bones_max_slots = tonumber(minetest.settings:get("bones_max_slots")) or 15 * 10
```
SmallJoker, line 9: Needs documentation in `minetest.conf.example`.

SmallJoker (commented):
The scrollbar does fit much better now. Perhaps the hacky scrollbar size calculation can be fixed in the future directly within Minetest. We'll see.
Your calculation does work well on my end. Thanks.

I also tested the partially dropped inventory case, which also works as expected.
See comments for the last remaining issues.

mods/bones/init.lua
SmallJoker: A `max=` value directly expressed in inventory slots would be even better. I added it to my long term notes to have a look at in the Minetest Engine.

imre84: well, either  inventory slots or inventory rows but it would be easier to use, I agree

mods/bones/init.lua
```diff
@@ -16,18 +23,86 @@ local function is_owner(pos, name)
...
-	default.get_hotbar_bg(0,4.85)
+local function appendmulti(tbl,...)
+	for _, v in pairs({...}) do
+		table.insert(tbl, v)
+	end
+end
+
+local function get_bones_formspec_for_size(numitems)
+	local cols, rows
+	local scroll=false
```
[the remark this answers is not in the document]
imre84: >     * too long lines

suggestion: too long lines can be checked by luacheck

mods/bones/init.lua
imre84: > Needs documentation in `minetest.conf.example`.

you most probably mean the minetest.conf.example inside minetest_game (not in minetest)
however there are two minetest.conf.example files, and they differ

imre84: fixed in 357f87a

imre84:
ok, hopefully all done, let me know if I missed something

SmallJoker (approved):
Works.

SmallJoker:
@imre84 Rebase needed.

imre84:
hello, I haven't played luanti in a while, I don't have the patience for this right now. Hopefully I reverted to the state it was yesterday. sorry about that. hopefully someone else picks up from here.

PR #3032, opened 2023-05-19
fix logic error in bed rotation

fluxionary:
the affected code seems intended to prevent the rotation of a bed if the "top" of the bed has been replaced or rotated. 

this was discovered by a simple script that was checking for this type of logic error.


appgurueu (approved):
I wonder why Luacheck didn't catch this. I vaguely remember Luacheck catching this type of mistake.

fluxionary:
> I wonder why Luacheck didn't catch this. I vaguely remember Luacheck catching this type of mistake.

luacheck (at least, v 1.1.0, the 0.26 version is bugged) *does* catch `if not a == b then`, but it seems to get confused by `if not a or not a == b then`. possibly i should bug that upstream.

PR #3040, opened 2023-06-11
Bone logging

imre84:


mods/default/functions.lua
```diff
@@ -754,12 +754,12 @@ function default.set_inventory_action_loggers(def, name)
...
 	local on_put = def.on_metadata_inventory_put or nop
 	def.on_metadata_inventory_put = function(pos, listname, index, stack, player)
-		default.log_player_action(player, "moves", stack:get_name(), "to", name, "at", pos)
+		default.log_player_action(player, "moves", stack:to_string(), "to", name, "at", pos)
 		return on_put(pos, listname, index, stack, player)
 	end
 	local on_take = def.on_metadata_inventory_take or nop
 	def.on_metadata_inventory_take = function(pos, listname, index, stack, player)
-		default.log_player_action(player, "takes", stack:get_name(), "from", name, "at", pos)
+		default.log_player_action(player, "takes", stack:to_string(), "from", name, "at", pos)
```
sfan5, line 762: I believe this was intentional as to not print potentially huge amounts of metadata to the log

imre84: what I care about is name and amount. but I can see what you're talking about

appgurueu: Then why not format just name and amount?

imre84: > Then why not format just name and amount?

that would mean storing default:cobble 1 to the chest instead of storing default:cobble to the chest. Would that be okay?

imre84: do you mean renaming name_and_amount to name_and_count or deleting that function completely?

appgurueu (approved):
Looks fine to me.

sfan5 (approved):
lgtm

PR #3045, opened 2023-08-02
don't remove node above door if it's not a doors:hidden node

fluxionary:
currently, doors will remove the node above the door if the door is destroyed in various ways. however, there's times when the node above the door may *not* actually be a "doors:hidden" node, because of other mods. in particular, a mesecons sticky piston can be used to destroy an arbitrary node, whether or not it is private (can only be destroyed by a certain player) or protected (can only be modified if `minetest.is_protected` returns false). 

this PR checks that the node above the door is actually a "doors:hidden" node before destroying it.

### DIFF
--- a/mods/doors/init.lua
+++ b/mods/doors/init.lua
@@ -19,6 +19,10 @@ local function replace_old_owner_information(pos)
 	end
 end
 
+local function is_doors_upper_node(pos)
+	return minetest.get_node(pos).name == "doors:hidden"
+end
+
 -- returns an object to a door object or nil
 function doors.get(pos)
 	local node_name = minetest.get_node(pos).name
@@ -388,7 +392,9 @@ function doors.register(name, def)
 		end
 	end
 	def.after_dig_node = function(pos, node, meta, digger)
-		minetest.remove_node({x = pos.x, y = pos.y + 1, z = pos.z})
+		if is_doors_upper_node(pos) then
+			minetest.remove_node({x = pos.x, y = pos.y + 1, z = pos.z})
+		end
 		minetest.check_for_falling({x = pos.x, y = pos.y + 1, z = pos.z})
 	end
 	def.on_rotate = function(pos, node, user, mode, new_param2)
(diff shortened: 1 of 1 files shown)


mods/doors/init.lua
```diff
@@ -426,15 +432,21 @@ function doors.register(name, def)
 		def.node_dig_prediction = ""
 	else
 		def.on_blast = function(pos, intensity)
-			minetest.remove_node(pos)
+			if is_doors_upper_node(pos) then
+				minetest.remove_node(pos)
+			end
 			-- hidden node doesn't get blasted away.
-			minetest.remove_node({x = pos.x, y = pos.y + 1, z = pos.z})
+			if is_doors_upper_node(pos) then
```
sfan5, line 439: this is also wrong, you're checking pos but removing one above that

mods/doors/init.lua
```diff
@@ -654,9 +655,7 @@ function doors.register_trapdoor(name, def)
 		def.node_dig_prediction = ""
 	else
 		def.on_blast = function(pos, intensity)
-			if is_doors_upper_node(pos) then
-				minetest.remove_node(pos)
-			end
+			minetest.remove_node(pos)
 			return {name}
```
appgurueu: Should both of these `return {name}`? I believe the `return {name}` here should be removed, otherwise we get varying drops based on a race condition:

* If the top is blasted first, we get 2 doors.
* If the bottom is blasted first, the top is removed, and we only get one door.

Also: Am I missing something, or could you duplicate doors by blowing them up with TNT (before this patch)?

And in general, concerning the logic: Shouldn't both nodes remove each other to avoid inconsistent states? A "bottom without hidden top" shouldn't happen either, should it?

fluxionary: i'm deliberately not looking into whether `return {name}` is correct right now, or what the deal is w/ tnt. i'll try to look at it soon though. perhaps that should be a separate issue. 

the door and "hidden" node can currently be separated by e.g. a mesecons piston. i really doubt anyone wants the doors mod to depend (optionally) on mesecons, and other mods could introduce that same kind of mechanic in the future, so that wouldn't even solve the problem in general. i suppose mesecons could automatically register all doors as MVPS stoppers? e.g. iterate `doors.registered_doors` and also override `doors.register`.

appgurueu (approved):
Definitely an improvement over removing an arbitrary node. Further improvements can be made later on.

The `local above = ...; if is...(above) then remove_node(...) end` is repeated three times. Perhaps rather than having `is_door_upper_node`, extract this into a function `try_remove_upper_node`?

sfan5 (approved):

PR #3046, opened 2023-08-11
close chest: don't replace a node if it's not the expected open chest

fluxionary:
this is similar to #3045. currently, closing a chest dialogue will replace *anything* currently at the location of the chest when the dialogue was opened, even if it's another node or air. currently this can be exploited very reliably w/ mesecons sticky pistons, but other players have reported that it can be exploited unreliably w/ tnt (even though open chests should be tnt-proof?). but in general, there's no way to guarantee that the node hasn't changed since we opened it.


appgurueu (approved):

appgurueu:
If you know even more such issues, please report them :smile:

PR #3049, opened 2023-08-18
New Plank Textures

TheOnlyJoeEnderman:
This Pull Request makes the plank textures prettier and smoother. This is the single biggest thing that I wanted changed, and I was able to do so. These are clean room, CC0 takes on a wood plank design. They do have a similar appearance to Minecraft on purpose but as I mentioned before, are strictly clean room designs.

Comparison between my Aspen planks, vanilla, and Minecraft Birch for those curious.
wood_comparison


appgurueu:
I feel this breaks with the bland Minetest Game aesthetic in favor of a more Minecraft-like aesthetic. It is probably better left to a texture pack or Minecraft clone game projects such as Mineclone. I'll leave this open for people to share their opinions.

SmallJoker:
I agree that the current texture is not so good because they are simply recycled from the other plank textures. They could benefit from some darker patches, noise and cuts to look more like planks. Moretrees does this for example (cedar, sequoia from right to left):

planks comparison


At least the colours are currently okay-ish and going into a good direction with your proposal. The reference image I used for this comment:

aspen planks

TheOnlyJoeEnderman:
I could make world aligned versions that have longer boards. Though I feel that is sort of cheating as far as texture creation.

TheOnlyJoeEnderman:
It won't tile like I would like, but I can make the world aligned version. In that at certain world coordinates the texture will look wrong where two sides meet. Which is why I prefer Minecraft's method of handling randomly textured blocks, but this method may not look all that bad.

TheOnlyJoeEnderman:
default_woods
here is a better picture of them as is.

TheOnlyJoeEnderman:
default_woods_big

Bituvo:
They look a bit too shiny/smooth for me.

GNUAn:
@TheOnlyJoeEnderman I think you should send that pull request to Minerev(https://codeberg.org/GNUAn/Minerev)

sfan5:
Given the lack of discussion/interest happening and the fact that this is quite clearly not a maintenance change, I'll close this.

PR #3060, opened 2023-09-03
Stairs: Add helper function for textures

tenplus1:
Use function to tidy stair textures

### DIFF
--- a/mods/stairs/init.lua
+++ b/mods/stairs/init.lua
@@ -178,7 +178,7 @@ end
 
 function stairs.register_slab(subname, recipeitem, groups, images, description,
 		sounds, worldaligntex)
-	local light, alpha, sunlight, mode_sounds = get_node_vars(recipeitem)
+	local light, alpha, sunlight, node_sounds = get_node_vars(recipeitem)
 	local slab_images = set_textures(images, worldaligntex)
 	local new_groups = table.copy(groups)
 	new_groups.slab = 1
--- a/mods/stairs/init.lua
+++ b/mods/stairs/init.lua
@@ -63,18 +63,6 @@ local function warn_if_exists(nodename)
 	end
 end
 
--- Get node settings to use for stairs
-local function get_node_vars(nodename)
-
-	local def = minetest.registered_nodes[nodename]
-
-	if def then
-		return def.light_source, def.use_texture_alpha, def.sunlight_propagates, def.sounds
-	end
-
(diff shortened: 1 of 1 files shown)


mods/stairs/init.lua
```diff
@@ -103,6 +94,16 @@ function stairs.register_stair(subname, recipeitem, groups, images, description,
 			end
 		end
```
sfan5, line 95: you can even optimize the logic a bit more:
```lua
		stair_images[i] = type(image) == "string" and {name = image} or table.copy(image)
		if stair_images[i].backface_culling == nil then
			stair_images[i].backface_culling = true
		end
		if worldaligntex and stair_images[i].align_style == nil then
			stair_images[i].align_style = "world"
		end
	end
```

tenplus1:
@sfan5 - Thanks, amended pull.  Also, what do you think about stairs automatically using node sounds ?

appgurueu (approved):

sfan5:
> Also, what do you think about stairs automatically using node sounds ?

sounds useful

mods/stairs/init.lua
```diff
@@ -69,10 +69,10 @@ local function get_node_vars(nodename)
 	local def = minetest.registered_nodes[nodename]
```
appgurueu, line 69: If I'm seeing this correctly, the only thing this function is ever used for is to "transplant" a few definition fields? Why not refactor it to be a `local function transplant_def_fields(nodename, to_def)`?

The code would then become something like

```lua
local def = minetest.registered_nodes[nodename]
if not def then return end
for _, field in ipairs({"light_source", "use_texture_alpha", "sunlight_propagates", "sounds"}) do
    if to_def[field] == nil then
        to_def[field] = def[field]
    end
end
```

Or alternatively, something like `register_node_based_on(nodename, base_nodename, def)`?

appgurueu (approved):
Looks fine as the "minimal change" for the sounds, but ideally `get_node_vars` with its 4 return values - which fulfill pretty much just the purpose of implementing JS's `?.` - is a pretty dirty function.

tenplus1:
> Looks fine as the "minimal change" for the sounds, but ideally `get_node_vars` with its 4 return values - which fulfill pretty much just the purpose of implementing JS's `?.` - is a pretty dirty function.

This is how I handle it in Stairs Redo, seems a lot tidier too :)

appgurueu (approved):

PR #3062, opened 2023-09-14
Fix corals and kelps duping with mesecons' sticky piston

zmv7:
This pull request fixes a glitch/exploit that could be used to dupe `default:coral_skeleton` and `default:sand` nodes infinitely.
### Before patch
Place mesecons' sticky piston next to cyan/green/pink coral or `default:sand_with_kelp` and switch the piston couple times. In the result, extra `default:coral_skeleton` or `default:sand` node will appear on each piston retraction.
### After patch
The cyan/green/pink corals and `default:sand_with_kelp` are moving by piston normally, without creating extra nodes.  

screenshot1
screenshot2

### DIFF
--- a/mods/default/nodes.lua
+++ b/mods/default/nodes.lua
@@ -2024,7 +2024,7 @@ minetest.register_node("default:sand_with_kelp", {
 		return itemstack
 	end,
 
-	after_destruct  = function(pos, oldnode)
+	after_dig_node = function(pos, oldnode, oldmetadata, digger)
 		minetest.set_node(pos, {name = "default:sand"})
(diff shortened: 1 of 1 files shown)


appgurueu (approved):
`after_destruct` is called by `minetest.remove_node`. This is indeed wrong - when a mod wants to remove a node, that node shouldn't turn into sand; it is reasonable for the sand and the kelp attached to it to be gone.

Using `after_dig_node` instead makes more sense. You don't have to add unused parameters though.

sfan5 (approved):
LGTM

PR #3067, opened 2023-09-26
Fix bed night skip race condition

### LINKED ISSUE #3066: Spam-tapping on beds and more...
I thought that the bug I encountered was related to the "beds" mod (maybe it still is.)

https://github.com/minetest/minetest_game/assets/113429882/416645a3-9dec-437d-96b6-7c0fa3e679e8

Short description:
The current day was 92;
I spam-tapped on the bed to lie down and get up again and again;
After waking up and seeing "You can only sleep at night", the current day was 99.

==========

But then I remembered about a bug I found earlier.
I was spam-tapping under myself, swimming in lava. Somehow Sam has become invincible (although he can't interact with the world. (place blocks, execute commands etc.))

https://github.com/minetest/minetest_game/assets/113429882/4d5000e0-3609-43b0-af2a-625b37515cf7

https://github.com/minetest/minetest_game/assets/113429882/06f957c4-851c-43cc-b1e8-7f3fc5124760

Sorry for the low quality, GitHub doesn't allow files larger than 10MB.
Also, I don't fully understand how to replicate it. Sometimes it does, sometimes it doesn't.

If it helps, I was playing build 5.7.0 from F-Droid.

appgurueu:
Closes #3066 (partially, the other issue is an engine issue). How to test:

* Reproduce the bug. This is easiest if you introduce a `print` in `beds.skip_night`. Otherwise, don't spam click too fast, or the daytime won't have changed when the next set time of day hits, so the days counter will only be incremented by one. You have a whole 2 seconds to make your clicks, make about 3-4 clicks in that time (definitely don't make clicks with a < 100 ms (serverstep) frequency).
* Verify that the fix works.

### DIFF
--- a/mods/beds/functions.lua
+++ b/mods/beds/functions.lua
@@ -181,6 +181,26 @@ function beds.skip_night()
 	minetest.set_timeofday(0.23)
 end
 
+local update_scheduled = false
+local function schedule_update()
+	if update_scheduled then
+		-- there already is an update scheduled; don't schedule more to prevent races
+		return
+	end
+	update_scheduled = true
+	minetest.after(2, function()
+		update_scheduled = false
+		if not is_sp then
+			update_formspecs(is_night_skip_enabled())
+		end
+		if is_night_skip_enabled() then
+			-- skip the night and let all players stand up
+			beds.skip_night()
+			beds.kick_players()
+		end
+	end)
+end
+
 function beds.on_rightclick(pos, player)
 	local name = player:get_player_name()
 	local ppos = player:get_pos()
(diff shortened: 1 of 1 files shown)


mods/beds/functions.lua
```diff
@@ -181,6 +181,25 @@ function beds.skip_night()
 	minetest.set_timeofday(0.23)
 end
 
+local update_scheduled = false
+local function schedule_update()
+	if update_scheduled then
+		-- there already is an update scheduled; don't schedule more to prevent races
+		return
+	end
+	update_scheduled = true
```
sfan5, line 190: this is never set to false?

appgurueu: oops, fixed

Issue #3072, opened 2023-11-01
Vessel description does not update


rollerozxa:
Regression caused by https://github.com/minetest/minetest_game/commit/0c2ee1e41d71899f0d5e1da603a6d125a7cda9e9, making vessel's infotext not update. Bookshelves were broken too by this commit but seems to have been fixed since then.

Niklp09:
Bookshelves were fixed in https://github.com/minetest/minetest_game/commit/cd6ea53a524bfb6fdc95cfcd23112dc970fc15f9

PR #3073, opened 2023-11-01
Fix not updating vessel shelf infotext

Niklp09:
Fixes https://github.com/minetest/minetest_game/issues/3072


appgurueu (approved):
Added `on_metadata_inventory_move` because `update_vessels_shelf` also updates the formspec, which should be triggered on inventory move actions as well.

*Technically* we could just do `on_metadata_inventory_(put|take|move) = update_vessels_shelf` but that would be error-prone if someone were to add a second parameter to `update_vessels_shelf` (since we would be passing extraneous arguments).

PR #3074, opened 2023-11-04
Remove wearing screwdriver

alwayshopeless:
The screwdriver in the game does not exist to make the game less casual, wearing for the screwdriver is the worst idea you could come up with. If a player wants to unwrap an item and doesn't want to do it manually, do you think he would want to craft 100 non-stacking screwdrivers?

### DIFF
--- a/mods/screwdriver/init.lua
+++ b/mods/screwdriver/init.lua
@@ -157,10 +157,6 @@ screwdriver.handler = function(itemstack, user, pointed_thing, mode, uses)
 		minetest.check_for_falling(pos)
 	end
 
-	if not minetest.is_creative_enabled(player_name) then
-		itemstack:add_wear_by_uses(uses or 200)
-	end
-
 	return itemstack
 end
 


Andrey2470T:
I support this PR as it doesn't make sense for this sort of the tool to be weared since it is actually not damaged during usage.

Montandalar:
The screwdriver that rotates entire blocks should not be worn? I think I'd damage my screwdriver too if I tried to move a metre cube around with it. Also, have you ever tried to torque something hard with a screwdriver? It's quite possible to damage screws or even the driver. They wear out.

Why not use a mod that overrides the screwdriver to not wear? In fact, why not get a mod like Rhotator or Omnidriver?

appgurueu:
Unless there was a sweeping consensus that the current behavior *can not possibly be desirable*, we have to consider this a feature which may not be merged in MTG's "maintenance-only" mode as to not alter game mechanics.

sfan5:
Agreed.

PR #3080, opened 2023-12-10
changes max stack of beds and buckets

DragonWrangler1:
changes the max stack of beds and buckets from 1 to the normal 99.

### DIFF
--- a/mods/beds/api.lua
+++ b/mods/beds/api.lua
@@ -35,7 +35,7 @@ function beds.register_bed(name, def)
 		paramtype = "light",
 		paramtype2 = "facedir",
 		is_ground_content = false,
-		stack_max = 1,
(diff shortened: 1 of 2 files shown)


appgurueu:
Sorry, but Minetest Game is in a maintenance-only mode right now. Unless you can make a very compelling case why (1) the current behavior can be seen as a bug; (2) changing this behavior won't break mechanics of mods and games based on MTG, we'll likely close this PR. (I doubt both points. Having "stacks of beds" or "stacks of buckets" doesn't make much sense. The stack max of 1 is there for a reason. Given the mechanics associated with both, I could very well imagine that allowing absurd stacks could negatively impact gameplay mechanics.)

DragonWrangler1:
closing pr since I made a mistake.

PR #3081, opened 2023-12-15
Use a faster and newer LuaCheck

Panquesito7:
Things added/changed:

- Use a faster and newer LuaCheck.


appgurueu (approved):

.github/workflows/build.yml
```diff
@@ -1,16 +1,12 @@
-name: build
+name: luacheck
```
sfan5, line 1: please rename the file too

Issue #2966, opened 2022-07-12
Hidden fireflies don't work outside vanilla. Change code seems not to work.


CowboyLva:
Tested on 5.5.1 Arch release(includes mtg in the package) and before that self compiled 5.5.0 with a stable branch of mtg.

Both have same results, fireflies work on vanilla biomes, but once used outside those the hide/unhide fuctions seem not to work.

I can get permanent fireflies by changing hidden fireflies to fireflies. Said fireflies will not change to their hidden variant.

Tested mods Extra Biomes(ebiomes), Ethereal(ethereal).

sfan5:
> I'm kinda surprised fireflies use node timers started in `on_generated` rather than an ABM for this...

Don't you mean an LBM? You don't want an ABM to happen constantly when node timers are already running.

appgurueu:
> > I'm kinda surprised fireflies use node timers started in `on_generated` rather than an ABM for this...
> 
> Don't you mean an LBM? You don't want an ABM to happen constantly when node timers are already running.

I had intended replacing the node timers with the ABM entirely?

JoseDouglas26:
Would it be possible to resolve this issue by creating an API for fireflies? A function of this API could be something like fireflies.register_fireflies_decoration(param1, param2, ..., biomes), where the biomes parameter is a list of biomes to register the decorations. Mods (Ethereal, Extra Biomes, etc) could use this API to add firefly decoration.

appgurueu:
After reconsideration, I think we should probably fix this by replacing the `on_generated` with an LBM as suggested by sfan5, and possibly also starting the node timer in `on_construct` for manually placed fireflies; I believe an ABM would probably add quite some unnecessary overhead for the relatively sparse fireflies (though I don't know the ABM implementation details too well), and also require larger changes. Thoughts?

PR #3087, opened 2023-12-24
Fireflies API: An idea to resolve issue #2966

JoseDouglas26:
This is my initial idea to resolve the issues reported in issue #2966. Using the fireflies API, mods like Ethereal and Extra Biomes can register the decoration of fireflies in their custom biomes.

### DIFF
--- a/mods/fireflies/api.lua
+++ b/mods/fireflies/api.lua
@@ -4,7 +4,7 @@ fireflies = {}
 -- If necessary, add other functions related to fireflies here
 
 --------------------------------------------------------------
--- Global function to record the decoration of fireflies. 
+-- Global function to record the decoration of fireflies.
 -- Perhaps other parameters can be added to improve decoration registration control.
 function fireflies.register_decoration(biomes, place_on)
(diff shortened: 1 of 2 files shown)


JoseDouglas26:
The idea is that this API evolves to allow greater control over recording firefly decorations.

appgurueu (commented):
The basic idea of this PR - adding an API to register firefly decorations - is good. However, please remove redundant comments, please document the API in `game_api.txt`.

Code-wise, three things seem fishy to me: (1) Reusing decoration names; (2) not using the biomes parameter when registering decorations for mgv6; (3) the two calls to `register_decoration` seem fishy; I believe there should ultimately only be one if this is refactored properly.

Design-wise, all these functions basically do is set some different defaults for deco registrations and do two registrations, one for high, one for low. There's some potential for deduplication here. What if you made the function take a table of "overrides", which would be applied "on top" of the default values?

Ultimately, I frankly also have to say that I think this is low priority. Without this, mods that register biomes just have to register the decorations themselves (which also gives them finer-grained control).

PR #3089, opened 2023-12-24
Fix some placer nil checks

### LINKED ISSUE #2993: minetest.place_node() causes exception in default.log_player_action()
The working_villagers mod uses ```minetest.place_node()``` to plant saplings, for example, "default:sapling".

When placed, it calls ```default.sapling_on_place()```, which calls ```default.log_player_action(placer, "places node", sapling_name, "at", pos)```.  When the "on_place" callback is triggered via ```place_node()```, the ```placer``` is nil.

The log_player_action() derefs the nil player and crashes.
```lua
function default.log_player_action(player, ...)
	local msg = player:get_player_name()
	...
end
```

A simple fix is to add a guard to protect against player == nil.
```lua
function default.log_player_action(player, ...)
	if player == nil then
		return
	end
	local msg = player:get_player_name()
	...
end
```
I've looked through the c++ code (ModApiEnvMod::l_place_node() and ScriptApiItem::item_OnPlace()) and it looks like the player will always be null.
This appears to have been introduced with https://github.com/minetest/minetest_game/pull/2941

See https://github.com/theFox6/working_villages/issues/32 for more details in reproducing the issue.

sfan5:
closes #2993

not tested again, I don't have my dev setup

### DIFF
--- a/mods/carts/cart_entity.lua
+++ b/mods/carts/cart_entity.lua
@@ -413,7 +413,7 @@ minetest.register_craftitem("carts:cart", {
 		minetest.sound_play({name = "default_place_node_metal", gain = 0.5},
 			{pos = pointed_thing.above}, true)
 
-		local player_name = placer and placer:get_player_name()
+		local player_name = placer and placer:get_player_name() or ""
 		if not minetest.is_creative_enabled(player_name) then
 			itemstack:take_item()
 		end
--- a/mods/carts/cart_entity.lua
+++ b/mods/carts/cart_entity.lua
@@ -413,7 +413,8 @@ minetest.register_craftitem("carts:cart", {
 		minetest.sound_play({name = "default_place_node_metal", gain = 0.5},
 			{pos = pointed_thing.above}, true)
 
-		if not minetest.is_creative_enabled(placer:get_player_name()) then
+		local player_name = placer and placer:get_player_name()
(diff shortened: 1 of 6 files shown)


mods/carts/cart_entity.lua
```diff
@@ -413,7 +413,8 @@ minetest.register_craftitem("carts:cart", {
 		minetest.sound_play({name = "default_place_node_metal", gain = 0.5},
 			{pos = pointed_thing.above}, true)
 
-		if not minetest.is_creative_enabled(placer:get_player_name()) then
+		local player_name = placer and placer:get_player_name()
```
appgurueu, line 416: Shouldn't this have an `or ""`? The docs for `minetest.is_creative_enabled` say:

> name will be `""` for non-players or if the player is unknown.

appgurueu (commented):
I've tested that calling `minetest.place_node` doesn't result in crashes anymore for butterflies, saplings, and TNT. (The others are more work to check, since we'd have to fulfill all the preconditions so that they don't bail out early for a different reason. But the code looks fine.)

mods/carts/cart_entity.lua
sfan5, line 416:
```suggestion
		local player_name = placer and placer:get_player_name() or ""
```

PR #3092, opened 2024-01-04
Tweak capitalization in Esperanto translation

JadedCtrl:
This switches item-description translations to the more common name-casing style in Esperanto: Only capitalizing the first word in names, like in Spanish, French, Italian, Polish, etc.


appgurueu (approved):
(Assuming this is more "idiomatic", which I can't really judge)

Issue #2987, opened 2022-09-11
Butterflies don't respect other nodes on_rightclick


Skamiz:
**Observed behavior**
When I try to open a chest with a butterfly in hand the butterfly is always placed.

**Expected behavior**
The butterfly is placed only if I hold the sneak key. Just like other placeable nodes.

Edit:
Just noticed that the fireflies have the same issue.

Skamiz:
Oh, hey. just found that butterflies and fireflies can't be built to buildable_to nodes.

JoseDouglas26:
Maybe if someone changes the butterflies' on_place definition to the following definition, this bug will be resolved. The definition follows:

```lua
on_place = function(itemstack, placer, pointed_thing)
        local player_name = placer:get_player_name()
	local pos = pointed_thing.above
	local node = minetest.get_node(pointed_thing.under)

	if not minetest.is_protected(pos, player_name) and not 
        minetest.is_protected(pointed_thing.under, player_name) and 
        minetest.get_node(pos).name == "air" then
		if minetest.registered_nodes[node.name] and minetest.registered_nodes[node.name].on_rightclick then
			if placer and placer:get_player_control().sneak then
				minetest.set_node(pos, {name = "butterflies:butterfly_"..name})
				minetest.get_node_timer(pos):start(1)
				itemstack:take_item()
			else
				return minetest.registered_nodes[node.name].on_rightclick(pointed_thing.under, node, placer, itemstack)
			end
		end
	end
	return itemstack
end

PR #3093, opened 2024-01-07
Remove on_place callback from butterflies

JoseDouglas26:
New version of #3086

Definition of on_place callback removed from butterflies.
Suggestion made by: @SmallJoker

Related issue: #2987


SmallJoker:
Where do you start the timer?

PR #3094, opened 2024-01-08
Add /spawn command by default

Sovenok-Hacker:
I think it will be helpful since it is hard to find a mod for it now


appgurueu:
Sorry, but this is not an option:

* We're in feature freeze and this is, pretty clearly, a feature (and probably a current one at that);
* In its current form, this completely breaks the game - whenever things get out of hand, you can just teleport to spawn.

I'd suggest you to just make a mod to customize MTG to your liking. Then it won't be hard to find a mod for it anymore :)

PR #3097, opened 2024-01-15
Allow overriding the loop flag on `player_api.set_animation`

### LINKED ISSUE #3095: Allow overriding the loop flag on `player_api.set_animation`
### Description

Currently, I'd like to apply an animation to the player, but once it ends, just stop it there and not loop it anymore.
I know I could manually do this by using `player:set_animation`, but that'd also involve changing the collisionboxes and doing other checks, which I'd prefer if `player_api` does that.

Would be awesome if the `player_api.set_animation` function provides an optional parameter to override the loop flag, just like the animation speed. Let me know your thoughts about this.

Panquesito7:
Things added/changed:

- Allow overriding the loop flag on `player_api.set_animation`.
  - Closes #3095.

### DIFF
--- a/mods/player_api/api.lua
+++ b/mods/player_api/api.lua
@@ -71,7 +71,7 @@ function player_api.set_model(player, model_name)
 	player_data.model = model_name
 	-- Clear animation data as the model has changed
 	-- (required for setting the `stand` animation not to be a no-op)
-	player_data.animation, player_data.animation_speed = nil, nil
+	player_data.animation, player_data.animation_speed, player_data.animation_loop = nil, nil, nil
 
 	local model = models[model_name]
 	if model then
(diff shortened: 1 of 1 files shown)


appgurueu (commented):
Added a commit to implement `loop` analogeously to `speed`. In particular, it needs to be remembered, otherwise you wouldn't be able to change `loop` when setting the same animation.

I haven't this tested yet however, make sure it still works for you. I'll test this later.

Panquesito7:
> Added a commit to implement `loop` analogeously to `speed`. In particular, it needs to be remembered, otherwise you wouldn't be able to change `loop` when setting the same animation.
> 
> I haven't this tested yet however, make sure it still works for you. I'll test this later.

Thanks! Tested it, works perfectly.

PR #3099, opened 2024-01-27
Make bugs only pointable with Bug Net

cx384:
### Problem
Sometimes a firefly or a butterfly is in your way when you want to dig something, and it can't be destroyed by punching it.
It basically behaves like punching stone with a bare hand. This is unreasonable and annoying.

### What does this PR do?
This PR makes fireflies and butterflies only pointable using a bug net.
It utilizes the new pointabilities feature.

### Alternatives
- Make fireflies and butterflies digable without dropping anything.
- Move fireflies and butterflies to an adjacent position if someone tries to punch them.

### DIFF
--- a/mods/butterflies/init.lua
+++ b/mods/butterflies/init.lua
@@ -3,6 +3,9 @@
 -- Load support for MT game translation.
 local S = minetest.get_translator("butterflies")
 
+-- Legacy compatibility, when pointabilities don't exist, pointable is set to true.
+local pointable_compat = not minetest.features.item_specific_pointabilities
+
 -- register butterflies
 local butter_list = {
 	{"white",  S("White Butterfly")},
@@ -33,7 +36,7 @@ for i in ipairs (butter_list) do
 		sunlight_propagates = true,
 		buildable_to = true,
 		walkable = false,
-		pointable = false,
+		pointable = pointable_compat,
 		groups = {catchable = 1},
 		selection_box = {
 			type = "fixed",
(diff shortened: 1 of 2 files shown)


appgurueu (commented):
I don't think the "legacy compatibility" works if an old client connects to a new server: The server will set `pointable = false` for the fireflies and butterflies. The old client will not be able to point at them, since it doesn't understand the `pointabilities`.

I'm not sure what to do about this however. I currently see three options:

* Using the current API, we could let fireflies remain pointable, but add `pointabilities = false` to all tools by default (for example by monkey-patching `minetest.register_tool`). This is quite hacky, however.
* Perhaps we could adapt the pointabilities API to allow specifying a fallback for backwards-compatibility or similar?
* Just ignore it. Not all that bad, actually. New clients on new servers will be fine. Old servers will be fine to (and just not take advantage of this). And this is a pretty minor feature anyways.

cx384:
Yeah, the compatibility is only for old Minetest versions which use a new MTG version.
I don't know how compatibility issues are typically handled in Minetest, but the first two solutions you suggested are too much imo, considering how small the change is.
I'm not even sure if the legacy support I added is needed, since a new MTG version may depend on Minetest 5.9.0 anyway.

The only good way I see to solve this is to add a feature to Minetest which allows sending different item definitions to clients depending on the client version. However, this would be a huge effort and it will rarely be used. So I think it is reasonable to just ignore it, in case this PR gets merged.

sfan5:
To be clear (in the past) MTG did require the server to be up-to-date but we still couldn't use something (e.g. formspecs) that flat-out did not work on older clients.

appgurueu (approved):
I can live with older clients being unable to catch bugs. (And since bugs have `buildable_to = true`, it's not like they are indestructible for older clients.)

PR #3100, opened 2024-02-12
Secure TNT

acmgit:
Enables a new option. TNT is only working below a given height.
```
# The height below tnt can explode.
tnt_work_below_height (Under tnt explodes) int 31000
```

### DIFF
--- a/mods/tnt/init.lua
+++ b/mods/tnt/init.lua
@@ -5,7 +5,6 @@ tnt = {}
 -- Load support for MT game translation.
 local S = minetest.get_translator("tnt")
 
-
 -- Default to enabled when in singleplayer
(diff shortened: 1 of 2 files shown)


appgurueu:
Sorry, but I think this feature is too oddly specific to warrant inclusion in Minetest Game, especially as we are in a feature freeze. Consider overriding `minetest.is_protected` or `tnt.boom` or similar instead.

rudzik8:
@acmgit in case you're interested, I decided to incorporate your work into Voxelgarden: https://codeberg.org/voxelgarden/voxelgarden/pulls/52

acmgit:
Glad to hear this, of course, you have my agree to use this.
The mod is on my servers and i have changed it, to secure against griefing.

PR #3101, opened 2024-02-24
Set field_enter_after_edit for search fields

grorp:
Trivial change to improve the user experience on Android


sfan5 (approved):

appgurueu (approved):
This had me slightly worried about older clients, but it seems we handle unknown formspec elements rather gracefully so this is good to go.

PR #3105, opened 2024-03-25
Add workflow that runs simple test with the engine present

sfan5:
should be useful even if future development is going to be quite limited


appgurueu (approved):
Looks ok, doesn't fail. Why put it in `default` rather than a separate "test" mod though?

sfan5:
> Why put it in `default` rather than a separate "test" mod though?

Don't see the advantage for <10 LOC.

PR #3107, opened 2024-03-31
refactor of default/tools.lua

MistUnky:
Clean up and refactoring of tools.lua. Some of the names might need to be changed though.

### DIFF
--- a/mods/default/tools.lua
+++ b/mods/default/tools.lua
@@ -19,438 +19,128 @@ minetest.override_item("", {
 	}
 })
 
+-- Default-type tool registration function
+
+function default.register_tool_of_type(name, desc, img, wimg, punch_interval, max_drop_lvl, gc1, gc2, gc3, use_count, maxlvl, dmg, extra_groups, mat, typename, typedesc, typebreak, typetool, typeshape)
+	extra_groups[typetool] = 1
+	local mygroupcaps = {}
+	mygroupcaps[typebreak] = {times={[1]=gc1, [2]=gc2, [3]=gc3}, uses=use_count, maxlevel=maxlvl}
+	minetest.register_tool("default:"..typename.."_"..name, {
+		description = S(desc.." "..typedesc),
+		inventory_image = img,
+		wield_image = wimg,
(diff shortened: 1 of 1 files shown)


MoNTE48:
You made simple things complex without providing any benefit or optimization in return.

MistUnky:
@MoNTE48 Is there anything specific you would change about it?

appgurueu:
I think deduplicating this code, and potentially exposing an API, makes sense; some things (such as recipes) simply follow the same pattern for the same tool type.

However, I have to agree that the way this PR goes about it is presently not an improvement. First of all, the overly long argument lists are very messy. Passing a definition table would be much preferable (and would also make it easier to pass arguments on unchanged).

Let's first determine which fields *only* depend on tool type (sword / shovel / pick / axe / hoe) and material type *name* (wood / stone / iron / bronze / mese / diamond), or are invariant:

* Tool item name;
* Tool inventory image;
* Tool wield image;
* (Technically also tool description, but for translation reasons we may not leverage this);
* `sound = {breaks = "default_tool_breaks"}`;
* Tool recipe

since most of these depend on both type and material, I don't see much benefit in having a `default.register_tool` function at all; instead, simply have `register_pick`, `register_shovel` etc., potentially as local functions if you don't want to expose them. The only other thing that varies are the `tool_capabilities`, which need to be passed as a parameter.

So I would propose refactoring to `register_<tool>(material, tool_capabilities)` functions.

You could also consider introducing a `register_tools(material, material_properties)`, where `material_properties` has fields `full_punch_interval`, `max_drop_level`, `groupcap`, `damage_fleshy` to register tools of all five types given a material type, though I'm not so sure about this.

---

`S(desc.." "..typedesc)` means the translation updating script will be unable to pick up these strings. You should take an entire description instead and write out `S("Wooden Pickaxe")` etc. each time.

MistUnky:
@appgurueu I made the functions local and rewrote them as you specified. I made the function arguments register_<tool>(tool_def) rather than register_<tool>(name, tool_def).

PR #3108, opened 2024-04-02
Log bucket liquid placement

Emojigit:
This PR adds code to log liquid placements done via buckets. Previously, only failed attempts were logged.

This PR is ready for review.


appgurueu (approved):
Code looks fine. `text` in `log_action` should probably be called something like `action`? Reordering the parameters to be in the order the message is built (roughly "\<name> did \<action> at \<pos>") could also help make this slightly more readable IMO.

I can definitely see the logging for placing liquids being useful against griefers.

(Why log picking up liquids though?)

Tested briefly, does what it says on the tin, doesn't blow up.

I'm not sure if this technically counts as maintenance, but adding some more logging is by no means problematic.

Emojigit:
> Reordering the parameters to be in the order the message is built (roughly "<name> did <action> at <pos>") could also help make this slightly more readable IMO.

I kept the order mostly for consistency with the `check_protection` function above.

> (Why log picking up liquids though?)

Like the engine log digging node, IMO logging this is also needed.

mods/bucket/init.lua
```diff
@@ -166,18 +171,21 @@ minetest.register_craftitem("bucket:bucket_empty", {
 			local source_neighbor = false
 			if liquiddef.force_renew then
 				source_neighbor =
-					minetest.find_node_near(pointed_thing.under, 1, liquiddef.source)
+					minetest.find_node_near(pos, 1, liquiddef.source)
 			end
-			if not (source_neighbor and liquiddef.force_renew) then
-				minetest.add_node(pointed_thing.under, {name = "air"})
+			if source_neighbor and liquiddef.force_renew then
+				log_action(pos, pname, "picked up " .. liquiddef.source .. " (force renewed)")
```
sfan5, line 177: I don't think logging the force-renewed or not here is that important. You can always check this by looking into the node def.

mods/bucket/init.lua
```diff
@@ -166,18 +171,21 @@ minetest.register_craftitem("bucket:bucket_empty", {
...
+			else
+				minetest.add_node(pos, {name = "air"})
```
sfan5, line 179: unrelated but this could be `minetest.remove_node(pos)`

Emojigit: This exists before this PR... Is it good to fix it in this PR?

appgurueu: > This exists before this PR... Is it good to fix it in this PR?

Yes, generally you're free to fix some style issues as you go - just don't bloat the scope of the PR (e.g. only fix style issues in code you're actually touching, the rest is better left to cleanup PRs).

sfan5: There's actually a second occurrence of this in MTG so this is better done in a separate commit.

mods/bucket/init.lua
sfan5: I see, you're right.

PR #3112, opened 2024-04-10
Add API to weather mod

### LINKED ISSUE #2913: Weather mod needs API to disable it

sfan5:
resolves #2913, replaces #3020

#### Key points

* no is_enabled/set_enabled because that's problematic with multiple mods
* API always works, even on singlenode worlds or with `enable_weather = false`
  * in the latter case it does not do anything and won't interfere (*by default*)
* only one way to modify behavior: by overriding

#### How to test

```lua
local old_get = weather.get
weather.get = function(player)
	local data = old_get(player)
	data.clouds = data.clouds or {}
	data.clouds.thickness = 0.1
	return data
end
```

everything is as before except that clouds are paper-thin

### DIFF
--- a/mods/weather/api.lua
+++ b/mods/weather/api.lua
@@ -1,0 +1,32 @@
+local CYCLE = 8 -- Time period of cyclic clouds update in seconds
+
+weather = {}
+
+-- default implementation is empty
+function weather.get(player)
+	return {}
+end
+
+local function do_update()
+	for _, player in ipairs(minetest.get_connected_players()) do
+		local params = weather.get(player)
+		assert(params ~= nil, "weather.get() must not return nil")
+		if params.clouds then
+			player:set_clouds(params.clouds)
+		end
+		if params.lighting then
+			player:set_lighting(params.lighting)
+		end
+	end
+end
+
+local function cyclic_update()
+	do_update()
+	minetest.after(CYCLE, cyclic_update)
+end
+minetest.after(0, cyclic_update)
+
+-- Update on player join to instantly alter clouds from the default
+minetest.register_on_joinplayer(function(player)
+	do_update()
+end)
--- a/game_api.txt
+++ b/game_api.txt
@@ -1169,3 +1169,21 @@ the log.
  * after logging the action, the original callback (if any) is called
  * `def`     See [Node definition]
  * `name`    Description of the node in the log message
+
+
+Weather API
+-----------
+
+The weather mod will constantly adjust weather effects seen by the player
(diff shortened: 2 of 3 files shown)


mods/weather/api.lua
```diff
@@ -0,0 +1,32 @@
+local CYCLE = 8 -- Time period of cyclic clouds update in seconds
+
+weather = {}
+
+-- default implementation is empty
+function weather.get(player)
+	return {}
```
appgurueu, line 7: Reading this is a bit misleading: This function is overwritten in `init.lua`. So the first thing a reader is going to think is "wait, doesn't this disable weather?"

Why define it here at all? I'd just leave a comment here along the lines of "`weather.get` is defined in `init.lua`". `do_update` is not called at load time, so this should be fine.

Or if you do want to leave a function here, to make the "dependency" explicit, why not something like `error("this should be overridden in init.lua")`?

Though the ideal solution would probably be to move the appropriate code from `init.lua` into `api.lua`, or even something like a separate file `default.lua` or similar, such that you could `dofile` it from here and have it be loaded in time as to not confuse a reader where `weather.get` comes from.

appgurueu (commented):
Tested, works. Code looks fine besides my one minor comment. I like the simplicity of this. The docs are in line with the rest of `game_api.txt`, but formatted a bit weirdly if you view them as Markdown (for example the spacing used for delimiting will just be collapsed into a single space; the `weather.get = function(player)` "inline code" should probably be a subheading). Why not strive for proper Markdown formatting here?

sfan5:
> Why not strive for proper Markdown formatting here?

I'll try to reformat it.

mods/weather/api.lua
sfan5: init.lua redefines this, but only conditionally. For example it doesn't when `enable_weather` is false.

appgurueu: Right. A comment explaining this may be a good idea. The early return is easy to miss.

PR #3116, opened 2024-05-13
Changes on butterflies placement checks

### LINKED ISSUE #2987: Butterflies don't respect other nodes on_rightclick

### LINKED ISSUE #3102: Avoid using the global keyword `table` as a function argument name
In the file [functions.lua](https://github.com/minetest/minetest_game/blob/master/mods/default/functions.lua), many functions use `table` as an argument name. Since `table` is also a global variable, it ends up being shadowed or redefined locally. This is generally regarded as a poor programming practice as it can be confusing what `table` is referring to.

JoseDouglas26:
Fixes #2987 (by using `minetest.item_place`)
Fixes #3102 (by renaming the parameters)

### DIFF
--- a/mods/butterflies/init.lua
+++ b/mods/butterflies/init.lua
@@ -50,13 +50,13 @@ for i in ipairs (butter_list) do
 			if minetest.is_protected(pos, player_name) or minetest.is_protected(pointed_thing.under, player_name) then
 				return
 			end
-			
+
 			local _, sucess = minetest.item_place(itemstack, placer, pointed_thing)
-			
+
 			if sucess then
 				minetest.get_node_timer(pos):start(1)
 			end
-			
+
 			return itemstack
 		end,
 		on_timer = function(pos, elapsed)
@@ -86,13 +86,13 @@ for i in ipairs (butter_list) do
 			if minetest.is_protected(pos, player_name) or minetest.is_protected(pointed_thing.under, player_name) then
 				return
 			end
-			
+
 			local _, sucess = minetest.item_place(itemstack, placer, pointed_thing)
-			
+
 			if sucess then
 				minetest.get_node_timer(pos):start(1)
 			end
-			
+
 			return itemstack
 		end,
 		on_timer = function(pos, elapsed)
--- a/mods/butterflies/init.lua
+++ b/mods/butterflies/init.lua
@@ -44,7 +44,6 @@ for i in ipairs (butter_list) do
 		},
 		floodable = true,
 		on_place = function(itemstack, placer, pointed_thing)
-			local player_name = placer and placer:get_player_name() or ""
 			local pos = pointed_thing.above
 
 			local _, success = minetest.item_place(itemstack, placer, pointed_thing)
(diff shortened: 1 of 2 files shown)


JoseDouglas26:
Sorry, I made a commit to the master of my fork to resolve another issue and it ended up in this PR too. Now this PR can fix #3102

mods/butterflies/init.lua
```diff
@@ -47,13 +47,16 @@ for i in ipairs (butter_list) do
 			local player_name = placer and placer:get_player_name() or ""
 			local pos = pointed_thing.above
 
-			if not minetest.is_protected(pos, player_name) and
-					not minetest.is_protected(pointed_thing.under, player_name) and
-					minetest.get_node(pos).name == "air" then
-				minetest.set_node(pos, {name = "butterflies:butterfly_"..name})
+			if minetest.is_protected(pos, player_name) or minetest.is_protected(pointed_thing.under, player_name) then
```
SmallJoker, line 50: Please keep the lines below 80 characters.
https://dev.minetest.net/Lua_code_style_guidelines#Lines.2C_spaces.2C_and_indentation

Same for the hidden butterfly code.

mods/butterflies/init.lua
```diff
@@ -47,13 +47,16 @@ for i in ipairs (butter_list) do
...
+				return
+			end
+
+			local _, sucess = minetest.item_place(itemstack, placer, pointed_thing)
```
SmallJoker, line 54: Correction: `success`
Same for the hidden butterfly code.

mods/default/functions.lua
```diff
@@ -2,149 +2,149 @@
 -- Sounds
 --
 
-function default.node_sound_defaults(table)
-	table = table or {}
-	table.footstep = table.footstep or
+function default.node_sound_defaults(tbl)
```
SmallJoker, line 5: I'd rather name this parameter `sounds`, but I'll leave the decision up to you.

mods/butterflies/init.lua
SmallJoker, line 50: This protection check might not be necessary at all. `core.item_place_node` (called by `minetest.item_place`) already performs protection checks.

JoseDouglas26: I wasn't sure if the check was done so I kept

mods/butterflies/init.lua
JoseDouglas26: Poor english, sorry about that

SmallJoker (approved):
I found that butterflies now do replace grass. In master, the butterfly is put to `pointed_thing.above`. However, this can be avoided by using sneak + place.

Works for me.

Emojigit:
I submitted another PR, #3125, which uses the engine's routine to handle butterfly and firefly placement and catching. Every possible interaction exception can be handled by having everything cared for by the engine.

JoseDouglas26:
So can I close this?

Emojigit:
> So can I close this (again)?
> 
> 

This PR has been approved by core developers, so I would suggest letting them decide.

JoseDouglas26:
I will close this PR. #3125 will be better thant that one

JoseDouglas26:
I will close this PR. #3125 will be better thant that one

PR #3128, opened 2024-06-12
Added soil_group param to plant definition

ssdaniel24:
Added this param to give ability to create something like mushrooms that can grow on something like stones etc.

### DIFF
--- a/game_api.txt
+++ b/game_api.txt
@@ -381,7 +381,9 @@ The farming API allows you to easily register plants and hoes.
 		steps = 8,                             -- How many steps the plant has to grow, until it can be harvested
 		-- ^ Always provide a plant texture for each step, format: modname_plantname_i.png (i = stepnumber)
 		minlight = 13,                         -- Minimum light to grow
-		maxlight = default.LIGHT_MAX           -- Maximum light to grow
+		maxlight = default.LIGHT_MAX,          -- Maximum light to grow
+		soil_group = "soil"                    -- Soil node group where plant can grow ("soil" by default).
+                                               -- Group rating must be at least 2 for plant growing.
 	}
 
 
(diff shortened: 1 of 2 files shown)


game_api.txt
```diff
@@ -381,7 +381,8 @@ The farming API allows you to easily register plants and hoes.
 		steps = 8,                             -- How many steps the plant has to grow, until it can be harvested
 		-- ^ Always provide a plant texture for each step, format: modname_plantname_i.png (i = stepnumber)
 		minlight = 13,                         -- Minimum light to grow
-		maxlight = default.LIGHT_MAX           -- Maximum light to grow
+		maxlight = default.LIGHT_MAX,          -- Maximum light to grow
+		soil_group = "soil"                    -- Soil node group where plant can grow ("soil" by default)
```
appgurueu, line 385: Please also document that the group rating must be at least 2. (Otherwise I would expect any nonzero rating to work.)

mods/farming/api.lua
```diff
@@ -279,6 +279,9 @@ farming.register_plant = function(name, def)
 	if not def.fertility then
 		def.fertility = {}
 	end
+	if not def.soil_group then
```
appgurueu, line 282: `def.soil_group = def.soil_group or "soil"`

mods/farming/api.lua
```diff
@@ -130,7 +130,7 @@ local function tick_again(pos)
 end
 
 -- Seed placement
-farming.place_seed = function(itemstack, placer, pointed_thing, plantname)
+farming.place_seed = function(itemstack, placer, pointed_thing, plantname, soil_group)
```
appgurueu, line 133: Why make this a param rather than doing `local soil_group = farming.registered_plants[plantname].soil_group`?

tenplus1:
Rather than adding group checks it may be worthwhile to add a growth_check function to custom crops, that way it can check what it needs inside the function and return a True value when they are met, kinda like this:

https://codeberg.org/tenplus1/farming/src/branch/master/init.lua#L378

PR #3130, opened 2024-06-16
Avoid using the global keyword table as a function argument name

### LINKED ISSUE #3102: Avoid using the global keyword `table` as a function argument name

Emojigit:
Fixes #3102

This PR is cherry-picked from #3116. This PR is ready for review, and has gone through one maintainer's approval in the previous PR.

### DIFF
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -2,149 +2,149 @@
 -- Sounds
 --
 
-function default.node_sound_defaults(table)
-	table = table or {}
-	table.footstep = table.footstep or
+function default.node_sound_defaults(tbl)
+	tbl = tbl or {}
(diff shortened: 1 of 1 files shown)


sfan5 (approved):

appgurueu (approved):
ps. nitpick re PR title: technically `table` is just a global variable name used by the lua standard library, not a keyword syntactically speaking ;)

PR #3131, opened 2024-06-16
Added can_grow function to plant definition

ssdaniel24:
Adds custom function to check if can grow - this is redesigned idea of #3128.

Thanks @tenplus1 for [advice](https://github.com/minetest/minetest_game/pull/3128#issuecomment-2171275630) to see his code:

<https://codeberg.org/tenplus1/farming/src/branch/master/init.lua#L378>

### DIFF
--- a/mods/farming/api.lua
+++ b/mods/farming/api.lua
@@ -191,11 +191,8 @@ end
 
 -- check if on wet soil
 farming.can_grow = function(pos)
-	local below = minetest.get_node({x = pos.x, y = pos.y - 1, z = pos.z})
-	if minetest.get_item_group(below.name, "soil") < 3 then
-		return false
-	end
-	return true
+	local below = minetest.get_node(pos:offset(0, -1, 0))
+	return minetest.get_item_group(below.name, "soil") >= 3
 end
 
 farming.grow_plant = function(pos, elapsed)
@@ -233,7 +230,7 @@ farming.grow_plant = function(pos, elapsed)
 		return
 	end
 
-	if (def.can_grow and def.can_grow(pos)) or (not def.can_grow and not farming.can_grow(pos)) then
+	if not (def.can_grow or farming.can_grow)(pos) then
 		tick_again(pos)
 		return
 	end
--- a/mods/farming/api.lua
+++ b/mods/farming/api.lua
@@ -189,6 +189,15 @@ farming.place_seed = function(itemstack, placer, pointed_thing, plantname)
 	return itemstack
 end
 
+-- check if on wet soil
+farming.can_grow = function(pos)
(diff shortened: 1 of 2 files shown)


mods/farming/api.lua
```diff
@@ -224,11 +224,15 @@ farming.grow_plant = function(pos, elapsed)
 		return
 	end
 
-	-- check if on wet soil
-	local below = minetest.get_node({x = pos.x, y = pos.y - 1, z = pos.z})
-	if minetest.get_item_group(below.name, "soil") < 3 then
-		tick_again(pos)
+	if def.can_grow and not def.can_grow(pos) then
 		return
+	else
```
appgurueu, line 229: Unnecessary `else` (you already have an early `return` right above). You could rewrite this as `if (not def.can_grow) or def.can_grow(pos) then ... end` however if you prefer that.

mods/farming/api.lua
```diff
@@ -224,11 +224,15 @@ farming.grow_plant = function(pos, elapsed)
...
+		-- check if on wet soil
```
appgurueu, line 230: I don't think this logic makes sense: If `can_grow` returns `true`, you still check for the `soil` group.

What you want instead is for `can_grow` to default to a function checking the soil group, no?

mods/farming/api.lua
```diff
@@ -224,11 +224,15 @@ farming.grow_plant = function(pos, elapsed)
 		return
 	end
 
-	-- check if on wet soil
-	local below = minetest.get_node({x = pos.x, y = pos.y - 1, z = pos.z})
-	if minetest.get_item_group(below.name, "soil") < 3 then
-		tick_again(pos)
+	if def.can_grow and not def.can_grow(pos) then
 		return
```
appgurueu, line 228: Shouldn't there be a `tick_again(pos)` here?

mods/farming/api.lua
```diff
@@ -189,6 +189,15 @@ farming.place_seed = function(itemstack, placer, pointed_thing, plantname)
 	return itemstack
 end
 
+-- check if on wet soil
+farming.can_grow = function(pos)
+	local below = minetest.get_node({x = pos.x, y = pos.y - 1, z = pos.z})
```
appgurueu, line 194: "new" way to write this would be `pos:offset(0, -1, 0)` or `vector.new(pos.x, pos.y - 1, pos.z)` if you want to code defensively

appgurueu (approved):
Makes sense to dehardcode. Thanks.

mods/farming/api.lua
sfan5:
```suggestion
	local below = minetest.get_node(vector.offset(pos, 0, -1, 0))
```

PR #3132, opened 2024-06-17
Fix crafting for fences etc. starting with colon

### LINKED ISSUE #3127: Registering fence/fence rail with a name starting in : creates node but no recipe
So if I register a fence using default it wont register recipe if fence starts with : node appears fine otherwise.
Fence gates work fine with names starting with : however they then have an extra : so its best to use any namespace initially. 

gate registers node like: minetest.register_node(":" .. name .. "_closed", fence_closed)
fence registers node like: minetest.register_node(name, def)

gate registers recipe after node, fence registers recipe before node.
both use name as output

Heres some code to test this yourself:

```
local text = "default_stone.png"
local matt = "default:stone"
default.register_fence(
            ":testmod:stone_fence",
            {
                description = "stone fence",
                texture = text,
                inventory_image = "default_fence_overlay.png^".. text .. "^default_fence_overlay.png^[makealpha:255,126,126",
                wield_image = "default_fence_overlay.png^".. text .. "^default_fence_overlay.png^[makealpha:255,126,126",
                material = matt,
                groups = {choppy = 2, oddly_breakable_by_hand = 2, flammable = 2},
                sounds = xcompat.sounds.node_sound_wood_defaults()
            }
            )

```
This works but there is no recipe. If you change ":testmod:stone_fence" to "testmod:stone_fence" the recipe works again.

Changing fence to match gate and then removing the : from the name works oddly enough. However I don't know if this might have unintended side effects.

Registering recipe after separately without : works aswel, when fence name starts with : and no changes to default. I wonder if some check to register recipe later if name starts with : could work? Still might register recipe twice if anyone else is doing this.

Emojigit:
Fixes #3127

This PR fixes the crafting recipes of fences, fence rails, and mese post lights. After the changes, even if the node name starts with a colon, the crafting recipe still exists.

This PR is ready for review.

### DIFF
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -499,7 +499,7 @@ function default.register_mesepost(name, def)
 		recipe = {
 			{'', 'default:glass', ''},
 			{'default:mese_crystal', 'default:mese_crystal', 'default:mese_crystal'},
-			{'', def.material, ''},
+			{'', material, ''},
 		}
 	})
 end
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -364,9 +364,7 @@ function default.register_fence(name, def)
 
 	-- Register crafting recipe, trim away starting colon if any
 	if not material then return end
-	if string.sub(name, 1,1) == ":" then
-		name = string.sub(name, 2)
-	end
+	name = string.gsub(name, "^:", "")
 	minetest.register_craft({
 		output = name .. " 4",
 		recipe = {
@@ -438,9 +436,7 @@ function default.register_fence_rail(name, def)
 
 	-- Register crafting recipe, trim away starting colon if any
 	if not material then return end
-	if string.sub(name, 1,1) == ":" then
-		name = string.sub(name, 2)
-	end
+	name = string.gsub(name, "^:", "")
 	minetest.register_craft({
 		output = name .. " 16",
 		recipe = {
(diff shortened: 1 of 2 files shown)


mods/default/functions.lua
```diff
@@ -364,10 +356,24 @@ function default.register_fence(name, def)
...
 	def.groups.fence = 1
 
+	local material = def.material
 	def.texture = nil
 	def.material = nil
 
 	minetest.register_node(name, def)
+
+	-- Register crafting recipe, trim away starting colon if any
+	if not material then return end
```
sfan5, line 366: why would this be nil?

mods/default/functions.lua
```diff
@@ -364,10 +356,24 @@ function default.register_fence(name, def)
...
+	if string.sub(name, 1,1) == ":" then
+		name = string.sub(name, 2)
+	end
```
sfan5, line 369: `name:gsub("^:, "")` or similar should do (please double check)

Emojigit: Done in 23104f2

Though the if-sub pattern should be faster, I agree that code tidiness is more important than the effectiveness pre-globalstep.

mods/default/functions.lua
sfan5: Makes sense, just document it in `game_api.txt` too.

Emojigit: Done in 3b39117

Weirdly only fence is listed in that API documentation, so only that one is fixed.

mods/default/functions.lua
appgurueu: Personally I would've preferred `name:match(":?(.*)")` ;)

sfan5 (approved):

PR #3133, opened 2024-06-17
Get rid of recursive call in default.dig_up

### LINKED ISSUE #3075: stack overflow due to `default.dig_up`
```
AsyncErr: Lua: Runtime error from mod '' in callback node_on_dig(): Runtime error from mod '' in callback getAuth(): stack overflow
stack traceback:
	[C]: in function 'read'
	/opt/minetest/bin/../builtin/game/auth.lua:14: in function </opt/minetest/bin/../builtin/game/auth.lua:12>
	[C]: in function 'get_player_privs'
	/opt/minetest/bin/../builtin/game/misc.lua:28: in function 'check_player_privs'
	/opt/minetest/bin/../mods/areas/api.lua:87: in function 'canInteract'
	/opt/minetest/bin/../mods/areas/interact.lua:5: in function 'is_protected'
	/opt/minetest/bin/../builtin/game/item.lua:473: in function 'node_dig'
	...st/bin/../games/minetest_game/mods/default/functions.lua:301: in function 'dig_up'
	...netest/bin/../games/minetest_game/mods/default/nodes.lua:1416: in function 'after_dig_node'
	/opt/minetest/bin/../builtin/game/item.lua:547: in function 'node_dig'
	...
	/opt/minetest/bin/../builtin/game/item.lua:547: in function 'node_dig'
	...st/bin/../games/minetest_game/mods/default/functions.lua:301: in function 'dig_up'
	...netest/bin/../games/minetest_game/mods/default/nodes.lua:1416: in function 'after_dig_node'
	/opt/minetest/bin/../builtin/game/item.lua:547: in function 'node_dig'
	...st/bin/../games/minetest_game/mods/default/functions.lua:301: in function 'dig_up'
	...netest/bin/../games/minetest_game/mods/default/nodes.lua:1416: in function 'after_dig_node'
	/opt/minetest/bin/../builtin/game/item.lua:547: in function 'node_dig'
	...st/bin/../games/minetest_game/mods/default/functions.lua:301: in function 'dig_up'
	...netest/bin/../games/minetest_game/mods/default/nodes.lua:1416: in function 'after_dig_node'
	/opt/minetest/bin/../builtin/game/item.lua:547: in function </opt/minetest/bin/../builtin/game/item.lua:460>
stack traceback:
	[C]: in function 'get_player_privs'
	/opt/minetest/bin/../builtin/game/misc.lua:28: in function 'check_player_privs'
	/opt/minetest/bin/../mods/areas/api.lua:87: in function 'canInteract'
	/opt/minetest/bin/../mods/areas/interact.lua:5: in function 'is_protected'
	/opt/minetest/bin/../builtin/game/item.lua:473: in function 'node_dig'
	...st/bin/../games/minetest_game/mods/default/functions.lua:301: in function 'dig_up'
	...netest/bin/../games/minetest_game/mods/default/nodes.lua:1416: in function 'after_dig_node'
	/opt/minetest/bin/../builtin/game/item.lua:547: in function 'node_dig'
	...st/bin/../games/minetest_game/mods/default/functions.lua:301: in function 'dig_up'
	...netest/bin/../games/minetest_game/mods/default/nodes.lua:1416: in function 'after_dig_node'
	...
	/opt/minetest/bin/../builtin/game/item.lua:547: in function 'node_dig'
	...st/bin/../games/minetest_game/mods/default/functions.lua:301: in function 'dig_up'
	...netest/bin/../games/minetest_game/mods/default/nodes.lua:1416: in function 'after_dig_node'
	/opt/minetest/bin/../builtin/game/item.lua:547: in function 'node_dig'
	...st/bin/../games/minetest_game/mods/default/functions.lua:301: in function 'dig_up'
	...netest/bin/../games/minetest_game/mods/default/nodes.lua:1416: in function 'after_dig_node'
	/opt/minetest/bin/../builtin/game/item.lua:547: in function 'node_dig'
	...st/bin/../games/minetest_game/mods/default/functions.lua:301: in function 'dig_up'
	...netest/bin/../games/minetest_game/mods/default/nodes.lua:1416: in function 'after_dig_node'
	/opt/minetest/bin/../builtin/game/item.lua:547: in function </opt/minetest/bin/../builtin/game/item.lua:460>
```

steps to reproduce: 
1. create a very tall column of papyrus (1024 wasn't enough, 4096 was, worldedit was helpful, but make sure the mapblocks are loaded)
2. dig the bottom node

this is relatively easy to trigger as an unprivileged player, particularly if ethereal is also installed (papyrus becomes a solid node, and you can nerd-pole on them). 

in my local repos, i see it used in cool_trees, ethereal, and naturalbiomes, so the functionality can't be changed without potentially breaking things.

i can't think of a "clean" fix. perhaps, add an additional 4th parameter to `default.dig_up`, indicating how many nodes it's supposed to dig up. i'll call this `n` for now for lack of a better name. have this default to something sane like 16 or 128. then, if the method is called and there's "air" for `n` nodes below the current node, don't continue to dig up - assume we've already dug up `n` nodes.

Emojigit:
Fixes #3075

This PR fixes the recursive calling of `default.dig_up` and adds a parameter for a maximum number of nodes searched.

This PR is ready for review.

### DIFF
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -2,149 +2,149 @@
 -- Sounds
 --
 
-function default.node_sound_defaults(table)
-	table = table or {}
-	table.footstep = table.footstep or
+function default.node_sound_defaults(tbl)
+	tbl = tbl or {}
+	tbl.footstep = tbl.footstep or
 			{name = "", gain = 1.0}
-	table.dug = table.dug or
+	tbl.dug = tbl.dug or
 			{name = "default_dug_node", gain = 0.25}
-	table.place = table.place or
+	tbl.place = tbl.place or
 			{name = "default_place_node_hard", gain = 1.0}
-	return table
+	return tbl
 end
 
-function default.node_sound_stone_defaults(table)
-	table = table or {}
-	table.footstep = table.footstep or
+function default.node_sound_stone_defaults(tbl)
+	tbl = tbl or {}
+	tbl.footstep = tbl.footstep or
 			{name = "default_hard_footstep", gain = 0.2}
-	table.dug = table.dug or
+	tbl.dug = tbl.dug or
 			{name = "default_hard_footstep", gain = 1.0}
-	default.node_sound_defaults(table)
-	return table
+	default.node_sound_defaults(tbl)
+	return tbl
 end
 
-function default.node_sound_dirt_defaults(table)
-	table = table or {}
-	table.footstep = table.footstep or
+function default.node_sound_dirt_defaults(tbl)
+	tbl = tbl or {}
+	tbl.footstep = tbl.footstep or
 			{name = "default_dirt_footstep", gain = 0.25}
-	table.dig = table.dig or
+	tbl.dig = tbl.dig or
 			{name = "default_dig_crumbly", gain = 0.4}
-	table.dug = table.dug or
+	tbl.dug = tbl.dug or
 			{name = "default_dirt_footstep", gain = 1.0}
-	table.place = table.place or
+	tbl.place = tbl.place or
 			{name = "default_place_node", gain = 1.0}
-	default.node_sound_defaults(table)
-	return table
+	default.node_sound_defaults(tbl)
+	return tbl
 end
 
-function default.node_sound_sand_defaults(table)
-	table = table or {}
-	table.footstep = table.footstep or
+function default.node_sound_sand_defaults(tbl)
+	tbl = tbl or {}
+	tbl.footstep = tbl.footstep or
 			{name = "default_sand_footstep", gain = 0.05}
-	table.dug = table.dug or
+	tbl.dug = tbl.dug or
 			{name = "default_sand_footstep", gain = 0.15}
-	table.place = table.place or
+	tbl.place = tbl.place or
 			{name = "default_place_node", gain = 1.0}
-	default.node_sound_defaults(table)
-	return table
+	default.node_sound_defaults(tbl)
+	return tbl
 end
 
-function default.node_sound_gravel_defaults(table)
-	table = table or {}
-	table.footstep = table.footstep or
+function default.node_sound_gravel_defaults(tbl)
+	tbl = tbl or {}
+	tbl.footstep = tbl.footstep or
 			{name = "default_gravel_footstep", gain = 0.25}
-	table.dig = table.dig or
+	tbl.dig = tbl.dig or
 			{name = "default_gravel_dig", gain = 0.35}
-	table.dug = table.dug or
+	tbl.dug = tbl.dug or
 			{name = "default_gravel_dug", gain = 1.0}
-	table.place = table.place or
+	tbl.place = tbl.place or
 			{name = "default_place_node", gain = 1.0}
-	default.node_sound_defaults(table)
-	return table
+	default.node_sound_defaults(tbl)
+	return tbl
 end
 
-function default.node_sound_wood_defaults(table)
-	table = table or {}
-	table.footstep = table.footstep or
+function default.node_sound_wood_defaults(tbl)
+	tbl = tbl or {}
+	tbl.footstep = tbl.footstep or
 			{name = "default_wood_footstep", gain = 0.15}
-	table.dig = table.dig or
+	tbl.dig = tbl.dig or
 			{name = "default_dig_choppy", gain = 0.4}
-	table.dug = table.dug or
+	tbl.dug = tbl.dug or
 			{name = "default_wood_footstep", gain = 1.0}
-	default.node_sound_defaults(table)
-	return table
+	default.node_sound_defaults(tbl)
+	return tbl
 end
 
-function default.node_sound_leaves_defaults(table)
-	table = table or {}
-	table.footstep = table.footstep or
+function default.node_sound_leaves_defaults(tbl)
+	tbl = tbl or {}
+	tbl.footstep = tbl.footstep or
 			{name = "default_grass_footstep", gain = 0.45}
-	table.dug = table.dug or
+	tbl.dug = tbl.dug or
 			{name = "default_grass_footstep", gain = 0.7}
-	table.place = table.place or
+	tbl.place = tbl.place or
 			{name = "default_place_node", gain = 1.0}
-	default.node_sound_defaults(table)
-	return table
+	default.node_sound_defaults(tbl)
+	return tbl
 end
 
-function default.node_sound_glass_defaults(table)
-	table = table or {}
-	table.footstep = table.footstep or
+function default.node_sound_glass_defaults(tbl)
+	tbl = tbl or {}
+	tbl.footstep = tbl.footstep or
 			{name = "default_glass_footstep", gain = 0.3}
-	table.dig = table.dig or
+	tbl.dig = tbl.dig or
 			{name = "default_glass_footstep", gain = 0.5}
-	table.dug = table.dug or
+	tbl.dug = tbl.dug or
 			{name = "default_break_glass", gain = 1.0}
-	default.node_sound_defaults(table)
-	return table
+	default.node_sound_defaults(tbl)
+	return tbl
 end
 
-function default.node_sound_ice_defaults(table)
-	table = table or {}
-	table.footstep = table.footstep or
+function default.node_sound_ice_defaults(tbl)
+	tbl = tbl or {}
+	tbl.footstep = tbl.footstep or
 			{name = "default_ice_footstep", gain = 0.15}
-	table.dig = table.dig or
+	tbl.dig = tbl.dig or
 			{name = "default_ice_dig", gain = 0.5}
-	table.dug = table.dug or
+	tbl.dug = tbl.dug or
 			{name = "default_ice_dug", gain = 0.5}
-	default.node_sound_defaults(table)
-	return table
+	default.node_sound_defaults(tbl)
+	return tbl
 end
 
-function default.node_sound_metal_defaults(table)
-	table = table or {}
-	table.footstep = table.footstep or
+function default.node_sound_metal_defaults(tbl)
+	tbl = tbl or {}
+	tbl.footstep = tbl.footstep or
 			{name = "default_metal_footstep", gain = 0.2}
-	table.dig = table.dig or
+	tbl.dig = tbl.dig or
 			{name = "default_dig_metal", gain = 0.5}
-	table.dug = table.dug or
+	tbl.dug = tbl.dug or
 			{name = "default_dug_metal", gain = 0.5}
-	table.place = table.place or
+	tbl.place = tbl.place or
 			{name = "default_place_node_metal", gain = 0.5}
-	default.node_sound_defaults(table)
-	return table
+	default.node_sound_defaults(tbl)
+	return tbl
 end
 
-function default.node_sound_water_defaults(table)
-	table = table or {}
-	table.footstep = table.footstep or
+function default.node_sound_water_defaults(tbl)
+	tbl = tbl or {}
+	tbl.footstep = tbl.footstep or
 			{name = "default_water_footstep", gain = 0.2}
-	default.node_sound_defaults(table)
-	return table
+	default.node_sound_defaults(tbl)
+	return tbl
 end
 
-function default.node_sound_snow_defaults(table)
-	table = table or {}
-	table.footstep = table.footstep or
+function default.node_sound_snow_defaults(tbl)
+	tbl = tbl or {}
+	tbl.footstep = tbl.footstep or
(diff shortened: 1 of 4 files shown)


mods/default/init.lua
```diff
@@ -17,7 +17,7 @@ default.get_translator = S
 -- and avoids obscure, hard to debug runtime errors.
 -- This section should be updated before release and older checks can be dropped
 -- when newer ones are introduced.
-if ItemStack("").add_wear_by_uses == nil then
+if not minetest.features.node_interaction_actor then
```
sfan5, line 20: I can see that it's used but would it stop working without this feature? I don't think so.

mods/default/functions.lua
```diff
@@ -293,12 +293,26 @@ minetest.register_abm({
...
+	if in_dig_up then return end -- Avoid excess calls
 	if digger == nil then return end
-	local np = {x = pos.x, y = pos.y + 1, z = pos.z}
-	local nn = minetest.get_node(np)
-	if nn.name == node.name then
-		minetest.node_dig(np, nn, digger)
+	max_height = max_height or mapgen_limit
+
+	for y = pos.y + 1, math.min(pos.y + max_height, mapgen_limit) do
+		local up_pos  = { x = pos.x, y = y, z = pos.z}
```
sfan5, line 304: `vector.new(pos.x, y, pos.z)`

mods/default/functions.lua
```diff
@@ -293,12 +293,26 @@ minetest.register_abm({
 -- Dig upwards
 --
 
-function default.dig_up(pos, node, digger)
+local in_dig_up = false
+local mapgen_limit = tonumber(minetest.settings:get("mapgen_limit")) or 31007
+function default.dig_up(pos, node, digger, max_height)
+	if in_dig_up then return end -- Avoid excess calls
```
sfan5, line 299: I think `-- Do not recurse` would be clearer

mods/default/functions.lua
```diff
@@ -293,12 +293,26 @@ minetest.register_abm({
...
 	if digger == nil then return end
-	local np = {x = pos.x, y = pos.y + 1, z = pos.z}
-	local nn = minetest.get_node(np)
-	if nn.name == node.name then
-		minetest.node_dig(np, nn, digger)
+	max_height = max_height or mapgen_limit
```
sfan5, line 301: IMO it would be totally fine to just use 100 here

mods/default/functions.lua
Emojigit: Using the vector object results in a performance tradeoff (negligible though), and I don't see the need of the helpers.

mods/default/init.lua
Emojigit: I have to know whether the node digging was successful. However, `minetest.node_dig` returns no value/does not have documentation on its return value.

mods/default/functions.lua
sfan5: while that's true this goes against explicit recommendation of the docs:
https://github.com/minetest/minetest/blob/master/doc/lua_api.md#compatibility-notes

mods/default/init.lua
sfan5: Ah so it's about the return value, right.

mods/default/functions.lua
Emojigit: Oh my. Gotta do it then.

Emojigit:
@sfan5 Change applied in 0e79158. Ready for review again.

mods/default/init.lua
```diff
@@ -17,7 +17,7 @@ default.get_translator = S
...
 	error("\nThis version of Minetest Game is incompatible with your engine version "..
```
sfan5, line 21: `min_minetest_version` in game.conf needs updating too

Emojigit: Done in 6d61b5e

sfan5 (approved):
works

mods/default/functions.lua
```diff
@@ -293,12 +293,25 @@ minetest.register_abm({
...
+			in_dig_up = true
+			if not minetest.dig_node(up_pos, digger) then
+				in_dig_up = false
+				break
+			end
+			in_dig_up = false
+		else
+			break
+		end
 	end
```
SmallJoker, line 315: Technically, it's not necessary to change the variable more than once:
```suggestion
	in_dig_up = true
	for y = pos.y + 1, pos.y + max_height do
		local up_pos  = vector.new(pos.x, y, pos.z)
		local up_node = minetest.get_node(up_pos)
		if up_node.name ~= node.name then
			break
		end
		if not minetest.dig_node(up_pos, digger) then
			break
		end
	end
	in_dig_up = false
```

Emojigit: I decided to duplicate my code because I want `in_dig_up` to be closer to the actual function, so if we are `pcall()`ed (weird error always happens), we are more likely not stuck. Sadly we don't have `finally` keywords like in Python.

mods/default/functions.lua
```diff
@@ -293,12 +293,25 @@ minetest.register_abm({
 -- Dig upwards
 --
 
-function default.dig_up(pos, node, digger)
+local in_dig_up = false
+function default.dig_up(pos, node, digger, max_height)
```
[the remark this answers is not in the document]
Emojigit: Solved in 0b006fc

The function is not found in `game_api.txt` before my PR. I added Lua language server annotations here.

mods/default/functions.lua
appgurueu: That sounds like you should use `pcall` (or preferably `xpcall` so you don't lose the stack trace) yourself, then reset the variable after that.

It should be possible to *guarantee* that you aren't stuck. Well, maybe modulo coroutines at least.

ps. I'm not very happy with this mutable global state variable, but I see no good alternative yet.

Emojigit: Fixed in 5ae2b04d729254bd33a3c9693a9c389f72652ba2 and f19fa6eb4710f25757098d6f8accdaac35eec5e1

mods/default/functions.lua
```diff
@@ -293,13 +293,41 @@ minetest.register_abm({
...
+			break
+		end
+		local noerr, success = xpcall(function()
+			minetest.dig_node(up_pos, digger)
+		end, function(...)
+			in_dig_up = false
+			minetest.log("error", "Error raised during `default.dig_up` call:")
+			for line in debug.traceback(...):gmatch("([^\n]*)\n?") do
+				minetest.log("error", line)
+			end
```
SmallJoker, line 322: As an alternative to manual parsing and outputting - might `minetest.error_handler(...)` produce better output?

mods/default/functions.lua
```diff
@@ -293,13 +293,41 @@ minetest.register_abm({
...
+		end)
```
sfan5, line 323: not sure if to call this clever or a hack but here's a cleaner suggestion:
```lua
local in_dig_up = {}
setmetatable(in_dig_up, {__mode = "k"})

function default.dig_up(pos, node, digger, max_height)
	local sentinel = {}
	if in_dig_up[sentinel] then return end
	-- ...
	in_dig_up[sentinel] = {}
	-- ..
	-- no resetting needed
end
```

SmallJoker: This is a beautiful hack.

mods/default/functions.lua
Emojigit: Weirdly, using `minetest.error_handler` does not print anything about the error... IDK why.

mods/default/functions.lua
Emojigit: Doing so would prevent running this code multiple times before garbage collection, e.g. running pipework node breakers all powered by the same mesecon line.

sfan5: It doesn't work if you use a fixed key. try it exactly like in my example

mods/default/functions.lua
```diff
@@ -293,13 +293,41 @@ minetest.register_abm({
...
+	max_height = max_height or 100
+
+	in_dig_up = true
+	for y = pos.y + 1, pos.y + max_height do
+		local up_pos  = vector.new(pos.x, y, pos.z)
+		local up_node = minetest.get_node(up_pos)
+		if up_node.name ~= node.name then
+			break
+		end
+		local noerr, success = xpcall(function()
```
sfan5, line 315: thinking about this again: why does it even matter that it's left in a broken state when an error happens?
i don't see any situation in which mods doing a try-catch with dig_node makes sense

SmallJoker: I don't think that whether it makes sense actually limits the creativity of mod authors. It's an edge-case for sure.

mods/default/functions.lua
appgurueu: `minetest.error_handler` should give you a formatted error string you can log, it doesn't do the printing itself.

mods/default/functions.lua
```diff
@@ -293,13 +293,41 @@ minetest.register_abm({
 -- Dig upwards
 --
 
-function default.dig_up(pos, node, digger)
+local in_dig_up = false
+
+---Find all nodes above this one that is the same, then dig them all
+---@param pos vector The position of the base node
```
appgurueu, line 299: (btw do we want to set a precedent of adopting LDoc?)

mods/default/functions.lua
appgurueu: Technically there are *still* ways to force a broken state if you want to, e.g. using coroutines. But I don't think it's very worthwhile to try to guard against this.

sfan5: additional idea: reset `in_dig_up` on every server step

this minimizes the effects of the already unlikely error edge case to basically nothing

mods/default/functions.lua
Emojigit: Done in b49153c

mods/default/functions.lua
Emojigit: Done in b49153c

mods/default/functions.lua
Emojigit: I added LDoc because this function did not exist in `game_api.txt`. Should I move them there?

mods/default/functions.lua
```diff
@@ -293,15 +293,46 @@ minetest.register_abm({
...
+	max_height = max_height or 100
+
+	in_dig_up = true
+	for y = pos.y + 1, pos.y + max_height do
+		local up_pos  = vector.new(pos.x, y, pos.z)
+		local up_node = minetest.get_node(up_pos)
+		if up_node.name ~= node.name then
+			break
+		end
+		local noerr, success = xpcall(function()
```
sfan5, line 315: my opinion remains to remove all pcall handling and rely on the globalstep cleaning up after the unlikely case of an error

mods/default/functions.lua
```diff
@@ -293,15 +293,46 @@ minetest.register_abm({
...
 	if digger == nil then return end
-	local np = {x = pos.x, y = pos.y + 1, z = pos.z}
-	local nn = minetest.get_node(np)
-	if nn.name == node.name then
-		minetest.node_dig(np, nn, digger)
+	max_height = max_height or 100
+
+	in_dig_up = true
+	for y = pos.y + 1, pos.y + max_height do
+		local up_pos  = vector.new(pos.x, y, pos.z)
```
sfan5, line 310:
```suggestion
	for y = 1, max_height do
		local up_pos  = vector.offset(pos, 0, y, 0)
```

Emojigit:
Hang on - this change would break any servers using fake players (e.g. pipeworks and digtron) to dig papyrus or any nodes using default.dig_up. Though this is not in our scope, I recommend putting this on hold considering how fakelib is widely used.

mods/default/functions.lua
Emojigit: Done in b5c297c

mods/default/init.lua
Emojigit: I reverted these changes as using that engine function breaks fakelib. `minetest.node_dig` does provide basic checks (though not documented), so I decided to keep using it.

mods/default/functions.lua
Emojigit: Done in 45ef266667e41a6f3b2085c89e1f7126363be1ab

mods/default/functions.lua
```diff
@@ -293,15 +293,31 @@ minetest.register_abm({
...
+			break
+		end
+		if not minetest.node_dig(up_pos, up_node, digger) then
+			break
+		end
 	end
+	in_dig_up = false
 end
 
+minetest.register_globalstep(function()
```
sfan5, line 317:
```suggestion
-- errors are hard to handle, instead we rely on resetting this value the next step
minetest.register_globalstep(function()
```

sfan5 (approved):
LGTM

PR #3135, opened 2024-06-23
Add new recipe for getting 4 strings from any wool block

M4urus:
It would make mods such as Animalia or any bow/xbow mod integrate much more naturally to MTG, as cotton is sometimes very hard to find or impossible depending on the biome.

Also this recipe makes a lot of sense IMO.


M4urus:
Skill issues.

PR #3136, opened 2024-06-23
Add new recipe for strings from any wool block

M4urus:
It would make mods such as Animalia(creating saddles require strings) or any bow/crossbow mod integrate much more naturally to MTG, since cotton is very hard to find, or impossible depending on the biome.

Also this recipe makes a lot of sense IMO.

### DIFF
--- a/mods/wool/init.lua
+++ b/mods/wool/init.lua
@@ -28,8 +28,8 @@ end
 
 -- Alternative recipe for strings.
 minetest.register_craft({
-		output="farming:string 4",
-		recipe={{"group:wool"}}
+	output = "farming:string 4",
+	recipe = {{"group:wool"}},
 })
 		
 -- Legacy
--- a/mods/wool/init.lua
+++ b/mods/wool/init.lua
@@ -26,6 +26,12 @@ for i = 1, #dyes do
 	}
 end
 
+-- Alternative recipe for strings.
+minetest.register_craft({
+		output="farming:string 4",
+		recipe={{"group:wool"}}
+})
+		
 -- Legacy
 -- Backwards compatibility with jordach's 16-color wool mod
 minetest.register_alias("wool:dark_blue", "wool:blue")


M4urus:
Sorry, this is my first contrib. I don't get the issue with the white space, could somebody help me with that? I've seen recent PRs with white spaces added and there was no warning in luacheck.

mods/wool/init.lua
```diff
@@ -26,6 +26,12 @@ for i = 1, #dyes do
 	}
 end
 
+-- Alternative recipe for strings.
+minetest.register_craft({
+		output="farming:string 4",
```
appgurueu, line 31: You're using two tabs here when you should use just one. Overall, you should try to "blend in" formatting-wise, for example adding spaces around `=`.

appgurueu (commented):
I'm not sure whether we can accept this. I'm afraid we might have to view this as a feature and reject it accordingly - it changes game mechanics, even if only slightly.

If this is unanimously considered a good idea, I'd be open to it however.

mods/wool/init.lua
M4urus: > You're using two tabs here when you should use just one. Overall, you should try to "blend in" formatting-wise, for example adding spaces around `=`.

I changed the code to fit the style. Thank you!

SmallJoker:
This change makes sense from a real-world standpoint. Spinning wool is a common practice. Optimally we'd require a tool to wear out (there's currently nothing that would match in MTG), and time if we wanted to follow reality. Strings might still be easier to obtain by farming cotton - depending on the automation mods that are available.

I wouldn't be opposed to such change. Perhaps adding a loss (3 strings instead of 4) could artificially make cotton more attractive - or mods that add appropriate tools including a wool spinning wheel for lossless conversion.

sfan5:
@M4urus do you have time to address the comments?

SmallJoker:
Neglected. No activity in over a year, thus closing.

PR #3137, opened 2024-06-25
New dirt texture

GNUAn:


tenplus1:
@GNUAn - Minetest game has been locked and will only be updated with bug fixes, texture replacements will be done using texture packs.

appgurueu:
This is likely to fall in the feature / controversial change category and to be rejected as such.

I also see no justification for why this new texture should be unanimously considered better than the old one.

PR #3142, opened 2024-07-15
Replace deprecated `get_metadata()` calls

Niklp09:
Trivial.


appgurueu (commented):
Looks sensible. Is it documented anywhere that key `""` gives you the legacy metadata? (If not, then we should fix that in the engine docs.)

Niklp09:
No, that‘s currently not documented.

sfan5 (approved):

appgurueu:
> No, that‘s currently not documented.

Alright, documented in https://github.com/minetest/minetest/commit/90fccc15ebf86b58656c28ed9853780d92feb69e

PR #3144, opened 2024-07-25
Limit furnace sound volume on catch-up smelting

Emojigit:
This PR limits the volume of smelting sound if the furnace smelts multiple items in the same timer step.

This PR is ready for review.

### DIFF
--- a/mods/default/furnace.lua
+++ b/mods/default/furnace.lua
@@ -134,6 +134,7 @@ local function furnace_node_timer(pos, elapsed)
 	local fuel
 
 	local update = true
+	local items_smelt = 0
 	while elapsed > 0 and update do
 		update = false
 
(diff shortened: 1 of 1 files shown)


mods/default/furnace.lua
```diff
@@ -224,6 +223,11 @@ local function furnace_node_timer(pos, elapsed)
 		elapsed = elapsed - el
 	end
 
+	if items_smelt ~= 0 then
```
appgurueu, line 226:
```suggestion
	if items_smelt > 0 then
```

(just a matter of taste, feel free to ignore)

mods/default/furnace.lua
```diff
@@ -224,6 +223,11 @@ local function furnace_node_timer(pos, elapsed)
...
+		-- Play cooling sound
+		minetest.sound_play("default_cool_lava",
+			{ pos = pos, max_hear_distance = 16, gain = 0.07 * math.min(items_smelt, 7) }, true)
```
appgurueu, line 229: Makes sense to not let it get arbitrarily loud; how did you choose this? I assume you roughly went for ~0.5 max gain?

appgurueu (approved):
Tested with the max. of 7, sounds bearable :+1:

mods/default/furnace.lua
Emojigit: At first, I went for 0.05, realizing that it works but was too quiet, so I increased that to 0.07.

Issue #3145, opened 2024-07-27
Bug: mod bones - server crash when use 'minetest.punch_node()' from any mod


berengma:
Version:

```
Minetest 5.8.0 (Linux)
Using LuaJIT 2.1.0-beta3
BUILD_TYPE=Release
RUN_IN_PLACE=0
USE_CURL=1
STATIC_SHAREDIR="/usr/local/share/minetest"
```

Error message from log:

```
2024-07-27 01:58:25: ERROR[Main]: ServerError: AsyncErr: Lua: Runtime error from mod 'bones' in callback luaentity_Step(): "Runtime error from mod 'bones' in callback node_on_punch(): ...l/share/minetest/games/minetest_game/mods/bones/init.lua:100: attempt to index local 'player_inv' (a ni
2024-07-27 01:58:25: ERROR[Main]: l value)
2024-07-27 01:58:25: ERROR[Main]: stack traceback:
2024-07-27 01:58:25: ERROR[Main]:       ...l/share/minetest/games/minetest_game/mods/bones/init.lua:100: in function <...l/share/minetest/games/minetest_game/mods/bones/init.lua:85>
2024-07-27 01:58:25: ERROR[Main]:       [C]: in function 'punch_node'
```

reproduction:

- just minetest game and a a mod using the 'minetest.punch_node(pos)' method
- player with something in inv dies
- use a mod with the described method and use it on the pos of the placed bones
- server crashes with above message

possible solution:
 
- In init.lua line 85 not only check for owner but maybe check also for player:sPlayer() ?

SmallJoker:
Reproduced by spawning new bones (do not place manually).

```Lua
minetest.register_chatcommand("p", {
	func = function(name, param)
		local pos = minetest.get_player_by_name(name):get_pos()

		local node_name = #param > 0 and param or "bones:bones"
		local np = minetest.find_node_near(pos, 4, node_name, true)
		if not np then
			return false, "No node found"
		end
		minetest.punch_node(np)
		return true, "OK!"
	end
})
```

Even though the Lua API states that `puncher` may be `nil`, but the C++ side creates an empty ObjectRef, which in turn does not have an inventory.

What is the expected result? Would you like to remove the node and destroy all of its contents, drop the contents as item entities, or disallow any interaction?

berengma:
Difficult to say what the expected result is. Depends on the mod which uses the method.
Disallowing any interaction is probably not what I wanted to have, when I wrote a bone harvester mod.
[...]

PR #3146, opened 2024-07-30
Fix crash if bones punched by non-player

berengma:
- Added check for player:is_player()
- if bones are fresh and the bones are punched nothing happens
- if bones are old, the inventory is dropped as entities and the bones node is removed

### DIFF
--- a/mods/bones/init.lua
+++ b/mods/bones/init.lua
@@ -27,7 +27,7 @@ local drop = function(pos, itemstack)
 	end
 end
 
-local dropContents = function(pos)
+local drop_contents = function(pos)
 	local inv = minetest.get_meta(pos):get_inventory()
 
 	for i = 1, inv:get_size("main") do
@@ -109,7 +109,8 @@ local bones_def = {
 		end
 
 		if not player:is_player() then
-			dropContents(pos)
+			drop_contents(pos)
+			return
 		end
 
 		if minetest.get_meta(pos):get_string("infotext") == "" then
--- a/mods/bones/init.lua
+++ b/mods/bones/init.lua
@@ -16,7 +16,7 @@ local function is_owner(pos, name)
 	return false
 end
 
-local drop = function(pos, itemstack)
+local function drop(pos, itemstack)
 	local obj = minetest.add_item(pos, itemstack:take_item(itemstack:get_count()))
 	if obj then
 		obj:set_velocity({
--- a/mods/bones/init.lua
+++ b/mods/bones/init.lua
@@ -27,7 +27,7 @@ local function drop(pos, itemstack)
 	end
 end
 
-local drop_contents = function(pos)
+local function drop_contents(pos)
 	local inv = minetest.get_meta(pos):get_inventory()
 
 	for i = 1, inv:get_size("main") do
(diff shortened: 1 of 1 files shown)


SmallJoker (commented):
Works. Tested with the code in https://github.com/minetest/minetest_game/issues/3145#issuecomment-2254163679.

mods/bones/init.lua
```diff
@@ -16,6 +16,27 @@ local function is_owner(pos, name)
 	return false
 end
 
+local drop = function(pos, itemstack)
```
appgurueu, line 19:
```suggestion
local function drop(pos, itemstack)
```

mods/bones/init.lua
```diff
@@ -16,6 +16,27 @@ local function is_owner(pos, name)
...
+	if obj then
+		obj:set_velocity({
+			x = math.random(-10, 10) / 9,
+			y = 5,
+			z = math.random(-10, 10) / 9,
+		})
+	end
+end
+
+local dropContents = function(pos)
```
appgurueu, line 30:
```suggestion
local function drop_contents(pos)
```

mods/bones/init.lua
```diff
@@ -87,6 +108,10 @@ local bones_def = {
 			return
 		end
 
+		if not player:is_player() then
+			dropContents(pos)
```
appgurueu, line 112: Shouldn't this be followed by a `return`?

mods/bones/init.lua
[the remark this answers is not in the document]
berengma: Good to know that there are code style guidelines and where. I changed it.

mods/bones/init.lua
berengma: I cannot find anything about this in the guidelines. Furthermore this function had been there before. :)

mods/bones/init.lua
berengma: see above

mods/bones/init.lua
berengma: return is added now.

berengma:
Updated PR according to code style guidelines

appgurueu (approved):
Tested using SmallJoker's code by setting the bone share time to a few seconds, then `/killme`, then `/p` (nothing happens, bones are still fresh), then after a few seconds `/p` again (items are dropped).

mods/bones/init.lua
appgurueu: Well the function right above it uses the `local function` syntactic sugar. (I also prefer this style and it costs nothing to apply it consistently as we go.)

berengma:
Anyone who could do a second review ?

berengma (commented):
from my point everything is resolved

SmallJoker (approved):

mods/bones/init.lua
```diff
@@ -16,6 +16,27 @@ local function is_owner(pos, name)
...
+	if obj then
+		obj:set_velocity({
+			x = math.random(-10, 10) / 9,
+			y = 5,
+			z = math.random(-10, 10) / 9,
+		})
+	end
+end
+
+local drop_contents = function(pos)
```
sfan5, line 30:
```suggestion
local function drop_contents(pos)
```

PR #3153, opened 2024-08-31
Bring back old cracky sounds

### LINKED ISSUE #3015: Bring back the classic wood and stone sounds.
The new sounds are particular ear grading due to the Minetest sound system making sounds a bit too compressed sounding. The original sounds didn't have this issue due to them not being that crisp. Not to mention I just miss the original sounds. :(

cx384:
This closes #3015. (The choppy sound does not get replaced, since it is not as bad as the cracky sound, but I could also bring it back if someone has concerns.)
The cracky footstep sound stays the same, so the footstep and dug sound is no longer the same.

(I really don't like the new cracky dig sound, it is way too rough and hurts my ears, and even if you don't use a metal pickaxe it sounds metallic.)

**Hearing tests:**
[Current Version](https://whyp.it/tracks/embed?id=202473)
[With this PR](https://whyp.it/tracks/embed?id=202474)

I made a poll, to check what other MTG users think about this.
https://forum.minetest.net/viewtopic.php?t=30943


SmallJoker (approved):
Approval based on the forum feedback and my preference.

PR #3157, opened 2024-09-17
Move dungeon room scanning to mapgen thread

sfan5:
it isn't a bottleneck or anything, but shiny new stuff


sfan5:
on the other hand maybe it's better to [KISS](https://en.wikipedia.org/wiki/KISS_principle)

PR #3163, opened 2024-10-14
Set bloom intensity

grorp:
Since https://github.com/minetest/minetest/pull/15231 has been merged, it's now MTG's responsibility as a game to set a value here.

> The default value is to be changed from 0.05 to 0 in the future. If you wish to keep the current default value, you should set it explicitly.
> ~ [lua_api.md](https://github.com/minetest/minetest/blob/7435ea0d4eae807bfd88e27fbee4521e336af367/doc/lua_api.md?plain=1#L8793-L8795)

Note that bloom intensity > 0 is also needed for volumetric lighting to work.

Previously: #3158

### DIFF
--- a/game_api.txt
+++ b/game_api.txt
@@ -1205,7 +1205,7 @@ Weather API
 -----------
 
 The weather mod will constantly adjust weather effects seen by the player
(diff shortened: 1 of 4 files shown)


sfan5 (approved):
lgtm

PR #3168, opened 2024-11-15
[no sq] Kick players from beds that have ceased existing

### LINKED ISSUE #3167: Player floats after bed burns down
If the bed a player is sleeping in is burned down, the player keeps floating in the air.

The player should instead die, or stand up because of the horrible pain.

## How to reproduce

* *Explicitly* set the `enable_fire` setting to true.  (And make sure you don't have `disable_fire` in the settings.)
* Start a multiplayer server, and join with 2 players. (Needed to avoid skipping night.)
* Lay down in a bed.
* Try to burn the player alive.

appgurueu:
Fixes #3167. Done via a globalstep, which is not a concern. I believe this to be the most straightforward approach; it is unrealistic to efficiently hook all possible ways the node could be removed.

### DIFF
--- a/mods/beds/functions.lua
+++ b/mods/beds/functions.lua
@@ -170,20 +170,20 @@ end
 
 -- Public functions
 
-function beds.kick_player(player)
+function beds.kick(player)
 	lay_down(player, nil, nil, false)
 end
 
 function beds.kick_players()
 	for name in pairs(beds.player) do
-		beds.kick_players(core.get_player_by_name(name))
+		beds.kick(core.get_player_by_name(name))
 	end
 end
 
 core.register_globalstep(function()
 	for name, bed_pos in pairs(beds.bed_position) do
 		if not beds.is_bed_node[core.get_node(bed_pos).name] then
-			beds.kick_player(core.get_player_by_name(name))
+			beds.kick(core.get_player_by_name(name))
 		end
 	end
 end)
--- a/mods/beds/functions.lua
+++ b/mods/beds/functions.lua
@@ -170,12 +170,23 @@ end
 
 -- Public functions
 
+function beds.kick_player(player)
+	lay_down(player, nil, nil, false)
+end
+
 function beds.kick_players()
-	for name, _ in pairs(beds.player) do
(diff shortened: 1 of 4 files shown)


sfan5:
I have no issues with this approach but shouldn't `on_destruct` be enough actually? (except vmanips)

mods/beds/functions.lua
```diff
@@ -170,13 +170,24 @@ end
...
+function beds.kick_player(player)
+	lay_down(player, nil, nil, false)
+end
+
 function beds.kick_players()
-	for name, _ in pairs(beds.player) do
-		local player = minetest.get_player_by_name(name)
-		lay_down(player, nil, nil, false)
+	for name in pairs(beds.player) do
+		beds.kick_players(core.get_player_by_name(name))
```
SmallJoker, line 179: Infinite recursion. Did you probably want this?

```suggestion
		beds.kick_player(core.get_player_by_name(name))
```

That aside, I don't think that a separate helper function is needed. If other mods need it, then it should be documented.

appgurueu: I think introducing a `kick_player` function, be it exposed or not, is a good idea. Duplicating the `lay_down(player, nil, nil, false)` call (which tbh should probably have been refactored to begin with, but that's another issue) is overall less readable.

I think I'll call it just `kick(player)` to make it a bit more concise and avoid the similarity with `kick_players`.

appgurueu:
> I have no issues with this approach but shouldn't `on_destruct` be enough actually? (except vmanips)

I think so, yes. VoxelManips might appear naturally in gameplay though, e.g. explosions - you still want the player to be kicked from bed when it was blown up (we could now start with `on_blast` etc, but that puts us back in the catch all situation).

SmallJoker:
The sleep formspec does neither get closed nor updated on kick. Is that intentional? Granted, it does precisely fix the linked issue.

grafik

SmallJoker:
Furthermore, what is the purpose of `beds.is_bed_node`, or the `bed = 1` node group, respectively? Either of them seems superfluous.

PR #3169, opened 2024-11-28
Rename Minetest to Luanti

### LINKED ISSUE #3164: Rename Minetest to Luanti in files
Rename Minetest to Luanti in the files.

Wuzzy2:
This PR changes the name "Minetest" to "Luanti" in all files. Fixes #3164.

It also changes the links from minetest.net to luanti.org domains (all tested).

Finally, the `game.conf` author is changed from "Minetest" to "Luanti". This is done in a separate commit because I'm not sure if doing so is safe. If this is a problem, just drop that commit.

### DIFF
--- a/game.conf
+++ b/game.conf
@@ -1,4 +1,3 @@
 title       = Minetest Game
-author      = Minetest
 description = A basic exploration, mining, crafting, and building, sandbox game with no NPCs, monsters, or animals. Minetest Game is usually used with mods added, and many mods are available for this game. Reliably maintained by Luanti core developers.
 min_minetest_version = 5.8
--- a/game.conf
+++ b/game.conf
@@ -1,4 +1,4 @@
 title       = Minetest Game
 author      = Minetest
-description = A basic exploration, mining, crafting, and building, sandbox game with no NPCs, monsters, or animals. Minetest Game is usually used with mods added, and many mods are available for this game. Reliably maintained by Minetest Engine core developers.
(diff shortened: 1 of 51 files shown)


rollerozxa:
The author key is basically only used by ContentDB and doesn't really do anything without the release key right now. It's safe to change but will be overridden to "Minetest" when installing through the client as the MTG package on CDB is minetest/minetest_game.

game.conf
```diff
@@ -1,4 +1,4 @@
 title       = Minetest Game
-author      = Minetest
-description = A basic exploration, mining, crafting, and building, sandbox game with no NPCs, monsters, or animals. Minetest Game is usually used with mods added, and many mods are available for this game. Reliably maintained by Minetest Engine core developers.
+author      = Luanti
```
sfan5, line 2: yeah let's just remove it if it's useless anyway

mods/beds/README.txt
```diff
@@ -5,7 +5,7 @@ See license.txt for license information.
 Authors of source code
 ----------------------
 Originally by BlockMen (MIT)
-Various Minetest developers and contributors (MIT)
+Various Luanti developers and contributors (MIT)
```
sfan5, line 8: strictly speaking they were contributors to "Minetest" at the time
or more specifically "Minetest Game"?

Wuzzy2:
Author removed from `game.conf`.

About the credits line: I thought the consensus was the rename applied retroactively, was it not?
Anyway, I want to hear other core dev opinions regarding the credits line, just to be sure.

Wuzzy2:
Alright, I decided to replace the line in the credits to "Minetest Game developers".

PR #3170, opened 2024-12-12
Tweak beds moving check

tenplus1:
checking player movement is too sensative with 0.001 as the value, changing to 0.05 works a lot better especially when 3rd party mods that change player physics is in use.


sfan5 (approved):
sounds good

appgurueu (approved):
looks like a more sensible threshold

PR #3173, opened 2025-01-05
Beds: allow digging stray top nodes

### LINKED ISSUE #3171: Different param2 of the bed bottom the top causes crash on dig.
If you change the param2 of the base of the bed to be very or slightly different from the top, maybe param2 tool or some glitch, when dug or removed in any way(tnt, worldedit remover) it causes a crash where "b" is a nil value.

SmallJoker:
Stray nodes (or half beds) may occur by ...
1. rotating the bottom bed node without calling the 'on_rotate' callback for various reasons
2. and digging one of the two nodes.
In master, the top node cannot be dug and is left in a bugged state. You cannot place another bed on top to fix it. The placement aborts. Hence this PR allows digging the top node as well without causing item duplication or item loss.

This PR might or might not fix #3171 (unconfirmed bug)

## How to test

Test 1:
1. Check whether normal bed placement and digging works.

Test 2:
1. Rotate the bed (bottom) node using any manipulation tool. I used the technic sonic screwdriver for this.
4. Dig the bottom node. -> You must receive a bed drop
5. Dig the top node. -> There must not be a drop

Test 3:
1. Rotate the bed (bottom) node
2. Dig the top node. -> There must not be a drop
4. Dig the bottom node. -> You must receive a bed drop

Test 4:
1. Rotate the nodes in any way
2. No errors or anomalies must occur.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -7,6 +7,7 @@ globals = {
 
 read_globals = {
 	"DIR_DELIM",
+	"core",
 	"minetest",
 	"dump",
 	"vector",
--- a/mods/beds/api.lua
+++ b/mods/beds/api.lua
@@ -10,6 +10,9 @@ end
 local function get_other_bed_pos(pos, n)
 	local node = core.get_node(pos)
 	local dir = core.facedir_to_dir(node.param2)
+	if not dir then
+		return -- There are 255 possible param2 values. Ignore bad ones.
+	end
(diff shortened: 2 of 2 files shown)


SmallJoker:
Fixes #3171 @Niklp09

sfan5 (approved):
tested

PR #3174, opened 2025-01-05
Fix optimize_textures.sh

EGYT5453:
I don't know when it stopped working, but now it gives the following error:

> find: warning: ‘-name’ matches against basenames only, but the given pattern contains a directory separator (‘/’), thus the expression will evaluate to false all the time.  Did you mean ‘-wholename’?

Changing it from `find -name '../*.png'` to `find .. -name '*.png'` fixed it for me.

### DIFF
--- a/utils/optimize_textures.sh
+++ b/utils/optimize_textures.sh
@@ -3,4 +3,4 @@
 # Colors with 0 alpha need to be preserved, because opaque leaves ignore alpha.
 # For that purpose, the use of indexed colors is disabled (-nc).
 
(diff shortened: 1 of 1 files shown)


sfan5 (approved):

SmallJoker:
I suppose this script was intended to be run from the git root like so:
```sh
bash util/optimize_textures.sh
```

A proper solution would be to use 
```sh
SCRIPT_DIR=`realpath "$(dirname "$0")"`
find "$SCRIPT_DIR" -name '*.png' -print0 | xargs -0 optipng -o7 -zm1-9 -nc -strip all -clobber
```

PR #3177, opened 2025-01-12
Add safety checks before overriding builtin item entity

sfan5:
it seems unlikely that the engine would change the methods in major ways because we're not the only game that does this, but if it happens we can at least not crash


appgurueu (approved):
sure

PR #3178, opened 2025-01-13
Add Bulgarian translation

StoyanDimitrov:


SmallJoker (approved):
The @-placeholders appear correctly. LGTM.

appgurueu (approved):
Looks legit in Google Translate. Something I noticed: It seems you translated "jungle tree" with "baobab"? That feels a bit too specific, but I don't mind much. What seems a bit problematic however is that baobabs aren't really jungle trees: They occur in "dry deciduous forests" (Wikipedia).

StoyanDimitrov:
Nice catch. I'll try to find a more appropriate alternative. In our culture, there really is no concept for a jungle tree. I could translate it directly as 'tree from the jungle,' but then the resources with that name would become long and clumsy, like 'plank from a tree from the jungle' (imagine that being 30-50% longer). I’ll need to dig around in the (translatable) jungle flora.

PR #3180, opened 2025-01-18
Fix TNT explosions not removing metadata

### LINKED ISSUE #3179: explosions leave metadata behind
explosions remove all nodes without an `on_blast`-callback in VoxelManip. This leaves their metadata behind. 
I would expect they are removed. It's not good to have metadata in positions where nobody expects it...

### How to reproduce:
- place a `default:sign_wall_wood` and write some text on it
- create an explosion that will destroy the sign
- the flame at the sign position will now have an infotext and a formspec

appgurueu:
Fixes #3179. Check that:

1. Metadata of nodes that are blasted is correctly removed. Easiest way to test that is via the sign as described in the issue.
2. Metadata of nodes that aren't blasted is preserved. For example put something in a locked chest and check that it is still in there after the explosion.

### DIFF
--- a/mods/tnt/init.lua
+++ b/mods/tnt/init.lua
@@ -347,9 +347,8 @@ local function tnt_explode(pos, radius, ignore_protection, ignore_on_blast, owne
 
 	-- Used to efficiently remove metadata of nodes that were destroyed.
 	-- Metadata is probably sparse, so this may save us some work.
-	local poses_with_meta = minetest.find_nodes_with_meta(p1, p2)
 	local has_meta = {}
-	for _, p in ipairs(poses_with_meta) do
+	for _, p in ipairs(minetest.find_nodes_with_meta(p1, p2)) do
 		has_meta[a:indexp(p)] = true
 	end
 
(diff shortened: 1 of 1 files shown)


mods/tnt/init.lua
```diff
@@ -345,6 +345,14 @@ local function tnt_explode(pos, radius, ignore_protection, ignore_on_blast, owne
...
 	basic_flame_on_construct = minetest.registered_nodes["fire:basic_flame"].on_construct
 
+	-- Used to efficiently remove metadata of nodes that were destroyed.
+	-- Metadata is probably sparse, so this may save us some work.
+	local poses_with_meta = minetest.find_nodes_with_meta(p1, p2)
+	local has_meta = {}
+	for _, p in ipairs(poses_with_meta) do
+		has_meta[a:indexp(p)] = true
+	end
+
```
sfan5, line 355:
```suggestion
	local has_meta = {}
	for _, p in ipairs(minetest.find_nodes_with_meta(p1, p2)) do
		has_meta[a:indexp(p)] = true
	end

```

appgurueu: sure, why not. though i don't think it matters much.

sfan5 (approved):
lgmt, not tested

PR #3194, opened 2025-05-26
Workaround for sign regression with 5.12

ZenonSeth:
Issue is that when the build in "Proceed" button is pressed, on_receive_fields is called twice - once with a table of `{ quit = true, text = "user text" }` and then immediately after with a fields table of just `{text = "ExitButton"}`

Fix for https://github.com/luanti-org/luanti/issues/16187


ZenonSeth:
Closing since it's deemed an engine bug

sfan5 (approved):

PR #3196, opened 2025-05-30
Use new VoxelManip cleanup API

sfan5:
note:
* carts only used vmanips to load an area, this was replaced with `core.load_area`
* `vm:update_map()` is a no-op since at least 5.0.0, so was removed


appgurueu (approved):
The changes all look unproblematic to me.

We could abstract this pattern: https://github.com/sfan5/minetest_game/compare/newapi...appgurueu:minetest_game:newapi2?expand=1 (there is a slight difference in that this always updates liquids, but that is probably for the better)

sfan5:
Feel free to merge your changes after this.

PR #3197, opened 2025-06-01
Refactor: Introduce `default.do_with_area`

appgurueu:
Abstracts the typical VoxelManip usage pattern in MTG.

A minor difference is that now `vm:update_liquids()` is always called, but this should be unproblematic if not more correct than what was done previously.

### DIFF
--- a/mods/default/functions.lua
+++ b/mods/default/functions.lua
@@ -835,3 +835,17 @@ function default.can_interact_with_node(player, pos)
 
 	return false
 end
+
+function default.do_with_area(p1, p2, func)
+	local vm = VoxelManip()
+	local minp, maxp = vm:read_from_map(p1, p2)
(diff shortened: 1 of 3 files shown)


appgurueu:
> tbh I would much rather have this as a builtin solution (`core.do_with_area`?)

the problem is that an engine feature needs to be more general (e.g. what about param1, param2? MTG currently happens not to touch these (it probably should though)) and more well-engineered than an internal MTG helper which we are free to change (note that i did not document it).

> `default` namespace litter

well, the thing is `default` is simply MTG's kitchen sink utility mod, i don't plan to change that here (which would also introduce another dependency for the MTG mods that need it)

SmallJoker (commented):
Looks good overall. Unfortunately saving a few LoC means adding another indent level for each (already) deeply nested loop. It's not beautiful but should scale better in the long-run by avoiding the boilerplate code.

PR #3199, opened 2025-06-06
BUGFIX: Fix itemname usage for mods and submods using same namespace

sem-geologist:
A narrative.
There are plenty mods which are modular, and thanks to `minetest.register_craftitem` and its enforced check if namespace of itemname is the same as mod, often in those cases a colon is prepended to itemname. 
i.e. 
mobs  - mobs API, main lib
mobs_animals - submod defining some animals (i.e. cow, which gives milk)
when doing `regiter_craftitem` code use semicolon to put everything in short named `mobs` namespace, i.e.:
`minetest.register_craftitem(":mobs:milk_bucket")".

When trying to define liquid with bucket lib like this in submod `mobs_animals`:
```lua
bucket.register_liquid(
	"mobs:milk_source",
	"mobs:milk_flowing",
	"mobs:milk_bucket",
	"mobs_milk_bucket.png",
	S("Milk Bucket"),
	{tool = 1}
)
```
The game would crash with error, as itemname is `mobs` instead of `mobs_animals`.

Adding colon like this:
```lua
bucket.register_liquid(
	"mobs:milk_source",
	"mobs:milk_flowing",
	":mobs:milk_bucket",
	"mobs_milk_bucket.png",
	S("Milk Bucket"),
	{tool = 1}
)
```
would not throw error.
It also would allow spilling milk on the ground. However when collecting milk back to bucket, it would try creating `:mobs:milk_bucket` instead of `mobs:milk_bucket`, thus would end up with creating unknown item.

And that is what this fix tries to address, so that liquids defined with prefixed colon in `itemname` could be correctly spilled and collected back to the bucket.

### DIFF
--- a/mods/bucket/init.lua
+++ b/mods/bucket/init.lua
@@ -52,7 +52,7 @@ end
 function bucket.register_liquid(source, flowing, itemname, inventory_image, name,
 		groups, force_renew)
 	local itemname_raw = itemname
-	itemname = itemname:match(":(.+)") or itemname
+	itemname = itemname and itemname:match(":(.+)") or itemname
 	bucket.liquids[source] = {
 		source = source,
 		flowing = flowing,


appgurueu (commented):
I think this patch is a bit too narrow. If I'm not missing anything, the *only* place where we want to keep the colon is for item registration. Anything afterwards should use the colon-less itemname. The code should thus strip the colon as soon as possible, to ensure that there can be no other occurrences later on where stripping the colon is forgotten.

sem-geologist:
oh, You are right, I am changing it to strip the colon from stored string as soon after register_craftitem body finishes.

mods/bucket/init.lua
```diff
@@ -116,6 +116,10 @@ function bucket.register_liquid(source, flowing, itemname, inventory_image, name
 				return ItemStack("bucket:bucket_empty")
 			end
 		})
+		if itemname:sub(1,1) == ":" then
+			-- If the itemname starts with a colon, correct that
+			bucket.liquids[source]["itemname"] = itemname:sub(2)
```
SmallJoker, line 121: Why is it necessary to overwrite this value? Couldn't you assign the correct value directly in L57?

e.g. `itemname = [itemname](itemname:sub(1,1) == ":" and itemname:sub(2) or itemname` (or perhaps a less dense version)

sem-geologist: Ok, I did as recommended.

appgurueu (approved):
thanks

SmallJoker (approved):
LGTM

sfan5:
I have reverted this due to #3201 and I couldn't immediately spot what was wrong. Please re-do the fix.

PR #3200, opened 2025-06-09
Beds: Add support for pova and player_monoids

SmallJoker:
'ObjectRef:set_physics_override' is prone to cause conflicts between several mods if they override the same fields. This change implements support for 'pova' and recent versions of player_monoids for improved compatibility.

Fixes https://github.com/minetest-mods/stamina/issues/54 (likely among others).

**How to test**

1. Install [pova](https://codeberg.org/tenplus1/pova) and [player_monoids](https://github.com/minetest-mods/player_monoids/)
2. Enable exactly one (or none) of the two mods
3. Join the world
4. `/time 0`
5. Right-click on a bed. You must no longer be able to move
6. Wait for dawn
7. You must be able to move
8. Repeat with no mods enabled and the other supported mod.

### DIFF
--- a/mods/beds/functions.lua
+++ b/mods/beds/functions.lua
@@ -6,8 +6,8 @@ if enable_respawn == nil then
 end
 
 -- Physics override management mods (shadow the global variable)
-local player_monoids = core.get_modpath("player_monoids") and player_monoids
-local pova           = core.get_modpath("pova")           and pova
+local player_monoids = core.get_modpath("player_monoids") and _G["player_monoids"]
+local pova           = core.get_modpath("pova")           and _G["pova"]
 
 if player_monoids and not player_monoids.speed.checkout_branch then
 	-- This function exists since 2025-05-17
--- a/mods/beds/functions.lua
+++ b/mods/beds/functions.lua
@@ -6,8 +6,8 @@ if enable_respawn == nil then
 end
 
 -- Physics override management mods (shadow the global variable)
-local player_monoids = core.get_modpath("player_monoids") and _G["player_monoids"]
-local pova           = core.get_modpath("pova")           and _G["pova"]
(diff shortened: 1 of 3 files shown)


mods/beds/functions.lua
```diff
@@ -5,6 +5,16 @@ if enable_respawn == nil then
 	enable_respawn = true
 end
 
+-- Physics override management mods (shadow the global variable)
+local player_monoids = core.get_modpath("player_monoids") and _G["player_monoids"]
```
sfan5, line 9: using `core.global_exists` would be nice and I wonder if we should be checking both the mod existence and global? just the latter would be enough.

mods/beds/functions.lua
```diff
@@ -50,6 +60,51 @@ local function check_in_beds(players)
...
+				pdata["monoid_branch_" .. k] = monoid:get_active_branch(player)
+				-- Change the "context" of the physics overrides
+				local branch = monoid:checkout_branch(player, IDENTIFIER)
+				branch:add_change(player, v)
+			end
+		elseif pova then
+			pova.add_override(name, "force", OVERRIDES)
+			pova.do_override(player)
+		else
+			-- Fall back to default. May conflict with other mods.
```
sfan5, line 83:
```suggestion
			-- Directly use engine API. May conflict with other mods.
```

sfan5 (commented):
code review only

mods/beds/functions.lua
sfan5: `local a = core.global_exists("a") and _G.a` sounds ok

SmallJoker: This does not check against global namespace poisoning, but that shouldn't happen anyway. At the end of the day there's many solutions and I don't think it really matters at all.

PR #3209, opened 2025-09-04
Translation into Danish

jeppebundsgaard:
I have translated all mods in minetest_game into Danish.


mods/sfinv/locale/sfinv.da.tr
```diff
@@ -0,0 +1,2 @@
+# textdomain: sfinv
+Crafting=
```
appgurueu, line 2: Intentional or just forgotten? If intentional, maybe add a comment that this should not be translated?

appgurueu (approved):
I don't know danish, but this feels like a solid translation (using online tools to assess it).

Ideally we'd find a danish speaker to review this, but for the time being we can just put it in; minor improvements can be made in later PRs.

PR #3211, opened 2025-09-15
Translation of the game description

alrito28:
I replicated what the Glitch does in the game, and it works.


mods/minetest_game_meta/locale/minetest_game_meta.fr.tr
```diff
@@ -0,0 +1,3 @@
+# textdomain: minetest_game_meta
+Minetest Game=Minetest Game
+A basic exploration, mining, crafting, and building, sandbox game with no NPCs, monsters, or animals. Minetest Game is usually used with mods added, and many mods are available for this game. Reliably maintained by Luanti core developers.=Un jeu bac à sable basique axé sur l'exploration, l'exploitation minière, l'artisanat et la construction, sans PNJ, monstres ni animaux. Minetest Game est généralement utilisé avec des mods supplémentaires, et de nombreux mods sont disponibles pour ce jeu. Entretien fiable assuré par les développeurs principaux de Luanti.
```
sfan5, line 3: Can't we just move this into `default`?

sfan5:
I am requesting this to be changed *before* we can merge the PR.

sfan5:
> Oh, okay, in mods/default/locale/default.fr.tr ?

No, that would get overwritten when automatically updating the files. The description should use a different textdomain.

alrito28:
What are you waiting for to accept my work?

What works, works, so go ahead.

SmallJoker:
The idea of translating is good. It's tricky to say how this should be implemented. I can see the following options: (preferred on top)

1. Extend the Luanti API to allow translating user-facing meta information (game and modpack descriptions, help dialogues or whatever might be added in the future). The locale would have to be put in a specific directory, however that would need blacklisting in the mod list, as we would not want to attempt to load a mod called `locale`.
2. I assume that each game (or modpack for this matter) has a central/core mod, hence the translation could be put there but using another textdomain. However, it would be an unreasonable effort for the translation script to find out which mod that would be.

I'd also appreciate more ideas on this topic.

sfan5:
Wouldn't just putting it into `mods/default/locale/game_description.fr.tr` work?
Then it's not an extra mod but the translation script would ignore it.

alrito28:
It's simple, it works for me. I based this on Glitch. Here is the link to the Glitch sources: https://codeberg.org/Wuzzy/Glitch

PR #3221, opened 2026-01-03
Fix confusing default for enable_fire and enable_tnt

### LINKED ISSUE #2678: minetest.conf.example and settingtypes do not explain that fire is disabled by default in multiplayer
Just like `enable_tnt` fire is disabled by default in multiplayer, but the documentation makes no mention of that.

### LINKED ISSUE #3220: State of `enable_fire` setting is confusing
As the title suggests. I have looked at settings and can not find anything on fire spread. 

Thank you in advance.

sfan5:
fixes #2678, fixes #3220

### DIFF
--- a/minetest.conf
+++ b/minetest.conf
@@ -1,0 +1,1 @@
+# empty, Minetest Game does not need any defaults.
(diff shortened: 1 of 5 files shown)


minetest.conf
```diff
@@ -0,0 +1 @@
+# empty, Minetest Game does not need any defaults.
```
sfan5: (we could also just delete this file)

SmallJoker (approved):
LGTM.

PR #3225, opened 2026-01-09
Fix player not kicked out of bed on destruct

### LINKED ISSUE #3167: Player floats after bed burns down

Wuzzy2:
Fixes #3167.

This guarantees any player in a given bed is kicked out of it when it is destroyed, using the `on_destruct` callback.

This PR has also been tested with the bed being destroyed by fire.

Competing PR: #3168

### DIFF
--- a/game_api.txt
+++ b/game_api.txt
@@ -49,6 +49,7 @@ Beds API
  * `beds.can_dig(bed_pos)` Returns a boolean whether the bed at `bed_pos` may be dug
  * `beds.read_spawns() `   Returns a table containing players respawn positions
  * `beds.kick_players()`  Forces all players to leave bed
+ * `beds.kick_player_at(pos)`  Forces player out of bed at pos (returns true on success, false otherwise)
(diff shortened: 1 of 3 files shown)


sfan5:
Will this update the sleep formspec for the affected player?
https://github.com/luanti-org/minetest_game/pull/3168#issuecomment-2491942343

Wuzzy2:
Ummm …
Well, NOW it does. :grin: 

Btw, for testing you might want to reduce the fire interval and disable the night skip so its easier to trigger the event of the bed being destroyed while you're in it.

SmallJoker (approved):
master:
grafik


PR: (the formspec is closed too)
grafik

PR #3229, opened 2026-01-19
Furnace: Check if item can be cooked before allowing it to be put int…

jordan4ibanez:
…o src


sfan5:
I think the ability to store random items in a furnace could be considered a "feature" that we don't want to break in the maintenance mode MTG is in.

jordan4ibanez:
> I think the ability to store random items in a furnace could be considered a "feature" that we don't want to break in the maintenance mode MTG is in.

I understand sir baguette

PR #3230, opened 2026-01-21
Fix description of overridden `give` privilege

### LINKED ISSUE #3224: Help text for `give` priv is wrong in Luanti 5.15 onwards
In MTG, the `pulverize` and `clearinv` commands do not need the `give` privilege. This was done to return to the behavior of pre-5.15 versions of Luanti. See `mods/default/chat.lua`.

However, this now means the help text of the `give` privilege will be wrong in Luanti 5.15:

> Can use /give, /giveme, /pulverize and /clearinv

The previous help text was:

> Can use /give and /giveme

I guess MTG needs to overwrite the help text of the `give` priv to return to the old help text.

Wuzzy2:
Fixes #3224.

This also imports the translations of the old `give` priv description from Luanti pre-5.15.

##### How to test

1. Start game in Luanti 5.15
2. Enter `/help privs`
3. Read the description of the `give` privilege

It should only mention `/give` and `/giveme` commands but no other command.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -2,7 +2,7 @@ unused_args = false
 allow_defined_top = true
 
 globals = {
-	"default"
+	"default",
 }
 
 read_globals = {
@@ -33,3 +33,6 @@ files["mods/player_api/api.lua"].globals = { "minetest" }
 
 -- Don't report on legacy definitions of globals.
 files["mods/default/legacy.lua"].global = false
+
+-- Overwrites a description of a privilege
+files["mods/default/chat.lua"].globals = { "core.registered_privileges" }
--- a/mods/default/chat.lua
+++ b/mods/default/chat.lua
@@ -1,6 +1,5 @@
 -- mods/default/chat.lua
 
--- support for MT game translation.
 local S = default.get_translator
 
 local function match_old(privs)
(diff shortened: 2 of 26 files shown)


Wuzzy2:
`luacheck` is being annoying:

> `mods/default/chat.lua:35:2: setting read-only field registered_privileges.give.description of global core`

This was intentional. I don't know of any other way on how to override the privilege description atm, as long we don't have something like `core.override_privilege`.

sfan5:
You just need to allow it in luacheckrc.

mods/default/chat.lua
```diff
@@ -1,5 +1,8 @@
 -- mods/default/chat.lua
 
+-- support for MT game translation.
```
sfan5:
```suggestion
```

mods/default/chat.lua
```diff
@@ -26,3 +29,8 @@ for _, cmd in ipairs({"pulverize", "clearinv"}) do
 		end
 	end
 end
+
+-- Revert description of 'give' privilege to what it was in Luanti 5.14
+if core.registered_privileges["give"] then
```
sfan5: this should only happen if the above code actually overwrote the privs

Wuzzy2:
This PR should be fine now.

.luacheckrc
```diff
@@ -2,7 +2,8 @@ unused_args = false
 allow_defined_top = true
 
 globals = {
-	"default"
+	"default",
+	"core.registered_privileges"
```
sfan5: would be better to allow this only for the relevant file

sfan5 (approved):

SmallJoker (approved):
LGTM

PR #3233, opened 2026-02-19
add cactus->dark_green dye

tenplus1:
Add recipe to cook cactus into dark_green dye.

### DIFF
--- a/mods/dye/init.lua
+++ b/mods/dye/init.lua
@@ -48,31 +48,32 @@ for _, row in ipairs(dye.dyes) do
 end
 
 -- Manually add coal -> black dye
-
 minetest.register_craft({
 	output = "dye:black 4",
 	recipe = {
 		{"group:coal"}
-	},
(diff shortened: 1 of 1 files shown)


mods/dye/init.lua
```diff
@@ -65,6 +65,15 @@ minetest.register_craft({
 	},
 })
 
+-- Manually add cooking cactus -> dark green dye
+
+minetest.register_craft({
```
SmallJoker: Should be guarded by `core.get_modpath("default")` because
1. dyes does not depend on default
2. recipe validity checkers might return false-positives

EDIT: Same for the other recipes that rely on `default`.

tenplus1: True, I didnt think a check was necessary since default:blueberries are in the dye recipes also.

SmallJoker (approved):
Thank you for the update. LGTM

PR #3234, opened 2026-03-22
Improve flame sound logic

AbbyRead:
I was looking at the code for mtg's fire, having been recommended to it when trying to figure out how to do looping sounds emitted from nodes.  I ended up finding ways to improve the current implementation.

This PR improves:
- namespace use (minetest to core)
- fpos logic to use vectors
- nearby fire node detection to filter to actual distance (spherical radius instead of a bounding box)
- sound update responsiveness
- adds playback fade in/out

### DIFF
--- a/mods/fire/init.lua
+++ b/mods/fire/init.lua
@@ -194,34 +194,17 @@ if flame_sound then
 			core.sound_stop(handles[player_name])
 			handles[player_name] = nil
 		end
-		-- If flames
+		-- If flames present, find center of flame positions and play sound
 		if flames > 0 then
-			-- Find centre of flame positions
-			local fposmid = fpos[1]
-			-- If more than 1 flame
-			if #fpos > 1 then
-				local fposmin = areamax
-				local fposmax = areamin
+			local fposmid
+			if #fpos == 1 then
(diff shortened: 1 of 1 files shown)


mods/fire/init.lua
```diff
@@ -34,14 +34,15 @@ end
 -- Flame nodes
 local fire_node = {
 	drawtype = "firelike",
-	tiles = {{
+	tiles = { {
```
SmallJoker: Most of the minetest_game code follow the code style from https://docs.luanti.org/for-engine-devs/lua-code-style-guidelines/ . Thus, I think your changes should follow that too for consistency.

AbbyRead: Ah, thanks for the heads up.  Will fix that soon.

SmallJoker:
Sorry for the delay. The changes seem OK. However, the PR does currently mix stylistic and functional changes. Please separate these either by two different PRs or by two separate commits.

EDIT: Also the bracket spacing does not match code code style guidelines: https://docs.luanti.org/for-engine-devs/lua-code-style-guidelines/#spaces (I assume this applied because it's a luanti-org repository too).

PR #3235, opened 2026-03-23
beds: allow setting spawn point during daytime

### LINKED ISSUE #3151: Beds: set spawn point at daytime
Set spawn point during daytime too when right clicking on the bed. Or add setting for this.

GP-commits:
## Summary
Allow players to set their spawn point by right-clicking a bed during daytime.

## Changes
- Set spawn point on right-click even during daytime
- Kept existing night sleep behavior unchanged

## Motivation
Improves convenience without affecting existing mechanics.

Closes #3151

### DIFF
--- a/mods/beds/functions.lua
+++ b/mods/beds/functions.lua
@@ -277,7 +277,6 @@ function beds.on_rightclick(pos, player)
 	if tod > beds.day_interval.start and tod < beds.day_interval.finish then
 		beds.set_spawns(player, pos)
 		if beds.player[name] then
-			
 			lay_down(player, nil, nil, false)
 		end
 		minetest.chat_send_player(name, S("You can only sleep at night."))
@@ -288,7 +287,6 @@ function beds.on_rightclick(pos, player)
 	-- move to bed
 	if not beds.player[name] then
 		lay_down(player, ppos, pos)
-		
 		beds.set_spawns(player, pos) -- save respawn positions when entering bed
 	else
 		lay_down(player, nil, nil, false)
--- a/mods/beds/functions.lua
+++ b/mods/beds/functions.lua
@@ -275,20 +275,19 @@ function beds.on_rightclick(pos, player)
 	local tod = minetest.get_timeofday()
 
 	if tod > beds.day_interval.start and tod < beds.day_interval.finish then
-    if beds.player[name] then
-        lay_down(player, nil, nil, false)
-    end
+		if beds.player[name] then
+			lay_down(player, nil, nil, false)
+		end
+		local current_spawn = beds.spawn[name]
+		if not current_spawn or vector.distance(current_spawn, pos) > 0.1 then
+			beds.set_spawns(player, pos)
+			minetest.chat_send_player(name, S("Spawn point set."))
+		else
+			minetest.chat_send_player(name, S("You can only sleep at night."))
+		end
+		return
+	end
 
-    -- Only set spawn if not already set to this bed
-    local current_spawn = beds.spawn[name]
-    if not current_spawn or vector.distance(current_spawn, pos) > 0.1 then
-        beds.set_spawns(player, pos)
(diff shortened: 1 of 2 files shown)


GP-commits:
Hi, the workflows are awaiting approval. Could you please approve them so checks can run?

ssdaniel24:
Seems like if i will spam clicking on bed at the day, it will write in the chat:
```
You can only sleep at night.
Spawn point set.
You can only sleep at night.
Spawn point set.
You can only sleep at night.
Spawn point set.
...
```

I think you should add check for already set spawnpoint.

GP-commits:
Good catch, thanks! I’ll add a check to prevent repeated spawn setting.

mods/beds/functions.lua
```diff
@@ -275,17 +275,25 @@ function beds.on_rightclick(pos, player)
...
+    -- Only set spawn if not already set to this bed
+    local current_spawn = beds.spawn[name]
+    if not current_spawn or vector.distance(current_spawn, pos) > 0.1 then
+        beds.set_spawns(player, pos)
+        minetest.chat_send_player(name, S("Spawn point set."))
+    end
+
+    minetest.chat_send_player(name, S("You can only sleep at night."))
+    return
+end
```
SmallJoker: Please indent the entire new code block so that it makes sense again.

mods/beds/functions.lua
```diff
@@ -275,17 +275,25 @@ function beds.on_rightclick(pos, player)
...
+    if beds.player[name] then
+        lay_down(player, nil, nil, false)
+    end
+
+    -- Only set spawn if not already set to this bed
+    local current_spawn = beds.spawn[name]
+    if not current_spawn or vector.distance(current_spawn, pos) > 0.1 then
+        beds.set_spawns(player, pos)
+        minetest.chat_send_player(name, S("Spawn point set."))
+    end
```
SmallJoker: How about `else` instead? Have the first click set the spawn point and the 2nd to notify the player that it's yet not time to sleep.

SmallJoker:
> Hi, the workflows are awaiting approval. 

FYI, running the checks locally is straight-forward. Install `luacheck` and give it the path to `minetest_game` to parse. Interestingly it does not catch indent (stylistic) issues.

mods/beds/spawns.lua
```diff
@@ -50,16 +50,12 @@ function beds.save_spawns()
...
-	for name,_ in pairs(beds.player) do
-		local player = minetest.get_player_by_name(name)
-		local p = player:get_pos()
-		-- but don't change spawn location if borrowing a bed
-		if not minetest.is_protected(p, name) then
-			beds.spawn[name] = p
-		end
-	end
-	beds.save_spawns()
+function beds.set_spawns(player, bed_pos)
```
sfan5, line 53: is this a documented API function? if so, won't this break compatibility?

SmallJoker: Same thought here. It's not documented, thus no compat break.
https://github.com/luanti-org/minetest_game/blob/95991f8dc4c97d3cc7945269bf2d5640c7fe6bc8/game_api.txt#L41-L57

GP-commits:
Hi, I've addressed all the review comments:

- Fixed indentation throughout the daytime block

- Changed logic to use else so first click sets spawn and second click shows "You can only sleep at night."

- No backwards compatibility needed since beds.set_spawns() is undocumented

SmallJoker (approved):

mods/beds/spawns.lua
```diff
@@ -50,16 +50,12 @@ function beds.save_spawns()
...
+    local name = player:get_player_name()
+    if not core.is_protected(bed_pos, name) then
```
sfan5, line 56:
```suggestion
    -- don't change spawn location if borrowing a bed
    if not core.is_protected(bed_pos, name) then
```

sfan5 (changes requested):

PR #3238, opened 2026-04-06
Refactor clay node drop

Wolfxyz16:
I am not sure if it was old syntax but if "default:clay" should drop 4 "default:clay_lump" it should use drop tables.


sfan5:
No, the syntax is perfectly okay. Your changes would make it drop only one clay lump.

PR #3159, opened 2024-10-02
Make walls connect to nodes in a new `wall_connected` group

alek13:
`walls`: add ability to specify group `wall_connected` for blocks to which the walls will be connected

PR #3239, opened 2026-04-11
Make fences connects to nodes with `fence_connected` node

alek13:
`default`: fences: add ability to specify group `fence_connected` for blocks that the fences can connect to.

Same as #3159


sfan5 (approved):
OK

PR #3240, opened 2026-05-03
Add {soulbound} check to Bones.

tenplus1:
Change minetest. to core.

### DIFF
--- a/game_api.txt
+++ b/game_api.txt
@@ -85,7 +85,7 @@ in `bones.player_inventory_lists`.
 
 e.g. `table.insert(bones.player_inventory_lists, "backpack")`
 
-Also note that items with the {soulbound=1} group will not be dropped on death.
+Items with the group `soulbond = 1` will not be moved out of the inventory on death.
 
 
 Creative API
--- a/game_api.txt
+++ b/game_api.txt
@@ -85,6 +85,8 @@ in `bones.player_inventory_lists`.
 
 e.g. `table.insert(bones.player_inventory_lists, "backpack")`
 
+Also note that items with the {soulbound=1} group will not be dropped on death.
+
 
 Creative API
 ------------
--- a/mods/bones/init.lua
+++ b/mods/bones/init.lua
@@ -4,20 +4,20 @@
 -- See README.txt for licensing and other information.
 
 -- Load support for MT game translation.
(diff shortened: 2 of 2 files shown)


game_api.txt
```diff
@@ -85,6 +85,8 @@ in `bones.player_inventory_lists`.
 
 e.g. `table.insert(bones.player_inventory_lists, "backpack")`
 
+Also note that items with the {soulbound=1} group will not be dropped on death.
```
SmallJoker: According to the implementation they're also not *moved to bones*.

I'd word this as follows:
```
Items with the group `soulbond = 1` will not be moved out of the inventory on death.
```

Furthermore, I question whether `soulbond` is self-explanatory. How about `keep_on_death` or `bones_keep`? Other ideas are surely welcome too.

tenplus1: Soulbound items are well known in games, and since 3d_armor mod already uses it, and my own luck blocks mod checks for it's presence before taking player items :)

SmallJoker:
Cannot create review comments. GitHub returns HTTP status 500. Hence here:
```Lua
				if not stack:is_empty() and
						core.get_item_group(stack:get_name(), "soulbound") == 0 then
					drop(pos, stack)
```
1. is_empty
2. `core` namespace for new code
3. re-using existing `stack` variable
Points 1 and 2 also for the 2nd code block.

Aside from that - looks good.

tenplus1:
@SmallJoker - Changed minetest. to core. , the is_empty just makes sure there's something to remove, can you explain stack part ?

SmallJoker:
@tenplus1 The changes overall look good. However, they're somewhat difficult to tell apart. Would you please be so nice to have three separate commits? 1x minetest -> core rename, 1x soulbound feature, 1x other improvements?
Otherwise - in case of an issue - we'd have to revert all changes if they cannot be easily separated.

tenplus1:
@SmallJoker - Without making a new pull request I have no idea how to do that, but I can assure you the changes have been well tested.

SmallJoker:
@tenplus1 You might look into `git rebase -i HEAD~7` --> re-order/squash/drop commits (lists the 7 most recent ones). Also `git push --force REMOTE/BRANCH` to overwrite existing history (should only be done in development branches).

PR #3242, opened 2026-05-09
[NO SQUASH] Update Bones mod

tenplus1:
Change minetest. to core.

### DIFF
--- a/game_api.txt
+++ b/game_api.txt
@@ -85,6 +85,8 @@ in `bones.player_inventory_lists`.
 
 e.g. `table.insert(bones.player_inventory_lists, "backpack")`
 
+Items with the group `soulbond = 1` will not be moved out of the inventory on death.
+
 
 Creative API
 ------------
--- a/mods/bones/init.lua
+++ b/mods/bones/init.lua
@@ -282,7 +282,7 @@ core.register_on_dieplayer(function(player)
 	for _, list_name in ipairs(player_inventory_lists) do
 		for i = 1, player_inv:get_size(list_name) do
 			local stack = player_inv:get_stack(list_name, i)
-			if stack:get_count() > 0 and
+			if not stack:is_empty() and
 					core.get_item_group(stack:get_name(), "soulbound") == 0 then
 				if bones_mode == "bones" and inv and inv:room_for_item("main", stack) then
 					inv:add_item("main", stack)
--- a/mods/bones/init.lua
+++ b/mods/bones/init.lua
@@ -210,7 +210,7 @@ local function is_all_empty(player_inv)
 end
 
 core.register_on_dieplayer(function(player)
-	local bones_mode = core.settings:get("bones_mode") or "bones"
+	local bones_mode = core.settings:get("bones_mode")
 	if bones_mode ~= "bones" and bones_mode ~= "drop" and bones_mode ~= "keep" then
 		bones_mode = "bones"
 	end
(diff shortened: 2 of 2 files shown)


mods/bones/init.lua
```diff
@@ -167,12 +167,12 @@ local function may_replace(pos, player)
...
 	-- if the node is unknown, we return false
-	if not node_definition then
+	if not node_definition or node_name == "ignore" then
 		return false
 	end
 
-	-- allow replacing air
-	if node_name == "air" then
+	-- allow replacing air or flowing liquid
+	if node_name == "air" or node_definition.liquidtype == "flowing" then
```
SmallJoker, line 175: This can be abused to dry out fields. Though I think that's acceptable.

tenplus1: This is why bones can only replace water_source nodes if they aren't protected :)

mods/bones/init.lua
```diff
@@ -250,61 +245,52 @@ core.register_on_dieplayer(function(player)
...
+				player_name, pos_string))
+	end
+
+	-- loop through player inventories and place items inside bones or drop
+	for _, list_name in ipairs(player_inventory_lists) do
+		for i = 1, player_inv:get_size(list_name) do
+			local stack = player_inv:get_stack(list_name, i)
+			if stack:get_count() > 0 and
+					core.get_item_group(stack:get_name(), "soulbound") == 0 then
+				if bones_mode == "bones" and inv:room_for_item("main", stack) then
```
SmallJoker: Suggestion: `if inv and inv:room_for_item("main", stack) then` (to make this loop a little more generic).

mods/bones/init.lua
```diff
@@ -250,61 +245,52 @@ core.register_on_dieplayer(function(player)
...
-			meta:set_int("time", (share_bones_time - share_bones_time_early))
+			meta:set_string("infotext", S("@1's bones", player_name))
 		end
-
-		core.get_node_timer(pos):start(10)
+		notify("Bones placed", S("@1 died at @2, and bones were placed.",
+				player_name, pos_string))
 	else
-		meta:set_string("infotext", S("@1's bones", player_name))
+		notify("Inventory dropped", S("@1 died at @2, and dropped their inventory.",
```
SmallJoker, line 277: Nitpick (not a change request but remark): These two log messages are not entirely correct because the items are - at this point - yet not moved.

tenplus1: Saves putting another if statement after drops are done.

mods/bones/init.lua
tenplus1: Good idea :)

mods/bones/init.lua
sfan5, line 175: don't liquids have `buildable_to` anyway? I'm not a fan of this special-casing.

mods/bones/init.lua
```diff
@@ -250,61 +245,52 @@ core.register_on_dieplayer(function(player)
...
+				player_name, pos_string))
+	end
+
+	-- loop through player inventories and place items inside bones or drop
+	for _, list_name in ipairs(player_inventory_lists) do
+		for i = 1, player_inv:get_size(list_name) do
+			local stack = player_inv:get_stack(list_name, i)
+			if stack:get_count() > 0 and
```
sfan5: `not stack:is_empty()`

game_api.txt
```diff
@@ -85,6 +85,8 @@ in `bones.player_inventory_lists`.
 
 e.g. `table.insert(bones.player_inventory_lists, "backpack")`
 
+Items with the group `soulbond = 1` will not be moved out of the inventory on death.
+
```
sfan5, line 89: This needs to be in the same commit as the implementation.

mods/bones/init.lua
```diff
@@ -284,7 +284,7 @@ core.register_on_dieplayer(function(player)
 			local stack = player_inv:get_stack(list_name, i)
 			if stack:get_count() > 0 and
 					core.get_item_group(stack:get_name(), "soulbound") == 0 then
-				if bones_mode == "bones" and inv:room_for_item("main", stack) then
+				if bones_mode == "bones" and inv and inv:room_for_item("main", stack) then
```
sfan5, line 287: and this fix should be squashed into the relevant commit too

tenplus1: Without making a whole new pull request, how do I do that ?

mods/bones/init.lua
tenplus1: The flowing check is specifically before the protection check on purpose, but I could remove the souce check afterwards.

tenplus1:
@sfan5 @SmallJoker - Any changes I make to a commit is automatically added as a new commit.  How do I make it so that it adds to the existing commit instead ???

SmallJoker:
@tenplus1 You can manage your commits like this: https://github.com/luanti-org/minetest_game/pull/3240#issuecomment-4413005052

Some (at the first glance) good guides:
* https://medium.com/@slamflipstrom/a-beginners-guide-to-squashing-commits-with-git-rebase-8185cf6e62ec
* https://www.codecademy.com/article/git-squash-commits "Git squash using interactive rebase"
* Video: https://www.youtube.com/watch?v=ElRzTuYln0M (uses `vim`, thus your interaction with the editor may vary)

Note 1: You can change the [git default editor](https://stackoverflow.com/a/2596835) to make your life easier.
Note 2: The interactive rebase also allows you to re-order commits before squashing. Just get the lines into the order you'd like. Note that you might need to fix merge conflicts if subsequent changes happen to overlap with the changes that you moved back or forth in history.
I hope this helps.

tenplus1:
@SmallJoker - I dont actually use git locally, I only use git gui for my own mods and the actual github website for pull requests.

PR #3243, opened 2026-05-10
[MANUAL SQUASH] Update Bones mod

tenplus1:
Change minetest. to core.

Update game_api.txt to show {soulbound = 1} feature.

### DIFF
--- a/game_api.txt
+++ b/game_api.txt
@@ -85,6 +85,8 @@ in `bones.player_inventory_lists`.
 
 e.g. `table.insert(bones.player_inventory_lists, "backpack")`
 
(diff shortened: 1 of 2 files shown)


SmallJoker:
Previous comments:
* https://github.com/luanti-org/minetest_game/pull/3242
* https://github.com/luanti-org/minetest_game/pull/3240

SmallJoker (approved):
Tested in singleplayer. Bones are still placed or dropped depending on the setting.

tenplus1:
@SmallJoker - The new update is running live on Xanadu server without issue.

PR #3247, opened 2026-05-16
[manual squash] Update Bucket mod

tenplus1:
Initial minetest. to core. tidy

Simplify: Early return checks, nil check, replace set_node air with remove_node (empty bucket).

Simplify2: Early return checks, nil check, default below position with check for above (filled bucket).

### DIFF
--- a/mods/bucket/init.lua
+++ b/mods/bucket/init.lua
@@ -136,7 +136,7 @@ core.register_craftitem("bucket:bucket_empty", {
 		local liquiddef = bucket.liquids[node.name]
 
 		if not liquiddef or not liquiddef.itemname
-		or node.name ~= liquiddef.source then
+				or node.name ~= liquiddef.source then
 			-- non-liquid nodes will have their on_punch triggered
 			local node_def = core.registered_nodes[node.name]
 			if node_def then
@@ -159,7 +159,7 @@ core.register_craftitem("bucket:bucket_empty", {
 
 			-- if space in inventory add filled bucked, otherwise drop as item
 			local inv = user:get_inventory()
-			if inv and inv:room_for_item("main", liquiddef.itemname) then
+			if inv:room_for_item("main", liquiddef.itemname) then
 				inv:add_item("main", liquiddef.itemname)
 			else
 				local upos = user:get_pos()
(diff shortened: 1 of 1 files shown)


sfan5:
My personal opinion is that we shouldn't be changing `minetest` to `core` in existing code. It's busywork.

tenplus1:
@sfan5 - I'll only do it if actually changing the code within.

sfan5:
But what did you actually change? The diff is huge and this is hard to review.

tenplus1:
@sfan5 - The main changes were in the "Simplify" part above with some early returns set and redo of the force_renew, also using remove_node instead of setting to air.

mods/bucket/init.lua
```diff
@@ -135,57 +136,52 @@ core.register_craftitem("bucket:bucket_empty", {
...
-
-		if liquiddef ~= nil
-		and liquiddef.itemname ~= nil
-		and node.name == liquiddef.source then
-			local pname = user:get_player_name()
-			if check_protection(pos, pname, "take ".. node.name) then
-				return
+	
+		if not liquiddef or not liquiddef.itemname
+		or node.name ~= liquiddef.source then
```
SmallJoker: Code style https://docs.luanti.org/for-engine-devs/lua-code-style-guidelines/

```Lua
if not liquiddef or not liquiddef.itemname
		or node.name ~= liquiddef.source then
```

mods/bucket/init.lua
```diff
@@ -135,57 +136,52 @@ core.register_craftitem("bucket:bucket_empty", {
...
-			-- force_renew requires a source neighbour
-			local source_neighbor = false
-			if liquiddef.force_renew then
-				source_neighbor = core.find_node_near(pos, 1, liquiddef.source)
-			end
-			if source_neighbor and liquiddef.force_renew then
-				log_action(pos, pname, "picked up " .. liquiddef.source .. " (force renewed)")
+			-- if space in inventory add filled bucked, otherwise drop as item
+			local inv = user:get_inventory()
+			if inv and inv:room_for_item("main", liquiddef.itemname) then
```
SmallJoker: `inv` is guaranteed to be non-nil. If it is nil for whatever reason, there's probably a mod creating imperfect fake players.

SmallJoker (commented):
Looks good aside from the comments. The adjustments will need manual squashing. Otherwise it'll be very difficult to differentiate renaming vs restructuring changes.

SmallJoker (approved):

PR #3251, opened 2026-06-09
Optimize png files with optipng

Sheriff-Unit-3:
Ran all files through `optipng` to decrease their file size.


sfan5:
We already did this in the past so additional savings out of nowhere are suspicious.
Did you use this script? https://github.com/luanti-org/minetest_game/blob/master/utils/optimize_textures.sh

Sheriff-Unit-3:
I used `find . -name "*.png" -exec optipng -o7 {} +` within my Linux terminal.

sfan5:
Then the results will be wrong. See the comment in the script.

PR #3252, opened 2026-06-15
Improve French translation

xinxinxinxinxin:


mods/keys/locale/keys.fr.tr
```diff
@@ -2,4 +2,4 @@
 
 Key=Clé
 Key to @1's @2=Clé pour @2 de @1
-Skeleton Key=Squelette
+Skeleton Key=Passe-partout
```
SmallJoker, line 5: Where is the "key" part? Does this word include it?

xinxinxinxinxin: The word "passe-partout" don't need to add the word "clé" to understand that it's a key. But if you prefer "Clé passe-partout" is also correct.

SmallJoker (approved):

PR #3261, opened 2026-07-30
Fix TNT API

philipmif:
This fixes unexpected behavior of the TNT API without changing existing functionality.

#### Fixed unexpected behavior:
- Right now, registering a tnt node via this API with `ignore_protection = true` results in a TNT causing no explosion at all. This seems to be a consequence of a careless implementation of the `explode_center` option.
- Secondly, when calling the tnt.boom() function, it is possible to define the explosion just like a tnt definition with `ignore_protection = false`. However as of now there is no possibility to specify a player responsible for the explosion.
- In Addition to that it's currently possible for anyone to ignite a TNT placed by the owner in protected areas, causing a destructive explosion.

#### What's changed:
- Fixed conflict between `ignore_protection` and `explode_center`.
- Added optional `igniter` parameter to tnt.boom(); if not set, behaves as usual.
- Added protection checks upon ignition.
----

This has been partially addressed in already existing PRs. While the tnt mod in general could use a revamp imo, this PR only aims to fix obvious bugs/inconsistencies, given that mtg currently is only in maintenance mode.

### DIFF
--- a/game_api.txt
+++ b/game_api.txt
@@ -553,15 +553,17 @@ TNT API
   * `side`  Side tiles. By default the name of the tnt with a suffix of `_side.png`.
   * `top`  Top tile. By default the name of the tnt with a suffix of `_top.png`.
   * `bottom` Bottom tile. By default the name of the tnt with a suffix of `_bottom.png`.
-  * `burning` Top tile when lit. By default the name of the tnt with a suffix of `_top_burning_animated.png".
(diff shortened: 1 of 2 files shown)


sfan5:
Duplicate of #2916?

philipmif:
> Duplicate of #2916?

Well, partially, you could say, but only with regard to the first issue mentioned, and as that PR hasn't been reviewed for at least 2 years now, I focused on minimal changes just fixing the bug without changing how it worked up to this point.

PR #3264, opened 2026-08-05
Add craftguide pause on load

tenplus1:
This adds a 0.1 second pause before building the crafting guide, as some mods wait until others have loaded before doing additional registrations.  This simple delay fixes the missing craft recipes in certain mods.

### DIFF
--- a/mods/mtg_craftguide/init.lua
+++ b/mods/mtg_craftguide/init.lua
@@ -145,7 +145,7 @@ local function cache_usages(recipe)
 	end
 end
 
-minetest.register_on_mods_loaded(function()
+minetest.after(1, function()
 	for name, def in pairs(minetest.registered_items) do
 		if show_item(def) then
 			local recipes = get_craftable_recipes(name)
--- a/mods/mtg_craftguide/init.lua
+++ b/mods/mtg_craftguide/init.lua
@@ -145,7 +145,7 @@ local function cache_usages(recipe)
 	end
 end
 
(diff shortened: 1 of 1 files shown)


SmallJoker:
For the sake of completeness:

* `unified_inventory` uses `core.after(0.01, ` (first server step)
* `i3` uses `core.register_on_mods_loaded`
* `craftguide` uses `core.register_on_mods_loaded`

EDIT: If a player happens to join within that 1 second, they will see the text `No items to show.` until they search for something. This might be an issue on system with very fast world joins. Hence, I would recommend lowering this value to `0` (or very close to zero) such that it is executed right after the deadline (which is the start of ServerEnv, after `on_mods_loaded`).

SmallJoker (approved):
LGTM. Thanks.

mods/mtg_craftguide/init.lua
```diff
@@ -145,7 +145,7 @@ local function cache_usages(recipe)
 	end
 end
 
-minetest.register_on_mods_loaded(function()
+minetest.after(0.01, function()
```
sfan5, line 148: this is 10ms (so in practice a single server step), not *1 second*

just saying

SmallJoker: Changed because 1 second may cause unintended side-effects when joining the world too fast --> https://github.com/luanti-org/minetest_game/pull/3264#issuecomment-5193941105

PR #3266, opened 2026-08-15
Update translation files

Wuzzy2:
This PR does the following:

* Update the locale files
* Update the German translation

One string is added: "Spawn point set." for the `beds` mod.


SmallJoker:
:+1: LGTM. Thanks.

PR #3268, opened 2026-08-30
player_api: Support multi-track animations & blend

### LINKED ISSUE #3267: `player_api` Support Multi-track Animation Models
Now that Luanti v5.17 has released there will likely be mod developers creating multi-track player models. As such it would make sense for `player_api` to support such models.

Sheriff-Unit-3:
Allows `player_api` to accept player models that use multi-track animations. Only works if the server is `v5.17+`.

Adds `player_api.stop_animation()` and `blend` to `player_api.set_animation()`.

To test enable the included mod on a MTG world and watch yourself through third-person.
[multitrack.zip](https://github.com/user-attachments/files/31606109/multitrack.zip)

closes #3267


game_api.txt
```diff
@@ -169,11 +169,11 @@ Doors API
 
 The doors mod allows modders to register custom doors and trapdoors.
 
-`doors.registered_doors[name] = Door definition` 
- * Table of registered doors, indexed by door name 
+`doors.registered_doors[name] = Door definition`
```
appgurueu, line 172: unrelated whitespace changes, better to keep this diff clean, there are a bunch of these (presumably: your editor strips trailing whitespace).

game_api.txt
```diff
@@ -485,6 +485,13 @@ The player API can register player models and update the player's appearance.
 	* `player`: PlayerRef
 	* `anim_name`: name of the animation
 	* `speed`: keyframes per second. If nil, the default from the model def is used
+	* `loop`: boolean
+	* `blend`: float
```
appgurueu, line 489: this is also new? PR title should mention this

edit: edited the PR title

mods/player_api/api.lua
```diff
@@ -34,6 +29,7 @@ function player_api.register_model(name, def)
 		animation.eye_height = animation.eye_height or def.eye_height
 		animation.collisionbox = animation.collisionbox or def.collisionbox
 		animation.override_local = animation.override_local or false
+		animation.tracks = animation.tracks or false
```
appgurueu, line 32:
```suggestion
		animation.tracks = animation.tracks or nil
```

more idiomatic imo

mods/player_api/api.lua
```diff
@@ -116,31 +112,37 @@ function player_api.set_texture(player, index, texture)
...
+	player_data.animation_tracks = tracks
 	-- If necessary change the local animation (only seen by the client of *that* player)
 	-- `override_local` <=> suspend local animations while this one is active
 	-- (this is basically a hack, proper engine feature needed...)
-	if anim.override_local ~= previous_anim.override_local then
+	if tracks then
+		-- Can't set_local_animation with multi-track animations, engine limitation
+		anim.override_local = true
+	end
+	if (anim.override_local ~= previous_anim.override_local) or tracks then
```
appgurueu, line 145: isn't the `or tracks` redundant?

the intent here is to only call `set_local_animation` if we need to toggle the local animation on/off. but if it's already in the correct state from the last call there's nothing to toggle?

mods/player_api/api.lua
```diff
@@ -153,13 +155,50 @@ function player_api.set_animation(player, anim_name, speed, loop)
...
 	end
 	-- Set the animation seen by everyone else
-	player:set_animation(anim, speed, animation_blend, loop)
-	-- Update related properties if they changed
-	if anim._equals ~= previous_anim._equals then
-		player:set_properties({
-			collisionbox = anim.collisionbox,
-			eye_height = anim.eye_height
-		})
+	if tracks and core.features.hud_hideable_field == true then
```
appgurueu, line 158: nit: why we're checking for `core.features.hud_hideable_field` is not obvious. would need at least a comment.

but more importantly: what does this check buy us? it is not enough to guarantee that this actually works, since the client could be on a lower protocol version.

since this is an API and there is no reasonable fallback, we can make this the API user's responsibility: if you use the multi-track player_api, you must ensure server and client support it. (this of course needs to be documented in `game_api.txt`.)

if someone tries to use this on a *server* that doesn't even support it, i'd strongly lean towards making it a hard error as early as possible, so on model registration: something like `assert(not anim.tracks or core.features.hud_hideable_field, "Multi-track animations not supported, needs Luanti 5.17.0 or newer")`.

(side note, if you already have an objref, you don't need a feature flag to determine if the API is available, you can just check if `player.play_animation` exists. but obviously this wouldn't really work at model registration without hacks.)

mods/player_api/api.lua
```diff
@@ -153,13 +155,50 @@ function player_api.set_animation(player, anim_name, speed, loop)
...
+				min_frame = track.x or 0,
+				max_frame = track.y or math.huge,
+				start_frame = track.start or 0,
+				speed = track.speed or speed,
+				loop = track.loop or loop,
+				blend = track.blend or blend,
+				priority = track.priority or 0
+			})
+		end
+	elseif core.features.hud_hideable_field ~= true then
```
appgurueu, line 173: as said, i'd get rid of this here

mods/player_api/api.lua
```diff
@@ -153,13 +155,50 @@ function player_api.set_animation(player, anim_name, speed, loop)
...
-			collisionbox = anim.collisionbox,
-			eye_height = anim.eye_height
-		})
+	if tracks and core.features.hud_hideable_field == true then
+		if previous_anim_name then
+			player_api.stop_animation(player, previous_anim_name)
+		end
+		for track_name, track in pairs(tracks) do
+			player:play_animation(track_name, {
+				min_frame = track.x or 0,
```
appgurueu, line 164: thought: maybe tracks should use `min_frame` / `max_frame` to be consistent with the engine. i think x and y is poor naming. (then again we might also prefer consistency within player_api 🤷 )

mods/player_api/api.lua
```diff
@@ -153,13 +155,50 @@ function player_api.set_animation(player, anim_name, speed, loop)
...
+			})
+		end
+	end
+end
+
+function player_api.stop_animation(player, anim_name)
+	local player_data = get_player_data(player)
+	local model = models[player_data.model]
+	local anim = model.animations[anim_name]
+	if not (model and anim) or core.features.hud_hideable_field ~= true then
```
appgurueu, line 191: same thing about the feature check here

also: i wouldn't just turn missing model/anim into a no-op. should be an error.

mods/player_api/api.lua
```diff
@@ -153,13 +155,50 @@ function player_api.set_animation(player, anim_name, speed, loop)
...
+		return
+	end
+	local current_anims = player:get_animations()
+	local tracks = anim.tracks or false
```
appgurueu, line 195:
```suggestion
	local tracks = anim.tracks
```

mods/player_api/api.lua
```diff
@@ -153,13 +155,50 @@ function player_api.set_animation(player, anim_name, speed, loop)
...
+	if current_anims and tracks then
+		for track_name, _ in pairs(tracks) do
```
appgurueu, line 197:
```suggestion
		for track_name in pairs(tracks) do
```

mods/player_api/api.lua
```diff
@@ -153,13 +155,50 @@ function player_api.set_animation(player, anim_name, speed, loop)
...
+			if current_anims[track_name] then
```
appgurueu, line 198: do you need the current_anims check? why not call `stop_animation` indiscriminately? it's a no-op if the animation isn't currently playing.

game_api.txt
```diff
@@ -515,14 +522,30 @@ The player API can register player models and update the player's appearance.
...
 			--   -- suspend client side animations while this one is active (optional)
 			--   override_local = <true/false>
 			-- },
+			-- [anim_name] = {               -- For multi-track animation models
+			--   tracks = {
+			--     [track_name] = {          -- Name of the animation track in 3d modeling software
+			--       x = <min_frame>,        -- (optional)
+			--       y = <max_frame>,        -- (optional)
+			--       start = <start_frame>,  -- (optional)
+			--       speed = 30              -- (optional) Falls back to animation_speed
```
appgurueu, line 539: then = 30 is misleading? should be = animation_speed

mods/player_api/api.lua
```diff
@@ -153,13 +155,50 @@ function player_api.set_animation(player, anim_name, speed, loop)
...
+			player:set_properties({
+				collisionbox = anim.collisionbox,
+				eye_height = anim.eye_height
+			})
+		end
+	end
+end
+
+function player_api.stop_animation(player, anim_name)
+	local player_data = get_player_data(player)
```
appgurueu, line 188: shouldn't this touch `player_data.animation*`? otherwise how does `set_animation` know the animation has ceased playing?

appgurueu (commented):
This kind of "shoehorns" multi-track animation support into player_api. 
Stuffing an API that expects a single animation playing at a time and an API that lets you control multiple channels into the same interface is messy, which is why I avoided it on the engine side and went with new APIs instead.

Maybe we could also do that for player_api?
Introduce a new `play_animation` API distinct from `set_animation`. Have a mapping so `walk` + `mine` becomes `walk_mine` in the getter and `walk_mine` in the setter becomes `walk` + `mine` in terms of tracks again. (Ideally `walk_mine` would not need to exist with the new API, but well compatibility...)
You seem to already be going in that direction a bit with `stop_animation` (which then suffers from a mismatch with `set_animation`).

mods/player_api/api.lua
Sheriff-Unit-3: I attempted to use `player:get_animations()` in place of the feature flag, but that crashes older versions of Luanti because they can't find `get_animations` for object player.

mods/player_api/api.lua
Sheriff-Unit-3: I do prefer `min_frame`/`max_frame` as well, makes it clearer what it refers to.

game_api.txt
Sheriff-Unit-3: `loop` existed before in lua code (but not in documentation), but `blend` is new. I see that you likely figured this out already though.

mods/player_api/api.lua
appgurueu: hmm `if player.get_animations then ... end` should not crash any luanti version.

what exactly was the test you performed? `if player:get_animations() then ... end` is a different test, it will try to call `player.get_animations(player)` which will fail on an older luanti version as `player.get_animations` will be `nil` there.

mods/player_api/api.lua
Sheriff-Unit-3: Your right, will remove it.

mods/player_api/api.lua
Sheriff-Unit-3: I tried `player:get_animations()`, didn't know that `player.get_animations` existed at all until now lol.

PR #3270, opened 2026-09-20
fix dupe placement in leaves

tenplus1:
This fixes the duplicate placement of leaf nodes.


sfan5:
What's the root cause? How to reproduce it? Has this been reported as a bug?

mods/default/functions.lua
```diff
@@ -536,7 +537,7 @@ local function leafdecay_after_destruct(pos, oldnode, def)
 		local node = minetest.get_node(v)
 		local timer = minetest.get_node_timer(v)
 		if node.param2 ~= 1 and not timer:is_started() then
-			timer:start(math.random(20, 120) / 10)
+			timer:start(math.random(20, 120) * 0.1)
```
sfan5: ?

tenplus1: Oh sorry, I find multiplication faster in lua than division, especially for leafdecay :P

tenplus1:
The player initially places the leaf then it uses default.after_place_leaves to set_node again with param2 as 1, so double placemenet functions all around.

SmallJoker:
huh? `default.after_place_leaves` is for `ndef.after_place_node`, which is executed by `core.item_place`, the default action for `ndef.on_place` (given by the `core.nodedef_default` metatable).

`core.set_node` does not invoke `ndef.after_place_node`. You can test this by running `//lua core.set_node(player:get_pos(), { name = "default:leaves" })` with worldedit enabled and a `print("foobar")` line in `default.after_place_leaves`.

If you want the callback to run, do use `core.place_node(...)`.

tenplus1:
@SmallJoker - Added an override to default:leaves and put an on_construct function just to print hello, placing leaves prints it twice because it's placed twice, and the swap_node fixes this.

SmallJoker:
I see. In that case it  makes sense. Knowing where you're coming from would've saved myself several minutes digging through the code...

SmallJoker (approved):

mods/default/functions.lua
SmallJoker: LuaJIT should already optimize this. Not sure about plain Lua.

